PR #3, opened 2021-04-22
Add SEP defining transactions required

### LINKED ISSUE #5: Draft a SEP for the Payment Channel Protocol
CAP-21 contains a brief description of how to use CAP-21's new preconditions to support a two-way payment channel protocol, but there are more things that participants need to agree on in regards to the transactions and blockchain state. We should document all these things in a new SEP. We can work on and iterate on the SEP in this repo for the moment until we have some confidence we want to propose it as a SEP.

leighmcculloch:
### What
Add a SEP that defines the transactions required to form a payment channel, using the new preconditions in CAP-21, and based on the two-way payment channel in CAP-21's rationale section.

### Why
CAP-21 contains a brief description of how to use CAP-21's new preconditions to support a two-way payment channel protocol, but there are more things that participants need to agree on in regards to the transactions and blockchain state. We should document all these things in a new SEP.

We can work on and iterate on the SEP in this repo for the moment until we have prototyped it and have some confidence we want to propose it as a SEP in the stellar-protocol repository.

This SEP contains no discussion of a transport protocol because it is out-of-scope. We aren't at a point where it is really important to document that concretely as any prototypes we develop now will use whatever simple transport mechanism they can to make the prototype work. We'll document that separately in some future milestone as a separate SEP.

Close #5

### DIFF
--- a/specifications/sep-payment-channel-mechanism.md
+++ b/specifications/sep-payment-channel-mechanism.md
@@ -285,20 +285,39 @@ declaration transaction and closing the channel at the actual final state.
 4. Wait observation period O
 5. Submit C_i
 
-#### Mutating the Channel
+#### Changing the Channel Setup
 
 The payment channel setup can be altered with on-chain transactions after
-channel setup. The following operations can take place. Each operation is
-implemented in a two-step process where participants first agree on a new state
-at a future iteration that is not yet executable, then participants sign a
-transaction to make that new state possible.
+channel setup. Some of the operations used to alter the channel setup may fail
+even if the transactions are valid, while others will always succeed if the
+transactions are valid.
+
+Each operation is implemented in a two-step process. Participants agree on a new
+closing state at a future iteration by signing C_i and D_i transactions where i
+has skipped an iteration that is not yet executable because the D_i's
+`minSeqNum` is also set in the future. Participants then sign a transaction to
+make the change that only moves the sequence of escrow account E to satisfy the
+`minSeqNum` of the future D_i.
+
+Operations that can fail have the following requirements as well:
+
+- The transaction that can fail must have its source account set to an account
+that is not escrow account E.
+- The transaction that can fail must contain a `BUMP_SEQUENCE` operation that
+bumps escrow account E's sequence number to a sequence number that makes the D_i
+executable.
+
+Operations that cannot fail:
+
+- [Change the Observation Period](#Change-the-Observation-Period)
+
+Operations that can fail:
 
 - [Add Trustline](#Add-Trustline)
 - [Remove Trustline](#Remove-Trustline)
 - [Deposit by Initiator](#Deposit-by-Initiator)
 - [Deposit by Responder](#Deposit-by-Responder)
 - [Withdraw](#Withdraw)
-- [Change the Observation Period](#Change-the-Observation-Period)
 
 ##### Add Trustline
 
--- a/specifications/sep-payment-channel-mechanism.md
+++ b/specifications/sep-payment-channel-mechanism.md
@@ -106,19 +106,20 @@ until all claimable balances created at close for R are claimed by R.
 
 ### Variables
 
-The two participants maintain the following two variables during the lifetime
-of the channel:
+The two participants maintain the following variables during the lifetime of
+the channel:
 
-- s, the _starting sequence number_, is initialized to one greater
-than the sequence number of the escrow account E after E has been created and
-configured. It is changed only when withdrawing from or topping up the
+- s, the _starting sequence number_, is initialized to one greater than the
+sequence number of the escrow account E after E has been created and
+configured.  It is changed only when withdrawing from or topping up the
 escrow account E.
 
-- i, the _iteration number_ of the payment channel, is initialized to
-((s+1)/2)+1.  It is incremented with every off-chain update of the payment
-channel state.
+- i, the _iteration number_ of the payment channel, is initialized to zero.
+It is incremented with every off-chain update of the payment channel state,
+or on-chain setup, deposit, withdrawal.
 
-TODO: Change how the iteration number is initialized and used.
+- s_i, the _iteration sequence number_, is the sequence number that the
+iteration's transactions set starts at.  It is always computable as, s+(i*2).
 
 ### Processes
 
@@ -128,9 +129,16 @@ To setup the payment channel:
 
 1. I creates the escrow account E.
 2. R creates the reserve account V.
-3. I and R sign and exchange signatures for the first closing transaction C_i.
-4. I and R sign and exchange signatures for the first declaration transaction D_i.
-5. I and R sign and exchange signatures for the formation transaction F.
+3. Set variable initial states:
+   - s to E's sequence number + 1.
+   - i to 0.
+5. I and R prepare the formation transaction F.
+6. I and R follow the [Update Process](#Update-Process), including the step
+to increment i, to prepare, sign, and exchange declaration and closing
+transactions that allow the payment channel to be closed with disbursements
+matching the initial contributions.
+7. I and R sign and exchange signatures for formation transaction F.
+8. I or R submit F.
 
 The transactions are constructed as follows:
 
@@ -138,9 +146,10 @@ The transactions are constructed as follows:
 
 - D_i, see [Update Process](#Update-Process).
 
-- F, the _formation transaction_, deposits R's contribution to escrow account
-E, and changes escrow account E and reserve account V to be 2-of-2 multisig
-accounts.  F has source account E, and sequence number set to s.
+- F, the _formation transaction_, deposits I and R's contributions to escrow
+account E, R's reserves to reserve account V, and changes escrow account E
+and reserve account V to be 2-of-2 multisig accounts.  F has source account
+E, and sequence number set to s_i.
 
   F contains operations:
 
@@ -173,14 +182,14 @@ accounts.  F has source account E, and sequence number set to s.
 To update the payment channel state, the participants:
 
 1. Increment i.
-2. Sign and exchange a closing transactions C_i.
+2. Sign and exchange a closing transaction C_i.
 3. Sign and exchange a declaration transaction D_i.
 
 The transactions are constructed as follows:
 
 - C_i, the _closing transaction_, disburses funds to R and changes the
 signing weights on E such that I unilaterally controls E.  C_i has source
-account E, sequence number 2i+1, and a `minSeqAge` of O (the observation
+account E, sequence number s_i+1, and a `minSeqAge` of O (the observation
 period).
 
   The `minSeqAge` prevents a misbehaving party from executing C_i when the
@@ -199,24 +208,53 @@ period).
   - One or more `SET_OPTIONS` operation adjusting reserve account E's
   thresholds to give R full control of V, and removing I's signers.
 
-- D_i, the _declaration transaction_, declares an intent to execute
-the corresponding closing transactions IC_i/RC_i.  D_i has source account E,
-sequence number 2i, and `minSeqNum` set to s.  Hence, D_i can execute at any
-time, so long as E's sequence number n satisfies s <= n < 2i.  Because C_i
-has source account E and sequence number 2i+1, D_i leaves E in a state where
-C_i can execute.
+- D_i, the _declaration transaction_, declares an intent to execute the
+corresponding closing transaction C_i.  D_i has source account E, sequence
+number s_i, and `minSeqNum` set to s.  Hence, D_i can execute at any time, so
+long as E's sequence number n satisfies s <= n < s_i.  Because C_i has source
+account E and sequence number s_i+1, D_i leaves E in a state where C_i can
+execute.
 
   D_i does not require any operations, but since Stellar disallows empty
   transactions, it contains a `BUMP_SEQUENCE` operation with sequence value 0
   as a no-op.
 
-#### Uncooperative Close
-
-TODO: 
-
 #### Cooperative Close
 
-TODO: 
+Participants can agree to close the channel immediately by modifying and
+resigning the most recently signed declaration transaction.  The participants
+change the `minSeqAge` to zero.
+
+1. Modify the most recent D_i `minSeqAge` to zero.
+2. Resign and exchange the modified declaration transaction D_i.
+3. Submit modified D_i
+3. Submit C_i
+
+#### Uncooperative Close
+
+Either participant can close the channel at the latest state by submitting
+the most recently declaration transaction, waiting the observation period O,
+then submitting the closing transaction.
+
+1. Submit most recent D_i
+2. Wait observation period O
+3. Submit C_i
+
+#### Contesting an Uncooperative Close
+
+Either participant can attempt to close the channel at an earlier state that
+benefits them using the uncooperative close process.  The other participant
+can identify that the close process has started at an earlier state by
+monitoring changes in the accounts sequence.  If the other participant sees
+the sequence number of escrow account E change to a value that is not the
+most recent consumed s_i, they can use the following process to contest the
+close.  A participant contests a close by submitting a more recent
+declaration transaction.
+
+1. Monitor if E's sequence number changes to a value that is not s_i
+2. Submit most recent D_i
+3. Wait observation period O
+4. Submit C_i
 
 #### Top-up by Initiator
 
@@ -335,17 +373,25 @@ written-off cost to themselves.
 
 ## Security Concerns
 
-The closing transaction, C_i, creates one or more new claimable balance
-ledger entries that each require sponsoring.  If the sponsor has insufficient
-native asset the closing transactions will fail and consume the sequence
-number of the escrow account E, locking the funds.  If that situation occurs
-with the most recently agreed upon closing transaction, fair distribution of
-the assets depends on both participants signing new closing transactions.
+### Closing Transaction Failure
 
-To avoid this situation it is critical that participants lock sufficient
-funds up-front to provide the reserve, and that both participants monitor
-base reserve changes in the network and respond by adding additional native
-asset if required.
+The closing transaction, C_i, must never fail.  Under the conditions of the
+Stellar Consensus Protocol as it is defined today, and under correct use of
+this protocol, there is no known conditions that will cause it to fail.  It
+will be either invalid or valid and successful, but not valid and failed.  If
+C_i was to be valid and fail it would consume a sequence number and fair
+distribution of the assets within the escrow account would require the
+cooperation of all participants.
+
+If this protocol is not implemented correctly one condition that can result
+in the closing transaction failing is if there is not sufficient native asset
+to sponsor the ledger entries created by the transaction.  The closing
+transaction creates one or more new claimable balance ledger entries that
+each require sponsoring.  If the sponsor has insufficient native asset the
+closing transaction will fail.  To avoid this situation it is critical that
+participants lock sufficient funds up-front to provide the reserve, and that
+both participants monitor base reserve changes in the network and respond by
+adding additional native asset if required.
 
 ## Limitations
 
--- a/specifications/sep-payment-channel-mechanism.md
+++ b/specifications/sep-payment-channel-mechanism.md
@@ -395,8 +395,7 @@ This step yields transactions D_i' and C_i' where i' = i+1.
 increment i, to build, sign, and exchange declaration and closing transactions
 that define how the assets held by the escrow account will be disbursed at close
 of the channel such that the deposited amount included in P_i will be disbursed
-   to participant R. This step yields transactions D_i'' and C_i'' where i'' =
-   i+2. **D_i''**'s maxSeqNo is additionally set to i.
+to participant R. This step yields transactions D_i'' and C_i'' where i'' = i+2.
 4. I and R sign and exchange signatures for deposit transaction P_i.
 5. I or R submit P_i.
 6. Wait for E's sequence number to be P_i's.
--- a/specifications/sep-payment-channel-mechanism.md
+++ b/specifications/sep-payment-channel-mechanism.md
@@ -1,0 +1,66 @@
+## Preamble
+
+```
+SEP: ????
+Title: Payment Channel
+Author: Leigh McCulloch <@leighmcculloch>
+Track: Standard
+Status: Draft
+Discussion: https://github.com/stellar/experimental-payment-channels/issues
+Created: 2021-04-21
+Updated: 2021-04-21
+Version: 0.0.1
+```
+
+## Summary
+
+This protocol defines the transactions that two participants use to open,
+mutate, and close a payment channel.
+
+## Dependencies
+
+This protocol is dependent on [CAP-21] and is built on the example two-way
+payment channel protocol defined in that CAP's rationale.
+
+This protocol is also dependent on [CAP-23] providing claimable balances and
+[CAP-33] providing sponsorship.
+
+## Motivation
+
+Stellar account holders who frequently interact with a set of partners must
+choose between performing every transaction on-chain, or trusting each other
+to some degree and performing net-settlement. Transactions on-chain are fast
+and affordable but it would be beneficial to some operators if speed and cost
+were not bounded and only limited by the partners own capability to agree.
+
+## Abstract
+
+This protocol defines...
+
+## Use Cases
+
+
+## Specification
+
+TODO
+
+## Security Concerns
+
+TODO
+
+## Limitations
+
+This protocol defines the mechanisms of the Stellar network's core protocol
+that are used to enforce agreements made by two participants. This protocol
+does not define the transport through which the agreements are coordinated,
+or the methods through which more than two participants can coordinate and
+exchange dependent agreements. These issues are likely to be discussed in a
+separate proposal.
+
+## Implementations
+
+TODO
+
+[CAP-21]: https://stellar.org/protocol/cap-21
+[CAP-23]: https://stellar.org/protocol/cap-23
+[CAP-33]: https://stellar.org/protocol/cap-33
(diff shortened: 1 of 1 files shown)


specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,655 @@
...
+  - One `END_SPONSORING_FUTURE_RESERVES` operation that confirms reserve
+  account V's sponsorship.
+  - One or more `SET_OPTIONS` operation adjusting escrow account E's
+  thresholds to give I full control of E, and removing R's signers.
+  - One or more `SET_OPTIONS` operation adjusting reserve account E's
+  thresholds to give R full control of V, and removing I's signers.
+
+- D_i, the _declaration transaction_, declares an intent to execute the
+corresponding closing transaction C_i.  D_i has source account E, sequence
+number s_i, and `minSeqNum` set to s_e.  Hence, D_i can execute at any time, so
```
acharb, line 241: s_e is a little confusing to me. Say we start at this state:

```
s (E's seqNum) = 1000, i = 0, e = 0
s_i = 1000
s_e = 1000
```

First C_i/D_i get created but not submitted, i =1.
Second C_i/D_i get created and submitted, i = 2, e=1

This second D_i set E.seqNum to 1000 + 2*1 = 1002
This second C_i then sets E.seqNum to 1003

The current state is then:
```
s = 1003, i=2, e=1
s_i = 1004
s_e = 1002
```

So for a third D_i, shouldn't the minSeqNum be 1003, and thus E's current seqNum. And not s_e?

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,655 @@
...
+The payment channel utilizes two Stellar accounts that are both 2-of-2
+multisig accounts while the channel is open:
+
+- E, the _escrow account_, that holds the assets that both participants have
+contributed to the channel and that will be distributed to the participants
+at channel close according to the final close transactions submitted.
+Created by I.  Jointly controlled by I and R while the channel is open.
+Control is returned to I at close.
+
+- V, the _reserve account_, that holds an amount of native asset contributed
```
acharb, line 107: a second multisig account eventually owned by R for R's claimable balances makes sense to me

acharb (commented):
👋  Had 2 Qs. I might be doing something wrong since I've only been able to read through this once or twice, but I think that's the input that you're looking for 😬 

starting i and e at 0 makes sense to me, which the CAP-21 doesn't specify. The use of s_e is where I'm still a little confuzzled

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,655 @@
...
+by R that will be used to sponsor the claimable balance ledger entries
+created at disbursement.  Created by R.  Jointly controlled by I and R while
+the channel is open.  Control is returned to R at close.  Cannot be merged
+until all claimable balances created at close for R are claimed by R.
+
+### Constants
+
+The two participants agree on the following constants:
+
+- m, the _maximum transaction account of an iteration's transaction set_, is
```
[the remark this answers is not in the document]
leighmcculloch: Oops, this is a typo. Should be:
```suggestion
- m, the _maximum transaction count for an iteration's transaction set_, is
```
It's the number of transactions that can be in an iteration, which is yup, always 2.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: We chatted offline about this, so I think this is resolved. The things that we noted are:
 - Once D is submitted, for e.g. D_2 in the example above, the participants don't sign a third transaction set because the channel is closing.
 - s_e doesn't really matter for channel close, so the doc doesn't mention it being updated in the close steps. We could add it for completeness, but I think it's probably unnecessary? If the second D above was submitted, theoretically e would be equal to 2, the iteration number for the transaction set being submitted.

We used the following table to help think about how s_i and s_e change:
|Description                           |Transaction|Tx Seq|E's seq change|E's seq|s   |i change|i  |e change|e  |s_i |s_e |m  |
|--------------------------------------|-----------|------|--------------|-------|----|--------|---|--------|---|----|----|---|
|Setup: 1. I creates escrow account E  |           |      |1000          |1000   |    |        |   |        |   |    |    |2  |
|Setup: 3. Set variable initial states |           |      |              |1000   |1001|0       |0  |0       |0  |1001|1001|2  |
|                                      |           |      |              |1000   |1001|        |0  |        |0  |1001|1001|2  |
|Setup: 5. Build formation transaction.|F          |1001  |              |1000   |1001|        |0  |        |0  |1001|1001|2  |
|                                      |           |      |              |1000   |1001|        |0  |        |0  |1001|1001|2  |
|Setup: 6. Increment i                 |           |      |              |1000   |1001|1       |1  |        |0  |1003|1001|2  |
|Setup: 7. Build and sign C_i          |C_1        |1004  |              |1000   |1001|        |1  |        |0  |1003|1001|2  |
|Setup: 8. Build and sign D_i          |D_1        |1003  |              |1000   |1001|        |1  |        |0  |1003|1001|2  |
|                                      |           |      |              |1000   |1001|        |1  |        |0  |1003|1001|2  |
|Setup: 10. Submit F                   |           |      |1001          |1001   |1001|        |1  |        |0  |1003|1001|2  |
|                                      |           |      |              |1001   |1001|        |1  |        |0  |1003|1001|2  |
|------                                |           |      |              |1001   |1001|        |1  |        |0  |1003|1001|2  |
|                                      |           |      |              |1001   |1001|        |1  |        |0  |1003|1001|2  |
|Update: Increment i                   |           |      |              |1001   |1001|2       |2  |        |0  |1005|1001|2  |
|Update: Build and sign C_i            |C_2        |1006  |              |1001   |1001|        |2  |        |0  |1005|1001|2  |
|Update: Build and sign D_i            |D_2        |1005  |              |1001   |1001|        |2  |        |0  |1005|1001|2  |
|                                      |           |      |              |1001   |1001|        |2  |        |0  |1005|1001|2  |
|------                                |           |      |              |1001   |1001|        |2  |        |0  |1005|1001|2  |
|                                      |           |      |              |1001   |1001|        |2  |        |0  |1005|1001|2  |
|Update: Increment i                   |           |      |              |1001   |1001|3       |3  |        |0  |1007|1001|2  |
|Update: Build and sign C_i            |C_3        |1008  |              |1001   |1001|        |3  |        |0  |1007|1001|2  |
|Update: Build and sign D_i            |D_3        |1007  |              |1001   |1001|        |3  |        |0  |1007|1001|2  |
|                                      |           |      |              |1001   |1001|        |3  |        |0  |1007|1001|2  |
|---                                   |           |      |              |1001   |1001|        |3  |        |0  |1007|1001|2  |
|                                      |           |      |              |1001   |1001|        |3  |        |0  |1007|1001|2  |
|Close: Submit D_2                     |           |      |1005          |1005   |1001|        |3  |2       |2  |1007|1005|2  |
|Close: Submit D_3                     |           |      |1007          |1007   |1001|        |3  |3       |3  |1007|1007|2  |
|Close: Submit C_3                     |           |      |1008          |1008   |1001|        |3  |        |3  |1007|1007|2  |
|                                      |           |      |              |1008   |1001|        |3  |        |3  |1007|1007|2  |

Ref: https://docs.google.com/spreadsheets/d/1cOXL3mYdMTNDu0FE8lgNMa0fjJ2T-S2jolLG5Dl0xRM

acharb: yup that answered it thx. Also in case someone else sees this I was wondering about s_e after the close in case _I_ wanted to continue the channel (I was thinking i and e would remain at 3). But after close the variables reset to 0, so s_e after close doesn't matter (as you said)

leighmcculloch: That's a great point. This is an edge case we should make clearer. I added a section about it in ab21927.

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+
+1. Increment i.
+2. Sign and exchange a closing transaction C_i.
+3. Sign and exchange a declaration transaction D_i.
+
+The transactions are constructed as follows:
+
+- C_i, the _closing transaction_, disburses funds to R and changes the signing
+weights on E such that I unilaterally controls E.  C_i has source account E,
+sequence number s_i+1, a `minSeqAge` of O (the observation period time
```
accordeiro, line 219: Wouldn't this be `s_i+m`?

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+  - One or more `SET_OPTIONS` operation adjusting escrow account E's
+  thresholds to give I full control of E, and removing R's signers.
+  - One or more `SET_OPTIONS` operation adjusting reserve account E's
+  thresholds to give R full control of V, and removing I's signers.
+
+- D_i, the _declaration transaction_, declares an intent to execute the
+corresponding closing transaction C_i.  D_i has source account E, sequence
+number s_i, and `minSeqNum` set to s_e.  Hence, D_i can execute at any time, so
+long as E's sequence number n satisfies s_e <= n < s_i.  Because C_i has source
+account E and sequence number s_i+1, D_i leaves E in a state where C_i can
```
accordeiro, line 243: Same as above: `s_i+m`

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+  - One or more `PAYMENT` operations depositing R's contribution to E.
+  - One or more `PAYMENT` operations depositing R's reserves to V, for each
+  trustline on E that will be used to sponsor claimable balances at
+  disbursement.
+  
+- C_i, see [Update](#Update) process.
+
+- D_i, see [Update](#Update) process.
+
+#### Update
```
accordeiro, line 207: A general note about this section: one of the main reasons for updating a payment channel is to effectively make an off-chain payment between the two participants. 

This is clearly enabled by the mechanics described below (if I correctly understood you simply have to update the value on the `CREATE_CLAIMABLE_BALANCE` operation), but maybe it's worth making this more explicit? After all, I would guess this is the main functionality people are looking for when implementing this protocol.

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+
+The other participant can identify that the close process has started at an
+earlier state by monitoring changes in escrow account E's sequence. If the other
+participant sees the sequence number of escrow account E change to a value that
+is not the most recently used s_i, they can use the following process to contest
+the close. A participant contests a close by submitting a more recent
+declaration transaction and closing the channel at the actual final state.
+
+1. Get E's sequence number n
+2. If s_{e+1} >= n < s_i, go to step 3, else go to step 1
```
accordeiro, line 293: `s_{e+m}`

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+3. Submit most recent D_i
+4. Wait observation period O
+5. Submit C_i
```
accordeiro, line 296: Maybe it's worth adding a paragraph explaining why this works? Effectively you're submitting a transaction that bumps the sequence number to `s_i` which is higher than `s_{e+n}`, which invalidates the transaction the malicious participant submitted.

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+- [Withdraw](#Withdraw)
+
+##### Add Trustline
+
+Participants can add additional trustlines if they plan to make deposits of new balances.
+
+1. I and R sign and exchange signatures for trustline transaction TA_i.
+2. I or R submit TA_i.
+
+If the remove trustline transaction TA_i fails or is never submitted, there is
```
accordeiro, line 339: You probably meant **add** trustline here?

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+same value to participant I. Any amounts deposited to the payment channel
+without coordination will be disbursable to participant I at close.
+
+Participant R must coordinate with participant I to deposit any amount that it
+does not intend to immediately pay participant I. The participants use the
+following process:
+
+1. Increment i.
+2. I and R build the deposit transaction P_i.
+3. Set e' to e.
```
accordeiro, line 413: It took me a while to realize `e'` is an auxiliary var you're using, since all the other vars are well-defined concepts here. Maybe change the phrasing to `Store the current value for e on e'`?

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+
+##### Withdraw
+
+Participants must coordinate to withdraw an amount without closing the channel.
+The participants use the following process, where W is the participant
+withdrawing:
+
+1. Increment i.
+2. I and R build the withdrawal transaction W_i.
+3. Set e' to e.
```
accordeiro, line 454: Same comment about `e'` as above.

accordeiro (commented):
Took a first pass, this is looking pretty good, but we'd probably benefit from having more 👀 on the mechanics. I left a bunch of inline comments.

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+#### Setup
+
+To setup the payment channel:
+
+1. I creates the escrow account E.
+2. R creates the reserve account V.
+3. Set variable initial states:
+   - s to E's sequence number + 1.
+   - i to 0.
+   - e to 0.
```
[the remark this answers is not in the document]
leighmcculloch: I don't think so, because it's not a variable, it's a constant with a predefined value.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: The C tx needs to be the second tx in the tx set. s_i is the seq of the first tx in the tx set, that's why C is s_i+1.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: If we wanted to use m here it would be s_e+m, but the intention is to communicate the seq for the next iteration after e, which I think is clearer as it is. If m ever changed too it means the boosting here doesn't have to deal with that.

specifications/sep-payment-channel-mechanism.md
leighmcculloch:
```suggestion
3. Set e' to the value of e.
```

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Good call. Added in d653b78.

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,631 @@
...
+by R that will be used to sponsor the claimable balance ledger entries
+created at disbursement.  Created by R.  Jointly controlled by I and R while
+the channel is open.  Control is returned to R at close.  Cannot be merged
+until all claimable balances created at close for R are claimed by R.
+
+### Constants
+
+The two participants agree on the following constants:
+
+- m, the _maximum transaction count for an iteration's transaction set_, is
```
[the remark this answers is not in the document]
leighmcculloch: I think the definition of m is worthwhile because otherwise we have magic numbers, in this case 2, throughout the formulas. That was something I found difficult to understand in the original specification of the protocol in CAP-21.

---

I commented throughout about why I don't think it is suitable to use m in the other places.

> we should potentially replace `1` with `m`

Note that m has value 2.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Great call. I don't say why this works here at all. 24543b0

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Thanks! d4022e5

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Ah yes this is not particularly clear. Thanks. I used slightly different language in ff8ec83.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: ff8ec83

leighmcculloch:
Thanks @acharb @accordeiro. I've addressed all your comments. Please let me know if I didn't address any sufficiently. Could you please also review https://github.com/leighmcculloch/stellar--experimental-payment-channels/pull/1 which is intended to be merged into this PR?

specifications/sep-payment-channel-mechanism.md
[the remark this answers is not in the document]
accordeiro: Hm yeah I agree that's better than magic numbers. Makes sense to me!

specifications/sep-payment-channel-mechanism.md
accordeiro: Looks great!

specifications/sep-payment-channel-mechanism.md
accordeiro: Perfect, thanks!

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,657 @@
...
+
+- s, the _starting sequence number_, is initialized to one greater than the
+sequence number of escrow account E after E has been created. It is the first
+available sequence number for iterations to consume.
+
+- i, the _iteration number_, is initialized to zero.  It is incremented with
+every off-chain update of the payment channel state, or on-chain setup, deposit,
+withdrawal, etc.
+
+- e, the _executed iteration number_, is initialized to zero. It is incremented
```
acharb, line 144: ❓ I'm still not wrapping my head around why we need a formal variable e. It only increments when the payment channel is closed, at which point it doesn't matter

And therefore s_e is only either s, or the sequence # that E ends up at at the close

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,657 @@
...
+  account V's sponsorship.
+  - One or more `SET_OPTIONS` operation adjusting escrow account E's
+  thresholds to give I full control of E, and removing R's signers.
+  - One or more `SET_OPTIONS` operation adjusting reserve account E's
+  thresholds to give R full control of V, and removing I's signers.
+
+- D_i, the _declaration transaction_, declares an intent to execute the
+corresponding closing transaction C_i.  D_i has source account E, sequence
+number s_i, and `minSeqNum` set to s_e.  Hence, D_i can execute at any time, so
+long as E's sequence number n satisfies s_e <= n < s_i.  Because C_i has source
```
acharb, line 263: ❓ going off my s_e question, could't this just be
```suggestion
long as E's sequence number n satisfies s <= n < s_i.  Because C_i has source
```

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,657 @@
...
+3. Set variable initial states:
+   - s to E's sequence number + 1.
+   - i to 0.
+   - e to 0.
+5. I and R build the formation transaction F.
+6. Increment i.
+7. Sign and exchange a closing transaction C_i, that closes the channel with
+disbursements matching the initial contributions.
+8. Sign and exchange a declaration transaction D_i.
+9. I and R sign and exchange signatures for formation transaction F.
```
acharb, line 176: 💡 this wasn't initially obvious to me, wondering if it' worth adding a note about it
```suggestion
9. I and R sign and exchange signatures for formation transaction F.
    - it's important F is signed/exchanged after C_i and D_i in the case I or R disappear
```

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,657 @@
...
+
+  The `minSeqAge` and `minSeqLedgerGap` prevents a misbehaving party from
+  executing C_i when the channel state has already progressed to a later
+  iteration number, as the other party has the period O to invalidate C_i by
+  submitting D_i' for some i' > i.
+  
+  C_i contains operations:
+  - One `BEGIN_SPONSORING_FUTURE_RESERVES` operation that specifies reserve
+  account V as a sponsor of future reserves.
+  - One `CREATE_CLAIMABLE_BALANCE` operation for each trustline that is
```
acharb, line 251: 💡 it depends on if we go with the 2 escrow account idea from your [PR](https://github.com/leighmcculloch/stellar--experimental-payment-channels/pull/1), but if not then I think it's worth adding a point as to why this is a claimable balance op and not a payment op (to prevent a failed C_i)

maybe could be added in the `## Security Concerns` section instead

specifications/sep-payment-channel-mechanism.md
```diff
@@ -0,0 +1,657 @@
...
+the close. A participant contests a close by submitting a more recent
+declaration transaction and closing the channel at the actual final state. A
+more recent declaration transaction may be submitted because it has a higher
+sequence number than the declaration transaction that the malicious actor
+submitted. The more recent declaration transaction prevents the malicious actor
+from submitting the older closing transaction because it has a lower sequence
+number making that transaction invalid.
+
+1. Get E's sequence number n
+2. If s_{e+1} >= n < s_i, go to step 3, else go to step 1
```
acharb, line 319: would it be easier if we just said `2. If n != s_i go to step 3 because n is at a malicious value`, implying n is not at the value the latest D_i.seqNum?

the else statement of step 2 says go back to step 1 and re-get n. When would it differ?

acharb (commented):
Took another pass after understanding more and lgtm 🙌 

I had 1 question around the need for variable `e` and thus `s_e`. I'm not completely convinced they're needed but I could be missing something. 

If we'd like to name the sequence # that the channel closed at to reference posthumously, couldn't we use something like i_closed? Otherwise e just takes on 2 values 0 (start), and i_closed (channel closed), and I don't think we need a formal variable for i_closed since it's not used during the any of the processes (left the questions inline)

Also I had a couple inline 💡 for things that weren't initially obvious to me, and might be worth adding if it doesn't add tmi

specifications/sep-payment-channel-mechanism.md
leighmcculloch: e is set to the i of any transaction that the parties agree to submit to the network, so it is used in all the cases where the parties agree to modify the channel after setup, such as withdrawing.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: When e is zero, then s_e has the same value as s, but when e is greater than zero, it will have a greater value.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: For the majority of the channel lifespan n will not be equal to s_i because the escrow account's sequence number will be far in the past compared to the sequence in the last iteration. That's why we need it to be a range. It's okay if the sequence number isn't s_i, as long as it isn't earlier than s_i _and_ later than the last agreed sequence to execute.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Good call, I've added three new sections discussing this at relevant points in the doc. 86a02d0

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Good call, I added a paragraph in the security section where the closing transaction failure is discussed. This really relates to that because the use of a payment could cause the closing transaction to fail. a8ff8e7

leighmcculloch:
@acharb Thanks for the review. I've responded to or addressed your comments. I also applied the same changes to leighmcculloch#1 as well. Please reopen any comment threads if I didn't completely address a concern.

specifications/sep-payment-channel-mechanism.md
acharb: got it. I'm wondering if we can modify e's definition then, I think that's where my confusion stemmed from. What about

```
e, the executed iteration number, is initially set to zero, and thereafter represents the i of any transaction set the parties agree to submit to the network.
```

because it changes when the agreement happens. Not after an iteration is submitted on-chain (eg. in the deposit by R steps it changes before anything is submitted)

acharb: also, if I'm understanding this right I believe P_i should be bumping E.seqNum to s_e, not s_i. [In which case this line](https://github.com/stellar/experimental-payment-channels/pull/3/files#diff-b01917f0debe4bb9e19259c6e34888dd6c41d4f2da28f49af7fc820ba5a38dc7R478) would be 
```
One BUMP_SEQUENCE operation bumping the sequence number of escrow account E to s_e.
```

leighmcculloch: Ah good point, the description doesn't really say what was intended. Updated in a0b735f. Thank you!

leighmcculloch: > also, if I'm understanding this right I believe P_i should be bumping E.seqNum to s_e, not s_i. [In which case this line](https://github.com/stellar/experimental-payment-channels/pull/3/files#diff-b01917f0debe4bb9e19259c6e34888dd6c41d4f2da28f49af7fc820ba5a38dc7R478) (and the other changing channel setup cases) would be

I think they're the same thing essentially. In the deposit flow in step 2 when P_i is built s_i will be the sequence number of the escrow account we want to bump to. s_e is not currently set to the value we want to bump to, but it will be after in step 4.

acharb: ah good point. When I read the construction of P_i it was after going through the steps. At the point P_i is created in your steps s_i is right 👍

nikhilsaraf (commented):
LGTM

PR #10, opened 2021-05-05
Add more details to README about the forks

leighmcculloch:
### What
Add more details to the README about what features are implemented in the forks.

### Why
I keep forgetting which preconditions are implemented and it is helpful to knowledge share this.


acharb (approved):
LGTM!

PR #19, opened 2021-05-17
SEP: Add change trust ops to formation

leighmcculloch:
### What
Add change trust operations to the formation transaction.

### Why
Even though the escrow accounts will likely have all the trustlines required because the setup step involves adding the initial contributions, it is still really important that the change trust operations exist in the formation transaction otherwise an asset with a zero balance in either account but that the closing transaction will disburse could leave the channel in a broken state. A participant could remove a trustline just before the formation transaction is executed. The responder could do this undetected because their escrow accounts sequence number has no bearing on the formation transaction.

I removed the trustlines in 5b8afc1fbe195ab7032c08eb798d84f2cdb30fbb but overlooked this issue.

### DIFF
--- a/specifications/sep-payment-channel-mechanism.md
+++ b/specifications/sep-payment-channel-mechanism.md
@@ -188,6 +188,7 @@ multisig accounts. F has source account E, and sequence number set to s_i.
   - Operations sponsored by I:
     - One `BEGIN_SPONSORING_FUTURE_RESERVES` operation that specifies
     participant I as a sponsor of future reserves.
+    - One or more `CHANGE_TRUST` operations configuring trustlines on EI.
     - One `SET_OPTIONS` operation adjusting escrow account EI's thresholds such
     that I and R's signers must both sign.
     - One or more `SET_OPTIONS` operations adding I's signers to ER.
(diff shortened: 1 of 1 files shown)


specifications/sep-payment-channel-mechanism.md
```diff
@@ -196,12 +197,22 @@ multisig accounts. F has source account E, and sequence number set to s_i.
...
+    - One or more `CHANGE_TRUST` operations configuring trustlines on ER.
     - One `SET_OPTIONS` operations adjusting escrow account ER's thresholds such
     that R and I's signers must both sign.
     - One or more `SET_OPTIONS` operations adding R's signers to EI.
     - One `END_SPONSORING_FUTURE_RESERVES` operation that stops R sponsoring
     future reserves of subsequent operations.
   
+  The escrow accounts EI and ER will likely have all the necessary trustlines
+  before the formation transaction is built. This means the `CHANGE_TRUST`
+  operations will likely be no-ops. The `CHANGE_TRUST` operations must be
```
acharb, line 209: does no-op just mean an operation that has no effect? If so makes sense, just double checking

acharb (approved):
nice catch 💯 . LGTM
just had 1 probably non-blocking Q

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Yup.

leighmcculloch: I don't mind adding more length language to state that, but I noticed CAP-21 uses the same terminology, so I copied it.

PR #20, opened 2021-05-17
SEP: Add security sections about auth required and clawback

leighmcculloch:
### What
Add two sections about auth required and clawback and how they can be security concerns.

### Why
Both seems like relatively important topics to discuss. The choice of assets used in the payment channel can compromise the entire payment channel, not just the portion of the channel using that asset.

cc @stanford-scs

### DIFF
--- a/specifications/sep-payment-channel-mechanism.md
+++ b/specifications/sep-payment-channel-mechanism.md
@@ -575,7 +575,8 @@ possible without the participants coordinating.
 
 The closing transaction, C_i, must never fail.  Under the conditions of the
 Stellar Consensus Protocol as it is defined today, and under correct use of this
-protocol, there is no known conditions that will cause it to fail.  It will be
+protocol, and assuming no changes in the authorized state of the channels
+trustlines, there is no known conditions that will cause it to fail.  It will be
 either invalid or valid and successful, but not valid and failed.  If C_i was to
 be valid and fail it would consume a sequence number and fair distribution of
 the assets within the escrow account would require the cooperation of all
@@ -593,7 +594,7 @@ account has on making a payment, due to liabilities, or would exceed limits on
 the receiving account, such as a trustline limit. Participants must ensure that
 the payments they sign for are receivable by the escrow accounts.
 
-### Auth Required
+### Trustline Authorization
 
 Any trustlines on the escrow accounts that have been auth revoked, or could be
 auth revoked, could compromise the payment channels ability to close
@@ -603,7 +604,7 @@ If the issuer of any auth revocable asset submits an allow trust operation
 freezing the amounts in either escrow account, the close transaction may fail to
 process if its payment operations are dependent on amounts frozen.
 
-There is nothing participants can do to prevent this, other than only using auth
+There is nothing participants can do to prevent this, other than using only auth
 immutable assets.
 
 ### Clawback
--- a/specifications/sep-payment-channel-mechanism.md
+++ b/specifications/sep-payment-channel-mechanism.md
@@ -593,6 +593,36 @@ account has on making a payment, due to liabilities, or would exceed limits on
 the receiving account, such as a trustline limit. Participants must ensure that
 the payments they sign for are receivable by the escrow accounts.
 
+### Auth Required
+
+Any trustlines on the escrow accounts that have been auth revoked, or could be
+auth revoked, could compromise the payment channels ability to close
+successfully.
+
+If the issuer of any auth revocable asset submits an allow trust operation
+freezing the amounts in either escrow account, the close transaction may fail to
+process if its payment operations are dependent on amounts frozen.
+
+There is nothing participants can do to prevent this, other than only using auth
+immutable assets.
+
+### Clawback
+
+Any trustlines on the escrow accounts that have clawback enabled could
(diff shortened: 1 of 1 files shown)


specifications/sep-payment-channel-mechanism.md
```diff
@@ -593,6 +598,36 @@ account has on making a payment, due to liabilities, or would exceed limits on
 the receiving account, such as a trustline limit. Participants must ensure that
 the payments they sign for are receivable by the escrow accounts.
 
+### Trustline Authorization
+
+Any trustlines on the escrow accounts that have been auth revoked, or could be
+auth revoked, could compromise the payment channels ability to close
```
accordeiro, line 604:
```suggestion
auth revoked, could compromise the payment channel's ability to close
```

specifications/sep-payment-channel-mechanism.md
```diff
@@ -593,6 +598,36 @@ account has on making a payment, due to liabilities, or would exceed limits on
...
+
+There is nothing participants can do to prevent this, other than using only auth
+immutable assets.
+
+### Clawback
+
+Any trustlines on the escrow accounts that have clawback enabled could
+compromise the payment channels ability to close successfully.
+
+If the issuer of any clawback enabled trustline submits a clawback operation for
```
accordeiro, line 619: Could we treat a clawback as if it were a withdrawal operation?

accordeiro (approved):
These changes look right to me, nice catch! 

I do wonder if there's an alternative solution at a different layer, as having to reason about all these exceptions when establishing a channel seems pretty tricky.

I also imagine we tend to have more regulated assets (SEP-8 enabled) over time, so we should probably figure out a story for those.

specifications/sep-payment-channel-mechanism.md
leighmcculloch: Conceptually they're the same, but a withdrawal operation under normal circumstances requires both parties to sign new C_i and D_i transactions before the withdrawal transaction is signed. The signing of C_i and D_i first is required so that the channel has a valid closing state prepared before the withdrawal changes the state of the channel. There would be no way to guarantee that the participants would know they needed to sign C_i and D_i before a clawback occurred. We could define in a new section what participants should do if a clawback occurs.

However, I think it would be difficult to prescribe what they would do in all cases where clawback occurred. There are are three scenarios I see, assuming clawback of a single asset only:

1. The clawed back amount is an amount that belongs to both parties. I.e. the latest closing transaction at that time makes a payment using the amount. In this case both parties have a loss, but the amount that both parties lose isn't undefined. If EI has $100, $50 owed to ER, and $75 is clawed back, how should the loss be apportioned?

2. The clawed back amount is an amount that belongs to both parties, but the clawed back amount doesn't break the closing transaction. I.e. the latest closing transaction at that time makes a payment using the amount, but the escrow account clawed back from still has sufficient balance of that asset to make the closing transaction payment. This has similar properties are option 1, but at least the channel isn't broken. The closing transaction will still work, but it's possible the issuer knows about the channel and that an amount is owed to the other participant and actually wants to take from them.

3. The clawed back amount is an amount that belongs to a single party. I.e. the latest closing transaction at that time makes no payments using the amount. In this case it is clear only one party is at a loss, and in this case it operates like a withdrawal, assuming all parties see it occur. In this case the channel isn't broken actually, which is nice. However, if one party sees it occur before the other, they may trick the other party into accepting a payment for the amount. Participants are not required to monitor the blockchain constantly. They need only monitor the blockchain once per observation period.

The channel is technically only broken in scenario 1. In scenario 2 the channel can still close, but we don't know if the closure will be fair. In scenario 3 the channel can still close and it will be fair.

leighmcculloch:
>I do wonder if there's an alternative solution at a different layer, as having to reason about all these exceptions when establishing a channel seems pretty tricky.
>
>I also imagine we tend to have more regulated assets (SEP-8 enabled) over time, so we should probably figure out a story for those.

I think regulated assets scoped to what is defined in SEP-8 is mostly independent to payment channels. Two participants could involve a regulated asset server in the auth and deauth of their escrow accounts by having the regulated asset server sign every closing transaction. Nothing would need to change about SEP-8 or the channel protocol to support this today, except for the inclusion of allow trust operations within the setup, formation, and closing transactions. If the regulated asset server disagrees with a payment channel payment it would be just like one party disagreed (which we don't discuss what happens there, and I will add that to the SEP: https://github.com/stellar/experimental-payment-channels/issues/27).

The place where regulated assets fall apart is if the state of the issuer account changes, such as the signers change, or the account is merged. In both cases the closing transactions would fail, so regulated assets would only be safe if the high threshold is set higher than all the signers combined have weight for. This would disallow merging and changing of signatures, but continue to allow the allow trust operation.

Even if all the above is correct, an issuer using regulated assets who is using clawback could still break the channel.

So, in summary:
- Clawback is incompatible.
- Auth immutable is required, unless the closing transactions are signed by the issuer with allow trust operations, and the issuer has its high threshold locked.
- Regulated assets can be supported, assuming the above is true.

leighmcculloch:
@accordeiro Let's discuss the regulated assets issues in https://github.com/stellar/experimental-payment-channels/discussions,

PR #21, opened 2021-05-18
SDK API Design

leighmcculloch:
@nikhilsaraf I took the API design you shared in Google Docs, and converted it into a skeleton verbatim, that's the first commit. I find it helpful to view new APIs using the godoc command to give me an immediate experience of what it will be like to browser the API docs for that code.

If you install the [godoc](http://golang.org/x/tools/cmd/godoc) command locally you can view it also in web form. Or if you simply use the built in `go doc -a` command that'll give you a text version, albeit not so pretty and easy on the eyes.

cc @accordeiro @acharb @shanzzam

### DIFF
--- a/apidesign/api.go
+++ b/apidesign/api.go
@@ -1,0 +1,149 @@
+package apidesign
+
+import "time"
+
+type ObservationParams struct{}
+
+type Asset struct{}
+
+type ChannelFactory struct{}
+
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Here's an example of what I see of the function surface:
Screen Shot 2021-05-17 at 8 53 03 PM

PR #22, opened 2021-05-18
SDK API Design – Suggested Changes

leighmcculloch:
@nikhilsaraf I played around with your API design in #21, and modified some of the functions. 

The main changes are:
 - Replace the factory with a new function.
 - Replace the callbacks with a dependence on the calling code to call a check function frequently.
 - Replace the update function with a pay function.
 - Removed most timeouts, I think we can use `context` for that, and/or make it part of the config. Although, I left it out for the moment.
 - I organized the balances a little differently.

The balance changes shift the focus on storing what is owing between each party, instead of tracking the balances. The reason for this is that deposits can occur out-of-band and it will be difficult to lock-step update balances while payments are occurring. Storing an owing amount will make it easier to separate them.

I removed the factory because factories aren't too common in Go code, albeit a new function, which is essentially a factory, is.

I think we can still add the callbacks in if we need them, but that could be a type that sits atop this lower level channel. For example, the start monitoring service could be another type that the user chooses to instantiate to manage the channel for them.

What do you think?

cc @accordeiro @acharb @shanzzam

### DIFF
--- a/apidesign/api.go
+++ b/apidesign/api.go
@@ -2,148 +2,73 @@ package apidesign
 
 import "time"
 
-type ObservationParams struct{}
-
 type Asset struct{}
 
-type ChannelFactory struct{}
-
-type ChannelConnection struct {
-	ChannelID                string
-	Status                   string // waiting | active | closed_collaborative | closed_hostile
-	Asset                    Asset
-	InitiatorEscrowAccount   string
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Here's an example of what I see in godoc:
Screen Shot 2021-05-17 at 8 56 42 PM

PR #23, opened 2021-05-18
Add example using Go SDK

### LINKED ISSUE #7: Implement basic examples using the Go SDK
Implement some basic examples using the Go SDK, once it supports CAP-21 (see #4).

leighmcculloch:
### What
Add an example using the Go SDK to implement direct two-way payment channels in-process without any network connectivity.

### Why
For a number of reasons:
- As a starting point in understanding the type of code we will need to write.
- As a proof that the protocol works.
- To gain a better understanding of what data we aren't thinking about that we might need to store.
- To get a better understanding of gaps in the SEP.
- To get an idea how we might build a test framework that can execute different scenarios in-process and test outcomes.

Close #7

This PR is a WIP.


examples/02-using-go-sdk/friendbot.go
```diff
@@ -0,0 +1,42 @@
+package main
+
+import (
+	"github.com/stellar/go/clients/horizonclient"
+	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/txnbuild"
+)
+
+func friendbot(account *keypair.FromAddress, startingBalance string) error {
```
acharb, line 9: love this friendbot 🙌

sdk/txbuild/formation.go
```diff
@@ -0,0 +1,42 @@
...
+		Timebounds: txnbuild.NewTimeout(300),
+	}
+	tp.Operations = append(tp.Operations, &txnbuild.BeginSponsoringFutureReserves{SourceAccount: initiator.Address(), SponsoredID: initiatorEscrow.Address()})
+	tp.Operations = append(tp.Operations, &txnbuild.SetOptions{
+		SourceAccount:   initiatorEscrow.Address(),
+		MasterWeight:    txnbuild.NewThreshold(0),
+		LowThreshold:    txnbuild.NewThreshold(2),
+		MediumThreshold: txnbuild.NewThreshold(2),
+		HighThreshold:   txnbuild.NewThreshold(2),
+		Signer:          &txnbuild.Signer{Address: responder.Address(), Weight: 1},
```
acharb, line 24: should the formation transaction also require the initiator to be added as a signer to EI (and responder to ER) with weight of 1? Eg. if before `F` is submitted `I` changes their signing weight to 0 on `EI`, then `F` is submitted. That would lock `EI` right? Or `I` changes their signing weight to 2

another option could be part of validating F for responder is checking EI has initiator with weight 1.

leighmcculloch: That's a good point. This assumes I and R haven't messed things up by removing themselves as signers. Which is not so bad if the initial contributions are zero, but if they are zero, it is bad. You could trap someone that way. I'll change it here and in the SEP too.

leighmcculloch:
This still needs much more work, and is pretty messy, but the txbuild package this example introduced is useful to more than just me now so I'm going to merge it, and we can continue to iterate.

Issue #24, opened 2021-05-18
sdk: add top-level api skeleton


leighmcculloch:
Add an SDK skeleton made up of empty functions and basic types.

The skeleton should focus on functions and types required for the state machine portion of the SDK.

No need for it to be perfect, the intention is we use it to iterate.

PR #25, opened 2021-05-18
sdk: add skeleton

acharb:
SDK skeleton based on the current state of the [sdk doc](https://docs.google.com/document/d/10EuAvYm7j3IKpSHDEsCJsmG7q7VsALio_b2P8h5skPo/edit#)

will update this skeleton when the doc is updated from the first sdk meeting

closes https://github.com/stellar/experimental-payment-channels/issues/24


sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+
+}
+
+func NewChannelListenerService() {
+
+}
+
+func GetChannel() {
+
+}
```
leighmcculloch, line 17: I think `GetChannel` was only required when the factory existed as well. I think you can remove it.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+
+//// Channel State Machine
+
+type ChannelStateMachine struct {
+	ChannelID              string
+	Status                 string // waiting | active | closing | closed
+	Asset                  Asset
+	InitiatorEscrowAccount string
+	ResponderClaimAmount   int
+	ReserveEscrowAccount   string
```
leighmcculloch, line 27: The SEP doesn't have a reserve account any longer, but two mostly equal escrow accounts. The only thing that differs between the two is which escrow account is used for sequencing.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func GetChannel() {
+
+}
+
+//// Channel State Machine
+
+type ChannelStateMachine struct {
+	ChannelID              string
```
leighmcculloch, line 22: Let's drop the channel ID. I don't see how that would be used internally at the moment.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (c *ChannelStateMachine) GetLatestCloseTx(id string) (*TxInfo, error) {
+	return nil, nil
+}
+
+type TxInfo struct {
+	ID               string
+	Type             string // declaration | close
+	Seq              int64
```
leighmcculloch, line 75: I think we should include the iteration number here too.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+	ChannelID              string
+	Status                 string // waiting | active | closing | closed
+	Asset                  Asset
+	InitiatorEscrowAccount string
+	ResponderClaimAmount   int
+	ReserveEscrowAccount   string
+	InitiatorClaimAmount   int
+	IsInitiator            bool
+	MyAccount              string
+	MyClaimAMount          int
```
leighmcculloch, line 31:
```suggestion
	MyClaimAmount          int
```

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+func (c *ChannelStateMachine) GetLatestCloseTx(id string) (*TxInfo, error) {
+	return nil, nil
+}
+
+type TxInfo struct {
+	ID               string
+	Type             string // declaration | close
+	Seq              int64
+	InitiatorBalance int
+	ResponderBalance int
```
leighmcculloch, line 77: The transactions won't capture balances, but rather amount owing to each other which will only be a portion of the balances. @nikhilsaraf What's the goal of these two fields? I'm wondering if we need them, or need something else.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+	return nil
+}
+
+func (cls *ChannelListenerService) GetPendingChannelCreationRequests() {
+	return nil
+}
+
+func (cls *ChannelListenerService) RespondChannelCreationRequest(channelID string, response ChannelCreationResponse, timeoutMillis int) (*ChannelConnection, error) {
+	return nil, nil
+}
```
leighmcculloch, line 125: I think we can leave the listener type out for the moment and focus on the state machine for the moment.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,126 @@
...
+	OtherClaimAmount       int
+}
+
+func (c *ChannelStateMachine) Init() {
+	return nil
+}
+
+func (c *ChannelStateMachine) RegisterMonitoringNotificationHandler(handler MonitoringNotificationHandler) error {
+	return nil
+}
```
leighmcculloch, line 42: I think we should leave this out while we're focused on the state machine. Any handlers sound like orchestration.

leighmcculloch (commented):
Some initial thoughts inline based off our discussion today.

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/a7322ba7b51ff9ecf88e807adc9401641821a8a3

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/a7322ba7b51ff9ecf88e807adc9401641821a8a3

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/a7322ba7b51ff9ecf88e807adc9401641821a8a3

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/a7322ba7b51ff9ecf88e807adc9401641821a8a3

sdk/sdk-skeleton.go
acharb: right, I think ResponderEscrowAccount is more consistent with the docs
changed: https://github.com/stellar/experimental-payment-channels/pull/25/commits/bbbff035f84586fa306758c6eb2d02c9ecc4719b

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/ce73653f4ebdd5884d37aa91946d147fe348610a

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/ce73653f4ebdd5884d37aa91946d147fe348610a

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,77 @@
...
+
+type ChannelStateMachine struct {
+	Status                 string // waiting | active | closing | closed
+	Asset                  Asset
+	InitiatorEscrowAccount string
+	ResponderEscrowAccount string
+	InitiatorClaimAmount   int
+	ResponderClaimAmount   int
+	IsInitiator            bool
+	MyAccount              string
```
acharb, line 23: Do we need `MyClaimAmount` and `OtherClaimAmount`. Aren't those describing `InitiatorClaimAmount` and `ResponderClaimAmount`?
cc @nikhilsaraf

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,77 @@
...
+
+//// Methods directly on the SDK
+
+func NewChannel() {
+
+}
+
+func NewChannelListenerService() {
+
+}
```
leighmcculloch, line 11: This can be removed since there is no listener service yet.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (c *ChannelStateMachine) CheckChannel() (*ChannelCheckResponse, error) {
+	return nil, error
+}
+
+func (c *ChannelStateMachine) Payment(sendAmount int, timeoutMillis int) error {
+	return nil
+}
+
+func (c *ChannelStateMachine) CloseStart(declareID int) error {
```
leighmcculloch, line 41:
```suggestion
func (c *ChannelStateMachine) CloseStart(iterationNumber int) error {
```

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (c *ChannelStateMachine) Init() {
+	return nil
+}
+
+func (c *ChannelStateMachine) CheckChannel() (*ChannelCheckResponse, error) {
+	return nil, error
+}
+
+func (c *ChannelStateMachine) Payment(sendAmount int, timeoutMillis int) error {
```
leighmcculloch, line 37: Let's deal with timeouts in another way. We will probably add context's to all the functions that need it. We can think about that later though.
```suggestion
func (c *ChannelStateMachine) Payment(sendAmount int) error {
```

sdk/sdk-skeleton.go
acharb: `PayoutAmount`?

leighmcculloch: `BalanceOwing` is maybe another way to think about it, because the amount in the payment won't be the total payout amount, but rather just what doesn't already exist in the recipients account.

sdk/sdk-skeleton.go
leighmcculloch: Regarding the amounts I think there's two different things we need to know about:
- A cache of the last known balance of the escrow accounts. This value may go up without us knowing, but it will only go down with coordination. These amounts need to go down immediately for a signed withdrawal, but the balance of the escrow account won't actually go down immediately, and may never go down if the withdrawal fails, so there's some complexity we need to figure out.
- The amount that each participant owes the other participant. Payments should move these values. We could instead store the total amount paid in both directions and calculate the amount owing as it is needed, but that approach would be riskier for overflows.

sdk/sdk-skeleton.go
```diff
@@ -0,0 +1,77 @@
...
+	Asset                  Asset
+	InitiatorEscrowAccount string
+	ResponderEscrowAccount string
+	InitiatorClaimAmount   int
+	ResponderClaimAmount   int
+	IsInitiator            bool
+	MyAccount              string
+	MyClaimAmount          int
+	OtherAccount           string
+	OtherClaimAmount       int
```
leighmcculloch, line 26: We have two different perspectives in this type's field names:
- Initiator/Responder are objective.
- My/Other are subjective.

The differing perspectives are confusing and difficult to think about, at least for me.

I see a few ways we could represent this data. Ideally we pick only one, if possible:

1. Initiator / Responder, objective. The `IsInitiator` field would signal which the caller would be. This will result in `if`s in the code wherever the user is making an action to select the right escrow account, etc.
2. My / Other, subjective. This probably makes some of the functionality simpler because it operates immediately on the appropriate object. We'd still need an indicator for who is the initiator, for selecting the sequencing number.
3. ParticipantA / ParticipantB, objective, and disconnected from roles. Roles assigned.

It might be useful for us to group all the fields relating to a single participant into a Participant type.

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/462d4215919a3528d809eed806adc7360e4f9534

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/1162f242ff89332a2fa07c4d776588c150464a2b

sdk/sdk-skeleton.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/25/commits/71a7168df8ac07dc6ac331728a5910c64328add3

sdk/sdk-skeleton.go
acharb: I'm initially leaning towards My/Other. I think this is the most intuitive for the user of the sdk. Eg. a user has many payment channels going at once, it would be hard to remember which ones they are an Initiator vs Responder of w/o having to check IsInitiator all the time. Once the channel is live it doesn't really matter who Initiator vs Responder is 

Also not having to add `if` statements everywhere is nice 😄

leighmcculloch: Just thought of one more: Local/Remote. 

I think My / Other sounds good though. Let's use it.

There will still need to be ifs around IsIniator to select the appropriate sequencing account, but I think that's fine.

acharb:
oops I actually meant to put this commit in a separate PR: https://github.com/stellar/experimental-payment-channels/pull/25/commits/2b265a2c47704ff9a0f20eb3cd43191db80b2349

But this change is based off this convo: https://stellarfoundation.slack.com/archives/C0166LZF4VC/p1621448846008000

I can revert if we decide against

sdk/sdk-skeleton.go
nikhilsaraf: `MyClaimAmount` and `OtherClaimAmount` is intended to only be a convenience method. i.e. the SDK could be either the Initiator or Responder. The `is_initiator` bool allows them to check and pick the right account, so the `MyClaimAmount` just makes this easier -- this should probably be a function instead of a variable on the struct

sdk/sdk-skeleton.go
nikhilsaraf: I think the SDK should save both perspectives internally (my/other probably as functions instead of variables maybe?). This allows the SDK to be cognizant of the Initiator/Responder as it needs to deal with the details of the channel. However, we can decide what API to present to the user, and I agree am leaning to My/Other (or Local/Remote) since that's a better API for the SDK consumer

sdk/sdk-skeleton.go
nikhilsaraf: Yes, agreed. I was thinking this would be better as a function instead of as a variable, mostly for convenience to the SDK user so they can just call MyBalance() rather than doing the calculations themselves using the internals of this struct.

nikhilsaraf (commented):
looks good as a skeleton.

sdk/sdk-skeleton.go
acharb: moved to helper methods:
https://github.com/stellar/experimental-payment-channels/pull/25/commits/5bd4f54a1d3e26e434e2f21521191fde56c108d9

sdk/sdk-skeleton.go
acharb: moved to helper method:
https://github.com/stellar/experimental-payment-channels/pull/25/commits/5bd4f54a1d3e26e434e2f21521191fde56c108d9

sdk/sdk-skeleton.go
acharb: sounds good, leaving both perspectives for now then. If we decide to change that later it can be in a PR

acharb:
going to merge this in as is. Of course we can iterate it on it as we start doing implementation

PR #28, opened 2021-05-19
add channel state diagram

acharb:
Adds a state diagram to describe how the channel state can progress

FYI - I'm using the current SDK methods to describe the actions. I think that's the best way to describe what the user would do (but can make more generic if needed)


specifications/diagrams/state-diagrams/channel-state.txt
```diff
@@ -0,0 +1,12 @@
+stateDiagram-v2
+    [*] --> Waiting: NewChannel()
+    note left of [*]: Initiator opening a channel
+
+    Waiting --> Active: channel.Open()
+
+    Active --> Closing: channel.CloseStart()
+    
+    Closing --> Closed: Responder calls <br>channel.CloseCoordinated()
+    Closing --> Closed: channel.CloseUnCoordinated()
```
leighmcculloch, line 10: At a high-level I think this is right. I think the failure cases of the channel though are interesting and I wonder if we should capture them here or in another diagram. Most steps, if agreement isn't found, the participants should close the channel.

leighmcculloch (approved):
Looks great. :shipit: 

I also left a thought inline, but it doesn't need anything doing to this PR.

PR #29, opened 2021-05-19
SEP: Simplify the steps during setup

leighmcculloch:
### What
Move the building of the formation transaction to after the first close and declaration transactions are signed and exchanged.

### Why
It was only built first because the formation transaction's sequence number was defined as `s_i`, but there is only ever one formation transaction, and it is always `s_0`, so it can be defined simply as `s` and it can be built after `C_i` and `D_i`. This makes the process simpler both to understand, and to reduce in implementations since the agreement about the initial state can be entirely independent of the formation step.


acharb (approved):
lgtm!

PR #31, opened 2021-05-20
in formation transaction ensure other party is signer and weight of 1

acharb:
ensure the other party is a signer and has an equal weight for each escrow account. This ensures the Escrow account is in the right state for the channel, w/o relying on the other party to do it correctly themselves

### DIFF
--- a/sdk/txbuild/formation.go
+++ b/sdk/txbuild/formation.go
@@ -31,6 +31,10 @@ func Formation(p FormationParams) (*txnbuild.Transaction, error) {
 		HighThreshold:   txnbuild.NewThreshold(2),
 		Signer:          &txnbuild.Signer{Address: p.ResponderSigner.Address(), Weight: 1},
 	})
+	tp.Operations = append(tp.Operations, &txnbuild.SetOptions{
+		SourceAccount: p.InitiatorEscrow.Address(),
+		Signer:        &txnbuild.Signer{Address: p.InitiatorSigner.Address(), Weight: 1},
+	})
 	tp.Operations = append(tp.Operations, &txnbuild.EndSponsoringFutureReserves{SourceAccount: p.InitiatorEscrow.Address()})
 	tp.Operations = append(tp.Operations, &txnbuild.BeginSponsoringFutureReserves{SourceAccount: p.ResponderSigner.Address(), SponsoredID: p.ResponderEscrow.Address()})
 	tp.Operations = append(tp.Operations, &txnbuild.SetOptions{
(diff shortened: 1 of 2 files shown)


sdk/txbuild/formation.go
```diff
@@ -31,6 +31,10 @@ func Formation(p FormationParams) (*txnbuild.Transaction, error) {
 		HighThreshold:   txnbuild.NewThreshold(2),
 		Signer:          &txnbuild.Signer{Address: p.ResponderSigner.Address(), Weight: 1},
```
acharb, line 32: in [the SEP Reserves Section](https://github.com/stellar/experimental-payment-channels/blob/main/specifications/sep-payment-channel-mechanism.md#reserves) we say `I` is responsible for sponosring the reserves of `I` only. But here `I` sponsoring `R` for `EI`. I think this makes sense, instead of creating another Sponsor operation sandwich for `ER` sponsored by `I`. Since it evens out anyways and the escrow accounts eventually will be destroyed

Maybe a better way of saying it is "I is responsible for sponsoring the reserves of EI"?

specifications/sep-payment-channel-mechanism.md
```diff
@@ -194,7 +194,7 @@ multisig accounts. F has source account E, and sequence number set to s.
     - One or more `CHANGE_TRUST` operations configuring trustlines on EI.
     - One `SET_OPTIONS` operation adjusting escrow account EI's thresholds such
     that I and R's signers must both sign.
-    - One or more `SET_OPTIONS` operations adding I's signers to ER.
+    - One or more `SET_OPTIONS` operations adding I and R as equal signers to ER.
```
leighmcculloch, line 197: Subtle distinction here that we want to add I and R's signers to each escrow account, not necessarily I and R specifically. I and R may have multi-sig accounts. The code doesn't do that yet, nor do we need to prioritize the code doing that yet, but the SEP can speak into that future.
```suggestion
    - One or more `SET_OPTIONS` operations adding I and R's signers to ER.
```

leighmcculloch (approved):

sdk/txbuild/formation.go
leighmcculloch: Oh hmm, this is a bug in this code.

leighmcculloch: It evens out unless I and R have different numbers of signers. We shouldn't need new sponsor operations, we just need to reorder the inner operations into the other blocks I think?

specifications/sep-payment-channel-mechanism.md
acharb: I left the word `equal` in there as well. I feel like that's important to point out since ensuring their equal (and the other party didn't secretly boost their weight of the escrow account) is one of the reason to add both signers and not just 1
https://github.com/stellar/experimental-payment-channels/pull/31/commits/d30bcd9972d949b649059822162df31f42c6ec62

leighmcculloch: I need to think about the equal part some more. They may not be equal weight, even though the weights allow for equal control. The line further up is the important bit:
>adjusting escrow account EI's thresholds such that I and R's signers must both sign.

sdk/txbuild/formation.go
acharb: true I didn't think about if I and R have more than 1 signer, in that case the wording makes sense as is then

specifications/sep-payment-channel-mechanism.md
acharb: hermm ok. I guess I assumed they'd always be equal, but I guess doesn't have to be. I'll remove
https://github.com/stellar/experimental-payment-channels/pull/31/commits/9a8eb1f380ea4614e4471895c73d237eadfba8c5

sdk/txbuild/formation.go
acharb: gonna merge since the example code change can be committed separately

PR #32, opened 2021-05-20
sdk: add test

leighmcculloch:
### What
Add a test to the sdk that tests the existing tx building functions against a running CAP-21 forks of Horizon and Core.

### Why
So that we can ensure we maintain some stability and basic behavior as we're experimenting and starting to build out the API.

The test is a high-level integration test because unit tests for tx building would likely involve building the tx in the test and in the code and checking they are equal, which is not particularly useful at this stage.

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -8,28 +8,25 @@ on:
 
 jobs:
 
-  quickstart:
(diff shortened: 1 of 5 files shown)


acharb (approved):
lgtm 👍

PR #37, opened 2021-05-21
sdk: use fee bumps

### LINKED ISSUE #36: sdk: update test to use fee bumps
### What
Update the test to use fee bumps.

### Why
The SEP defines using fee bumps transaction envelopes so that the escrow accounts never need to account for fee amounts taken from the account which may be unpredictable during surge times.

The txbuild test doesn't use fee bumps, but it should be to be closer to what a channel will actually use.

leighmcculloch:
### What
Set the fee to zero in txbuild built transactions, and require the use of fee bumps by a participant when submitting a tx.

### Why
The SEP specifies that fee bumps should be used. The reason is to keep the balances of the escrow accounts unaffected be fees which makes accounting simpler.

Close #36

### DIFF
--- a/sdk/go.mod
+++ b/sdk/go.mod
@@ -2,7 +2,7 @@ module github.com/stellar/experimental-payment-channels/sdk
 
 go 1.16
 
-replace github.com/stellar/go => github.com/leighmcculloch/stellar--go v0.0.0-20210514225055-eafef939d8d2
+replace github.com/stellar/go => github.com/leighmcculloch/stellar--go v0.0.0-20210521212547-6e7c3ba30f97
 
 require (
 	github.com/stellar/go v0.0.0-00010101000000-000000000000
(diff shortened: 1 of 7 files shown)


sdk/go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/stellar/experimental-payment-channels/sdk
 
 go 1.16
 
-replace github.com/stellar/go => github.com/leighmcculloch/stellar--go v0.0.0-20210514225055-eafef939d8d2
+replace github.com/stellar/go => github.com/leighmcculloch/stellar--go v0.0.0-20210521212547-6e7c3ba30f97
```
leighmcculloch, line 5: This update was required to pull in stellar/go#3622.

acharb (approved):
lgtm 👍

PR #38, opened 2021-05-21
examples: delete go sdk example

leighmcculloch:
### What
Delete the Go SDK example.

### Why
It was converted into a test in #32. Since the example doesn't run on each commit like a test does, it's quickly getting out-of-date. 

Learning is we should write our examples as tests, which is partly why we ported the example to the test in the first place.


acharb (approved):
clean uppp. Lgtm 👍

PR #42, opened 2021-05-24
specifications/diagrams/state-diagrams: add proposal diagrams

acharb:
state diagrams for proposing a channel update

1. payment-proposal.txt
2. channel-proposal.txt

After creating diagrams some thoughts:
- opening a channel is a special case of updating channel state, so an abstraction for `Proposing/Accepting` could make sense
- communication should always be the pattern `Proposing/Accepting` so I used state names `Proposing, ResponseReceived, ProposalAccepted, etc.` here instead of more general `Sending/Receiving`

note: I made the assumption that connection/handshake happens first, and then messages are sent, as opposed to a handshake for each message


nikhilsaraf:
This is really helpful to understand, thanks for putting together Alec! I think you've done a brilliant job to capture the various types of messages we can send/receive and how they fit together.

I've made some suggestions below to help accommodate the multi-threaded nature of the SDK and provide an event-driven state machine. This is mostly restructuring some of the boxes to represent parallel threads but the basic logic is the same. Imo, this makes it easier to map directly to code, but if this representation does not work for you'll I'm happy to find an alternative. I think we should have a meeting to discuss the logic behind the state diagram in this PR and come to an agreement of how we want the ChannelStateMachine to function (and the various states) so we can use that as the foundation for moving forward on development.

What do you'll think @acharb @leighmcculloch?

Adding suggestions on diagram here:

1. The "Listening" state should be in a background thread. We should represent this with a dotted line to denote a new thread starting off in the "Listening" state. Before the "Idle" state we should have a dotted line going to the right showing this "Listening" state in a new thread. The result of "handshake successful" should be a "Ready/Connected" state (instead of "Listening"), which represents that the payment channel is connected and ready. There are two active threads in the system at this point.

1. When the background "Listening" thread receives a message, it should spawn a new thread, so the listening background thread can continue listening (we therefore have three threads in the system at this point: main, background listening, and new message received thread). Max number of threads and other protections are implementation details.

1. To represent the above third thread, we should have a dotted arrow coming out of the "Listening" state with the action text of "received message" taking it to a state of "Processing Received Message". The "Proposal Accepted" and "Proposal Rejected" states can hang off this new state "Processing Received Message". When this new thread "Processing Received Message" finishes it's "Proposal Accepted" or "Proposal Rejected" actions, it should not return to the "Listening" state since that is in the calling thread, instead it should end the thread, i.e. it can go to the "null" state to represent the end of the (third) thread. The "Listening" thread of course never dies.

1. The `ProposePayment()` action should come from the `Ready/Connected` state. This should send the message and return back to the "Ready/Connected" state, regardless of success or failure (the success or failure state for that specific action does not have it's own state. we can give it it's own state but we are not supporting retries at this point since that can be handled by the consumer calling the functions on the SDK). Note that any asynchronous message received by the other connected SDK will be handled in the "Listening" thread. 

1. We should in fact rename `ProposePayment()` to `SendMessage()` since this could be any message we want to send on the payment channel such as `CloseChannel()`, `CheckChannel()`, `CloseStart()`, etc., all of which return back to the `Ready/Connected` state directly. Note: the channel state is not really performing the networking logic to send the message, so maybe having this reflective function of `SendMessage()` in the state diagram is redundant. I'm in favor of removing it since it is likely going to be on the channel orchestrator and not on the channel itself since it doesn't affect the state of the channel.

Note: we should be clear on the difference between the state of the `Channel` (represented by `ChannelStateMachine`) and the `ChannelOrchestrator`. I propose that the `ChannelStateMachine` holds state such as `C_i`, `D_i`, etc. whereas the `ChannelOrchestrator` holds intermediate information such as the networking logic and networking errors etc. which may or may not result in a change to the state of the `ChannelStateMachine` depending on various factors such as valid/invalid C_i, D_i transactions etc.

acharb:
thx for the notes @nikhilsaraf  🙌  I pushed up the changes to the `payment-proposal-initiator.txt`. Specifically the "Suggestions" section. The "Additional Thoughts" section I think I'm still working through how that could work. While making these adjustments two things popped out to me:

1. If the passive "Listening" thread is handling the incoming payments, we will need a way to store those PaymentProposals. So that the user can access them on the active thread, and then decide to Confirm or Decline. At least, I'm not sure how the passive thread would otherwise be able to pass the proposal to the active thread for the user to address.

2. Splitting up the diagram into an Initiator initiating the payment and the Responder responding to the payment is a good idea imo. Since the different perspectives have different actions available. I created a second file for that perspective in payment-proposal-responder.txt

acharb:
atm I think the value of these diagrams are to be a reference of how networking/messaging could work, as we write the code for the state machine and getting it demo-able.

So I'm thinking of merging this in for now, and then it can be iterated on when we focus more on messaging/networking.

if you agree and the changes look good @nikhilsaraf pls approve :) and then I'll merge

specifications/diagrams/state-diagrams/payment-proposal-initiator.txt
```diff
@@ -0,0 +1,20 @@
...
+    Proposing --> ResponseReceived
+    note right of Proposing: waiting/listening while in proposing state
+    Proposing --> NoResponseReceived: timeout expires
+
+    NoResponseReceived --> Listening
+    ResponseReceived --> Listening: Response invalid, or rejected
+    ResponseReceived --> ProposalRejected: rejected, timeout error, <br> invalid response
+    ProposalRejected --> Listening
+    ResponseReceived --> ProposalAccepted: valid Response <br> receieved signed C_i/D_i
+    ProposalAccepted --> Listening: sign D_i and send both back
```
leighmcculloch, line 20: I think this is a factor of this being a state diagram and a flow chart in one, that this doesn't capture what happens if the initiator doesn't return the final signatures to the responder.

leighmcculloch:
> So I'm thinking of merging this in for now, and then it can be iterated on when we focus more on messaging/networking.

+1 I wouldn't get hung up on making it perfect.

leighmcculloch (approved):
I think there's still things to work through with this, but if you want to merge and iterate 👍🏻.

acharb:
> I think there's still things to work through with this, but if you want to merge and iterate 👍🏻.

agreed these can be re-worked a bit. I'm keeping payment-proposal-responder.txt (renaming to payment-proposal-threads.txt). The other two are (pretty much the same) and probably more confusing than helpful so going to remove those. I think the main gain from this exercise (to me at least) was forcing to think about the threading that Nikhil points out above. Which is shown in the diagram I'm leaving. Going to merge for now then iterate when needed

Issue #39, opened 2021-05-21
sdk: payment channel state machine: setup and open


leighmcculloch:
We need to add a type that tracks the state of a payment channel, and encapsulates knowledge about how the transactions are used. 

The type should take care of which txs need building, as well as tx signing, and should also ensure that txs are safe to sign and valid. 

The type should not have any responsibilities relating to networking or coordinating with the other party.

We should write a test that will act as a demo using the state machine with some basic coordination.

PR #51, opened 2021-05-25
sdk: move state machine into state package and add types

leighmcculloch:
### What
Move the state machine into a package that will own the state machine, called `state`.

Rename `ChannelStateMachine` to `state.Channel`.

Add appropriate types for some fields like accounts (`keypair.FromAddress`), amounts (`int64`).

Rename `IsInitiator` to `Initiator`.

Removed duplicate information with different names.

Store a single balance amount that represents both balances.

Some other small changes.

### Why
These changes generally make the code look and feel more like other Go code we've written and a little more ergonomic. 

The move will help us keep the name of the type shorter and provide scope, rather than try and build so much scope into the name.

The types will help us be clear about what we expect the values to be. The `int64` type is much more appropriate than `int` for the amounts because on the Stellar network amounts are always int64 regardless of the architecture of the CPU the software is running on and in Go int may be int32 or int64.

The `Is` pattern and other prefixes on function names are rarely used in Go and is not particularly idiomatic, see [effective Go](https://golang.org/doc/effective_go#Getters).

It may be dangerous to store the same information in multiple perspectives, such as `InitiatorEscrowAccount` and `MyAccount`. It increases the surface area of the channel state we need to test. Things like `MyAccount` can be calculated in a function if we need it without needing to store state.

We only need to store a single amount for each asset in the perspective of the initiator for whether they owe or are owed money.

Related to #39


acharb (approved):
lgtm 👍

PR #52, opened 2021-05-27
sdk: implement the state machine open process

### LINKED ISSUE #39: sdk: payment channel state machine: setup and open

leighmcculloch:
Close #39

### DIFF
--- a/sdk/state/state_open.go
+++ b/sdk/state/state_open.go
@@ -99,7 +99,7 @@ func (c *Channel) OpenConfirm(m Open) (Open, error) {
 
 	// If local has not signed close, sign it.
 	err = c.verifySigned(close, m.CloseSignatures, c.localSigner)
-	if errors.As(err, &ErrNotSigned{}) {
+	if errors.Is(err, ErrNotSigned{}) {
 		closeSig, err := c.sign(close)
 		if err != nil {
 			return m, fmt.Errorf("open confirm: close incomplete: %w", err)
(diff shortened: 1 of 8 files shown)


sdk/state/state_open.go
```diff
@@ -0,0 +1,151 @@
...
+		return m, err
+	}
+
+	// If remote has not signed close, error as is invalid.
+	err = c.verifySigned(close, m.CloseSignatures, c.remoteSigner)
+	if err != nil {
+		return m, fmt.Errorf("open confirm: close invalid %w", err)
+	}
+
+	// If local has not signed close, sign it.
```
acharb, line 100: I like doing it this way of checking if I've signed before signing, instead of the way I did. I'll change mine after we merge

acharb (approved):
overall lgtm. There are definitely some synergies between this and update that we can make consistent after both are merged

sdk/state/state_open.go
leighmcculloch: I also think we've used some different strategies around how we generate and handle errors we should make the same.

leighmcculloch:
Going to merge. This is definitely incomplete, but we'll iterate on it.

PR #54, opened 2021-05-27
sdk: add helper sign and verify funcs and use local/remote terms

leighmcculloch:
### What
Add helpers for signing and verification, and switches to storing internals of the channel state in the form of local/remote, and adds helper functions for getting at that data from the initiator/responder viewpoint.

### Why
This is a break out of some functionality in #52 that @acharb will probably be able to use while working on #46.

The reason to change the internals to be local/remote focused instead of initiator/responder is because signing keys will be different types for local vs remote, since the local will be a full key and remote will be only an address. So, this keeps all the internals in local/remote perspective.

### DIFF
--- a/sdk/go.mod
+++ b/sdk/go.mod
@@ -2,7 +2,7 @@ module github.com/stellar/experimental-payment-channels/sdk
 
 go 1.16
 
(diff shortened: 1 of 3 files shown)


acharb (approved):
lgtm 💯

PR #55, opened 2021-05-28
payment updates

acharb:
updating the payment channel with new payment + confirming that payment


sdk/state/update.go
```diff
@@ -0,0 +1,140 @@
...
+	ObservationPeriodLedgerGap int64
+	AmountToInitiator          int64
+	AmountToResponder          int64
+	CloseSignatures            xdr.DecoratedSignature
+	DeclarationSignatures      xdr.DecoratedSignature
+}
+
+// TODO - validate inputs? (eg. no negative amounts)
+// initiator will only call this
+func (c *Channel) NewPaymentProposal(me Participant, other Participant, payToInitiator int64, payToResponder int64, startSequence int64, i int64, e int64, o time.Duration, observationPeriodLedgerGap int64, networkPassphrase string) (*PaymentProposal, error) {
```
leighmcculloch, line 23: I think this can be simplified a little, the state machine should store state, so can track s, i, e, and o, as well as participant details.

I was thinking the caller would say how much they want to pay the other party and internally it'd work out what the new balances are. 

Wdyt?

sdk/state/update.go
```diff
@@ -0,0 +1,140 @@
...
+			StartSequence:              p.StartSequence,
+			IterationNumber:            p.IterationNumber,
+			AmountToInitiator:          p.AmountToInitiator,
+			AmountToResponder:          p.AmountToResponder,
+		})
+		if err != nil {
+			return nil, err
+		}
+		// TODO - use correct method
+		if !verifyTxSignatures(txC, p.CloseSignatures) {
```
leighmcculloch, line 91: I added a method in my branch you can use for this.

sdk/state/update_test.go
```diff
@@ -0,0 +1,348 @@
...
+		require.NoError(t, err)
+
+		//// INITIATOR: re-confirms P_i by signing D_i and sending back
+		paymentProposal = PaymentProposal{}
+		err = json.Unmarshal(j, &paymentProposal)
+		paymentProposal, err = initiatorChannel.ConfirmPayment(paymentProposal, networkPassphrase)
+		require.NoError(t, err)
+	}
+
+	// TODO - how does Initiator submit "latest", is it store in the Channel?
```
leighmcculloch, line 253: Good question, I think we'll store in channel, or rather the tx can regenerate the tx and add the signatures.

leighmcculloch (commented):
Left some thoughts inline. I think we should combine our branches.

sdk/state/update.go
acharb: ya  I think you're right. I added them all for this first dirty cut but it looks like everything but pay amount can live on channel, I'll move em 👍

sdk/state/state.go
```diff
@@ -18,18 +18,26 @@ const (
...
 
 type Channel struct {
-	Status ChannelStatus
-
+	Status                 ChannelStatus
 	Initiator              bool
 	InitiatorEscrowAccount *keypair.FromAddress
 	ResponderEscrowAccount *keypair.FromAddress
 
+	// TODO - is this the best way to rep Balance? (perspective of initiator/responder and not Me/Other)
```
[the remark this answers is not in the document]
acharb: I tried using subjective for Balance but it was annoying since payments are objective. I'll take another look, maybe something that just has to be dealt with, or possibly payments could be subjective 🤔

sdk/state/state.go
```diff
@@ -31,8 +31,15 @@ type Channel struct {
...
-	// TODO: balances         []Amount
+	startingSequence        int64
+	iterationNumber         int64
+	iterationNumberExecuted int64
+	// TODO - is this the best way to rep Balance? (perspective of initiator/responder and not Me/Other)
+	// The balance owing from the initiator to the responder, if positive, or
+	// the balance owing from the responder to the initiator, if negative.
+	// TODO - use Balance struct
+	Balance int64
+	Asset   Asset
```
leighmcculloch, line 42: You can use an `Amount` type here, it captures both the asset and amount value in the struct. That'll make it easier for us to add support for multiple balances.

sdk/state/state.go
```diff
@@ -31,8 +31,15 @@ type Channel struct {
 	observationPeriodTime      time.Duration
 	observationPeriodLedgerGap int64
 
-	startingSequence int64
-	// TODO: balances         []Amount
+	startingSequence        int64
+	iterationNumber         int64
+	iterationNumberExecuted int64
```
leighmcculloch, line 36: Can you leave the executed field out for now and hardcode zero for the value anywhere you need it? We won't need the executed value until we support withdrawal, and I'm not sure how we will set the value yet, I think we might need to rethink it.

sdk/state/update.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/stellar/experimental-payment-channels/sdk/txbuild"
+	"github.com/stellar/go/xdr"
+)
+
+type PaymentProposal struct {
```
leighmcculloch, line 11: I think you should use the term `Payment` here, rather than `PaymentProposal`. My reasoning for this is that it starts out as a proposal, but once all its payments are adopted, it is a complete payment and is no longer only a proposal.

sdk/state/update.go
```diff
@@ -0,0 +1,143 @@
...
+	IterationNumber            int64
+	ObservationPeriodTime      time.Duration
+	ObservationPeriodLedgerGap int64
```
leighmcculloch, line 14: I think you are probably on to something here, with capturing these values in the payment and we'll probably need this at some point down the line, but I think we can assume the values will be constants until we look at adding support for changing the observation period during the lifetime of the channel. Would you be okay removing these two for the moment?

leighmcculloch (commented):
I left a couple comments. I still need to look at the payment functions.

sdk/state/update.go
```diff
@@ -0,0 +1,143 @@
...
+	AmountToInitiator          int64
+	AmountToResponder          int64
+	CloseSignatures            []xdr.DecoratedSignature
+	DeclarationSignatures      []xdr.DecoratedSignature
+}
+
+// TODO - validate inputs? (eg. no negative amounts)
+// TODO - payments to be in Amount struct
+// initiator will only call this
+func (c *Channel) NewPaymentProposal(payToInitiator int64, payToResponder int64) (*PaymentProposal, error) {
```
leighmcculloch, line 24: I think you can reduce this interface down to a single amount, the amount the caller wants to send.

```suggestion
func (c *Channel) NewPaymentProposal(amount int64) (*PaymentProposal, error) {
```


Inside the function we can do the following to update an initiator/responder balance, which I see is what's in this PR. The rest of your logic would remain the same inside the function.

```go
if c.initiator {
    c.balance += amount
} else {
    c.balance -= amount
}
```

We could also support a local/remote balance like so, but I don't yet have an opinion on which would have better properties:

```go
c.balance += amount
amountToInitiator := c.initiatorAmountOwed()
amountToResponder := c.responderAmountOwed()
```

We would have functions like `initiatorAmountOwed()` that convert the local representation into the initiator's perspective, and similar for responder.

sdk/state/update.go
```diff
@@ -0,0 +1,143 @@
...
+		AmountToInitiator:          amountToInitiator,
+		AmountToResponder:          amountToResponder,
+		CloseSignatures:            txC.Signatures(),
+	}, nil
+}
+
+// TODO - what do to when you don't have the full other's KP? - Change to FullAddress?
+func (c *Channel) ConfirmPayment(p *PaymentProposal) (*PaymentProposal, error) {
+	// TODO - is c.initiator good way of differentiating?
+	if !c.initiator {
```
leighmcculloch, line 68: The initiator and responder can both make payments, so sometimes the initiator will need to sign the close and decl tx.

I think instead of this if, we should think about what is safe fort he participant to sign.

The participant is safe to sign the close as long as the close is valid.

The participant is safe to sign the declaration, as long as the remote participant has signed the close.

If the above rules are checked for, we can write the signing code to only occur in those cases. That's the methodology I used for the confirm close.

leighmcculloch (commented):
Couple more thoughts.

sdk/state/state.go
acharb: commented it out for now
https://github.com/stellar/experimental-payment-channels/pull/55/commits/cf0af2d2be8645f892e362e195f235c946a2fca5

sdk/state/update.go
acharb: that's a good point, changed to `Payment`
https://github.com/stellar/experimental-payment-channels/pull/55/commits/e2211d4850c38916a7bead62e8d1c2f7370704ab

sdk/state/update.go
acharb: I wasn't 100% if these should be passed from Channel to payments, in the case the Channel changes these values before close, payments before and after would store the different values which could be useful. Removing for now for simplicity sgtm though
https://github.com/stellar/experimental-payment-channels/pull/55/commits/69cb66636c808d80826fde64e5ea4f68999bcbfb

sdk/state/update.go
```diff
@@ -0,0 +1,127 @@
...
+			StartSequence:              c.startingSequence,
+			IterationNumber:            c.iterationNumber,
+			AmountToInitiator:          amountToInitiator,
+			AmountToResponder:          amountToResponder,
+		})
+		if err != nil {
+			return nil, err
+		}
+		if err := c.verifySigned(txC, p.CloseSignatures, c.remoteSigner); err != nil {
+			return nil, errors.Wrap(err, "the signed transaction may have different data")
```
leighmcculloch, line 84: Instead of using errors.Wrap, can you use `fmt.Errorf` with `%w` that will wrap?

```suggestion
			return nil, fmt.Errorf("the signed transaction may have different data: %w", err)
```

The pkg/errors package has some big tradeoffs and idiosyncrasies, and I saw the author stated a while ago that they don't use it themselves. If you want to go down a rabbit hole reading about problems the package causes, see these: https://github.com/pkg/errors/issues/236, https://github.com/pkg/errors/issues/228, https://github.com/pkg/errors/issues/174, https://github.com/pkg/errors/issues/90.

I know we use pkg/errors in the stellar/go repo, but we should do everything we can to reduce where we introduce it now that the Go stdlib has the same capabilities built in since Go 1.13 without the problems linked above.

leighmcculloch (commented):
One more:

sdk/state/update.go
acharb: ya  that is definitely a better way of thinking about it. Made it much more clear imo
https://github.com/stellar/experimental-payment-channels/pull/55/commits/c84ca0309abd42cd75d04d1f491b82f8c89f1bac

sdk/state/update.go
acharb: reduced to a single amount input: https://github.com/stellar/experimental-payment-channels/pull/55/commits/c84ca0309abd42cd75d04d1f491b82f8c89f1bac

sdk/state/update.go
acharb: TIL!
https://github.com/stellar/experimental-payment-channels/pull/55/commits/df02ced4c1911feeacc251acfb046fc1e2d03d19

sdk/state/state.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/55/commits/9add63d410c809c7cda00cedecd510ea5a83b87f

proposal to name the struct `Balance` instead of `Amount`? `channel.amount.Amount` feels weird, and a "balance" per asset makes sense

sdk/txbuild/close.go
```diff
@@ -27,8 +27,9 @@ func Close(p CloseParams) (*txnbuild.Transaction, error) {
 			AccountID: p.InitiatorEscrow.Address(),
 			Sequence:  startSequenceOfIteration(p.StartSequence, p.IterationNumber) + 1, // Close is the second transaction in an iteration's transaction set.
 		},
-		BaseFee:              0,
-		Timebounds:           txnbuild.NewTimeout(300),
+		BaseFee: 0,
+		// TODO - Timebounds needs to be explicit
+		Timebounds:           txnbuild.NewInfiniteTimeout(),
```
leighmcculloch, line 32: Good catch. These should be infinite since a channel is designed to be open indefinitely.
```suggestion
		BaseFee:              0,
		Timebounds:           txnbuild.NewInfiniteTimeout(),
```

sdk/state/update.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, err
+	}
+	return &Payment{
+		Amount:          amount,
+		CloseSignatures: txC.Signatures(),
+		FromInitiator:   c.initiator,
+	}, nil
+}
+
+func (c *Channel) ConfirmPayment(p *Payment) (*Payment, error) {
```
leighmcculloch, line 55: We'll need to normalize the names of the functions. In my PR I used OpenPropose and OpenConfirm, so they're grouped by Open, but they read a little odd. NewOpen would also be odd though, there can only ever be one open and close.

What do you think of the following? The names Begin and Commit are from database terms, although it works a bit differently since you have to commit multiple times.

```
OpenBegin
OpenCommit

PaymentBegin
PaymentCommit

CloseBegin
CloseCommit
```

leighmcculloch (approved):
I think we should merge this along with #52 and then address inconsistencies because things probably need changing in both. I can take addressing those things in a new PR and you can review.

sdk/state/state.go
```diff
@@ -15,15 +15,15 @@ type (
 	CreditAsset = txnbuild.CreditAsset
 )
 
-type Amount struct {
+type Balance struct {
```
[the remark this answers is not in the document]
acharb: ya  I changed it back for now, we can decide to change or not on a later PR

sdk/state/update.go
```diff
@@ -0,0 +1,124 @@
...
+	var newCloseSignatures, newDeclarationSignatures []xdr.DecoratedSignature
+	var amountFromInitiator, amountFromResponder int64
+	if p.FromInitiator {
+		amountFromInitiator = p.Amount
+	} else {
+		amountFromResponder = p.Amount
+	}
+	newBalance := c.balance.Amount + amountFromInitiator - amountFromResponder
+
```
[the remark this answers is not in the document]
acharb: right now it's living on the Payment, so iterated by the person sending the payment. I created a ticket to maybe do differently here
https://github.com/stellar/experimental-payment-channels/issues/61

Issue #63, opened 2021-06-02
sdk/state: store last confirmed and not-yet-confirmed payment on channel


acharb:
also store new balance on payment.

evaluate channels current balance and iteration number always off last confirmed payment.

edit: just evaluate current balance as part of this ticket. Iteration number will be done in: https://github.com/stellar/experimental-payment-channels/issues/64

acharb:
completed by https://github.com/stellar/experimental-payment-channels/pull/68

Issue #65, opened 2021-06-02
sdk/state: distinguish between the not signed and incomplete, or wrong signature / wrong data.


acharb:
Probably need to error on extra/unverifiable signatures.

leighmcculloch:
Let's add a third parameter to the confirm functions, `fullySigned`, that will not be false when the item is not fully signed by all parties. It's a little gross, and we can look for better ways, but this is very verbose and accurate representation about the outputs I think.

We should also still add an error for extra/unverifiable signatures as you mention.

leighmcculloch:
Closed by #76

PR #68, opened 2021-06-03
sdk: store lastPayment on channel  and evaluate balance from there 

acharb:
store the lastConfirmedPayment on the channel, and evaluate balances based off of that payment

also, removes the json marshalling/unmarshalling from the test to simulate the handoff. More trouble than it's worth

fulfills part of https://github.com/stellar/experimental-payment-channels/issues/63, but not the iteration part. Going to do that in separate PR

### DIFF
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -81,9 +81,9 @@ func (c *Channel) SetIterationNumber(i int64) {
 	c.iterationNumber = i
 }
 
-// Amount returns the amount owing from the initiator to the responder, if positive, or
+// Balance returns the amount owing from the initiator to the responder, if positive, or
 // the amount owing from the responder to the initiator, if negative.
-func (c *Channel) Amount() Amount {
+func (c *Channel) Balance() Amount {
 	if c.lastConfirmedPayment == nil {
 		return Amount{NativeAsset{}, 0}
 	}
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -43,7 +43,7 @@ type Channel struct {
 	localSigner  *keypair.Full
 	remoteSigner *keypair.FromAddress
 
-	lastConfirmedPayment *Payment
+	latestCloseAgreement *CloseAgreement
 	// TODO - need to store this too?
 	// lastUnConfirmedPayment *Payment
 }
@@ -84,10 +84,25 @@ func (c *Channel) SetIterationNumber(i int64) {
 // Balance returns the amount owing from the initiator to the responder, if positive, or
 // the amount owing from the responder to the initiator, if negative.
 func (c *Channel) Balance() Amount {
-	if c.lastConfirmedPayment == nil {
+	if c.latestCloseAgreement == nil {
 		return Amount{NativeAsset{}, 0}
 	}
-	return c.lastConfirmedPayment.NewBalance
+	return c.latestCloseAgreement.Balance
+}
+
+// newBalance is a hlper method for computing what the new channel balance will be if
+// the input payment is submitted successfully.
+func (c *Channel) newBalance(p *Payment) Amount {
+	var amountFromInitiator, amountFromResponder int64
+	if p.FromInitiator {
+		amountFromInitiator = p.Amount.Amount
+	} else {
+		amountFromResponder = p.Amount.Amount
+	}
+	return Amount{
+		Asset:  p.Amount.Asset,
+		Amount: c.Balance().Amount + amountFromInitiator - amountFromResponder,
+	}
 }
 
 func (c *Channel) initiatorEscrowAccount() *EscrowAccount {
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -36,17 +36,16 @@ type Channel struct {
 	// TODO - leave execution out for now
 	// iterationNumberExecuted int64
 
-	// The balance owing from the initiator to the responder, if positive, or
-	// the balance owing from the responder to the initiator, if negative.
-	// TODO - use Balance struct
-	amount Amount
-
 	initiator           bool
 	localEscrowAccount  *EscrowAccount
 	remoteEscrowAccount *EscrowAccount
 
 	localSigner  *keypair.Full
 	remoteSigner *keypair.FromAddress
+
+	lastConfirmedPayment *Payment
+	// TODO - need to store this too?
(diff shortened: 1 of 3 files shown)


sdk/state/state.go
```diff
@@ -36,17 +36,16 @@ type Channel struct {
...
 	initiator           bool
 	localEscrowAccount  *EscrowAccount
 	remoteEscrowAccount *EscrowAccount
 
 	localSigner  *keypair.Full
 	remoteSigner *keypair.FromAddress
+
+	lastConfirmedPayment *Payment
+	// TODO - need to store this too?
+	// lastUnConfirmedPayment *Payment
```
acharb, line 48: @leighmcculloch why again do we need to store lastUnconfirmed? Is this something that will be useful later?

leighmcculloch: Once a participant has signed C or D for an iteration number they really shouldn't sign a new payment with the same iteration number. So we need to store something, storing the incomplete payment seems possibly useful and really easy to understand as a reader of the code. We might be able to optimize it and store less but we don't need to optimize that yet.

sdk/state/state.go
```diff
@@ -82,9 +81,13 @@ func (c *Channel) SetIterationNumber(i int64) {
 	c.iterationNumber = i
 }
 
-// TODO: Remove
+// Amount returns the amount owing from the initiator to the responder, if positive, or
+// the amount owing from the responder to the initiator, if negative.
 func (c *Channel) Amount() Amount {
```
leighmcculloch, line 86: I don't remember why we named this Amount, but we should probably name it Balance.
```suggestion
func (c *Channel) Balance() Amount {
```

sdk/state/state_test.go
```diff
@@ -255,29 +254,18 @@ func Test(t *testing.T) {
...
 		require.NoError(t, err)
 
 		//// Receiver: receives new payment proposal, validates, then confirms by signing both
 		receivingChannel.SetIterationNumber(i)
-		payment = &state.Payment{}
-		err = json.Unmarshal(j, payment)
-		require.NoError(t, err)
-
 		payment, err = receivingChannel.ConfirmPayment(payment)
 		require.NoError(t, err)
```
leighmcculloch, line 266: This is not really specific to this PR, but this no error line seems suspicious. I think it should return an error since the payment isn't confirmed, because the sending channel hasn't signed the declaration yet.

sdk/state/update.go
```diff
@@ -11,21 +11,22 @@ import (
 
 type Payment struct {
 	IterationNumber       int64
-	Amount                int64
+	Amount                Amount
 	CloseSignatures       []xdr.DecoratedSignature
 	DeclarationSignatures []xdr.DecoratedSignature
 	FromInitiator         bool
+	NewBalance            Amount
```
leighmcculloch, line 18: I think I referred to this concept as 'new balance', but on seeing it in code this looks a little too much like a NewBalance function. I think it'd be fine to name this Balance.
```suggestion
	Balance               Amount
```
Wdyt?

sdk/state/update.go
```diff
@@ -126,7 +121,7 @@ func (c *Channel) ConfirmPayment(p *Payment) (*Payment, error) {
 
 	p.CloseSignatures = append(p.CloseSignatures, txClose.Signatures()...)
 	p.DeclarationSignatures = append(p.DeclarationSignatures, txDecl.Signatures()...)
-	c.amount.Amount = newBalance
+	c.lastConfirmedPayment = p
```
leighmcculloch, line 124: ❗ This records the payment as confirmed, but it is not confirmed until all the signatures are on the payment. I think we need a check before these few lines to check that if the remote has not signed the declaration, it should return an error. Similar to this code in the open func:

https://github.com/stellar/experimental-payment-channels/blob/2d90876b9f238302e57d0b3a2927e1a7e9c4d4e3/sdk/state/state_open.go#L142-L146

leighmcculloch (commented):
Couple things, only one critical (❗) that I think needs addressing.

sdk/state/state.go
leighmcculloch: I've opened an issue for this specifically: https://github.com/stellar/experimental-payment-channels/issues/69.

sdk/state/state.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/68/commits/acb91cd3780757ea147ce4165662c621e2db3a89

sdk/state/update.go
acharb: imo `payment.Balance` and `payment.Amount` sound very similar. What about `payment.ChannelBalance`?

leighmcculloch: Maybe we shouldn't be storing payment, but instead storing an `Agreement` or `CloseState` or `CloseAgreement`.

A payment needs an iteration number, source, amount, and signatures.

A close agreement needs an iteration number, balance, and signatures.

We've got some overlap but these feel like different things.

The main reason we're storing a payment is to be able to encode channel state like balance and be clear about which payment/agreement/iteration they belong. We still get that if we store something like these as a different type with the specific fields we need.

sdk/state/state.go
acharb: ok thx, I'll leave commented for this PR then

sdk/state/update.go
acharb: I agree CloseAgreement works better. Changed it here:
https://github.com/stellar/experimental-payment-channels/pull/68/commits/fac8cc922ab48591a87cfd903a80d5018c6c18ac

sdk/state/update.go
acharb: That's a good point. To avoid merge conflicts I can leave this change as part of the returning `payment, fullySigned, err :=` ticket here: https://github.com/stellar/experimental-payment-channels/issues/65

lmk if that sounds good, otherwise I can change it here first

sdk/state/state_test.go
acharb: true. Going to leave this change as part of this ticket if that sounds good to you
https://github.com/stellar/experimental-payment-channels/issues/65

sdk/state/update.go
leighmcculloch: That's good with me.

leighmcculloch (approved):

PR #76, opened 2021-06-03
sdk/state: add fully signed param to confirm functions

leighmcculloch:
### What
Add fully signed param to confirm functions that indicates if the value being returned is completely signed and valid for submission if the user wished to submit the transactions to the network. Also, the error value of the confirm functions no longer error when the value is not completely signed, and no other error occurred.

### Why
@acharb mentioned it was confusing that the functions returns values that would be usable when returning an errors. I also found it difficult to remember that the errors returned from these functions could be catastrophic, but then also just routine part of how the process works.

It's a bit more verbose but clearer if the function returns some solid signal that this is fully signed, or complete and submittable, separate to whether some other arbitrary error occurred.

This seemed like a good use case for error inspection, but in reality it was more confusing.

### DIFF
--- a/sdk/state/update.go
+++ b/sdk/state/update.go
@@ -131,7 +131,7 @@ func (c *Channel) ConfirmPayment(p *Payment) (payment *Payment, fullySigned bool
 		p.DeclarationSignatures = append(p.DeclarationSignatures, txDecl.Signatures()...)
 	}
 
-	// If remote has not signed close, error as is incomplete.
+	// If remote has not signed declaration, it is incomplete.
 	signed, err = c.verifySigned(txDecl, p.DeclarationSignatures, c.remoteSigner)
 	if err != nil {
 		return p, fullySigned, fmt.Errorf("verifying declaration signed by remote: %w", err)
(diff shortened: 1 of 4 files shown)


sdk/state/update.go
```diff
@@ -90,36 +90,63 @@ func (c *Channel) PaymentTxs(p *Payment) (close, decl *txnbuild.Transaction, err
...
+	if !signed {
 		txDecl, err = txDecl.Sign(c.networkPassphrase, c.localSigner)
 		if err != nil {
-			return nil, err
+			return p, fullySigned, err
 		}
+		p.DeclarationSignatures = append(p.DeclarationSignatures, txDecl.Signatures()...)
+	}
+
+	// If remote has not signed close, error as is incomplete.
```
acharb, line 134:
```suggestion
	// If remote has not signed declaration, error as is incomplete.
```

acharb: actually maybe this should say:
`// If remote has not signed declaration, return with not fully signed.`

acharb (approved):
lgtm 💯

sdk/state/update.go
leighmcculloch:
```suggestion
	// If remote has not signed declaration, it is incomplete.
```

leighmcculloch:
Close #65

PR #78, opened 2021-06-04
sep: change disbursement to be safer with multiple and regulated assets

leighmcculloch:
### What
Move the disbursement payment operations out of the closing transaction and into separate transactions, one per payment, that can be executed between the declaration and closing transaction.

### Why
In #34 we've been discussing the impact a regulated asset issuer could have on the state of a payment channel. A regulated asset issuer using clawback and auth revocation could inadvertently cause irreparable damage to a payment channel if the payment channel contains other assets not under their control. There would be no guarantee a participant could recover other assets from a channel that experiences clawback or revocation.

The participants trust the asset issuers to some degree, that's what trustlines imply, and if an asset issuer made a mistake and accidentally clawed back they could easily coordinate the return of funds to the appropriate escrow accounts. But if the channel becomes broken and one participant is gone, there is nothing the issuer can do to help the participant recover other assets.

This change separates the disbursements into separate transactions that must be executed between the declaration and close. The first payment after the declaration is prevented from executing until the observation period. Subsequent payments and the closing transaction must be executed in order following. If any payment fails its sequence number is consumed and subsequent payments and the closing transaction may proceed. The channel can close, even if one payment fails.

cc @accordeiro @nikhilsaraf @stanford-scs @acharb @MonsieurNicolas


leighmcculloch:
We've also been discussing how you could use SEP-8 with a payment channel. I should note that this PR doesn't add anything new to make that possible, I don't think we'll add those details to this SEP, but probably a separate SEP. This change solves the safety problems with using a regulated asset with the channel.

specifications/sep-payment-channel-transactions.md
```diff
@@ -229,45 +226,58 @@ payments to one another within the channel.
...
-period time duration), and a `minSeqLedgerGap` of O (the observation period
-ledger count).
+- C_i, the _closing transaction_, changes the signing weights on EI such that I
+unilaterally controls EI, and the signing weights on ER such that R unilaterally
+controls ER.  C_i has source account EI, sequence number s_i+1+n.
 
   The `minSeqAge` and `minSeqLedgerGap` prevents a misbehaving party from
   executing C_i when the channel state has already progressed to a later
-  iteration number, as the other party has the period O to invalidate C_i by
+  iteration number, as the other party has the period 2O to invalidate C_i by
```
acharb, line 255: 2O is hard to distinguish from 20 (twenty) when reading:
```suggestion
  iteration number, as the other party has the period 2*O to invalidate C_i by
```

specifications/sep-payment-channel-transactions.md
```diff
@@ -229,45 +226,58 @@ payments to one another within the channel.
...
-  to ER, or from ER to EI.
   - One or more `SET_OPTIONS` operation adjusting escrow account EI's thresholds
   to give I full control of EI, and removing R's signers.
   - One or more `SET_OPTIONS` operation adjusting reserve account ER's
   thresholds to give R full control of ER, and removing I's signers.
 
+- P_{i,a}, the _payment transaction_, disburses a single asset a from EI to ER
+or from ER to EI. P_{i,a} has source account EI, sequence number s_i+1+a, where
+a is the zero index of the asset where each asset in the channel is assigned an
+index. P_{i,0}, the first payment, is assigned a `minSeqAge` of O (the
```
acharb, line 267: "O" defining both `observation period time duration` and `the observation period ledger count` is confusing to me since it's implying that "O" would be the same value for both (which I don't think is the case). It looks like we're using "O" generally in this SEP to describe "observation period", which is describing two fields.

💭  proposal to change "O" here to `O_minSeqAge` and `O_minSeqLedgerGap`, or some other way of differentiating between its 2 uses

acharb (commented):
lgtm, had a couple of nits. I also had a more general question below:

❓ I think I haven't fully understood if one of the P_i fails, which is why we're separating it from C_i, how is C_i still able to be submitted successfully, since its seq number is off?

I think if we added a `Payment Transaction Failure` section under `Security Concerns` that would be a good way to address this.

specifications/sep-payment-channel-transactions.md
leighmcculloch: Oops, 2O should be gone, it should be O. 2O was required in an earlier commit on this PR, then I removed it. Fix incoming.

specifications/sep-payment-channel-transactions.md
leighmcculloch: Agree it is a little confusing, although I'm not sure there's a good formal notation like what you propose. The correct way to represent the two different values would be to define the separately, such as to define them as OA and OG. I'm not sure it is worth introducing two separate variables though. It's out of scope of this PR. Could you open an issue?

acharb: https://github.com/stellar/experimental-payment-channels/issues/79

specifications/sep-payment-channel-transactions.md
leighmcculloch: I've fixed this in 9b78df3.

Although I realize this creates a new issue, if there are no payments, the close is not delayed. I'll fix that shortly.

leighmcculloch: I fixed the second issue with the observation period missing on close with no payments in 9fa8ec1.

leighmcculloch:
> I think I haven't fully understood if one of the P_i fails, which is why we're separating it from C_i, how is C_i still able to be submitted successfully, since its seq number is off?

@acharb A failed payment, due to the asset being auth revoked, or clawed back, will result in the sequence number being consumed. The transaction is still valid so it will be included in the ledger, but it will fail when being processed. Failed transactions still consume the sequence number.

specifications/sep-payment-channel-transactions.md
acharb: cool, thanks for changing

acharb (approved):
lgtm 💯 thx for addressing the comments

accordeiro (approved):
I think this is a great improvement! Left a small inline suggestion.

specifications/sep-payment-channel-transactions.md
```diff
@@ -229,45 +227,61 @@ payments to one another within the channel.
...
+3. Sign and exchange zero or more payment transactions P_{i,a}.
+4. Sign and exchange a declaration transaction D_i.
 
-It is important that D_i is signed after C_i because D_i will invalidate any
-previously signed C_i. If I and R signed and exchanged D_i first either party
-could prevent the channel from closing without coordination by submitting D_i
-and refusing to sign C_i.  The participants would not be able to close the
-channel, or regain control of the accounts, and the assets within without
-coordinating with each other.
+It is important that D_i is signed after C_i and P_i because D_i will invalidate
```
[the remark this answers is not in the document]
leighmcculloch: I think this is a good change, but I think we should add some specificity to what's being exchanged. They shouldn't exchange signatures, but they may still exchange transactions.

```suggestion
It is critical that signatures for D_i are exchanged after C_i and P_i because D_i will invalidate
```

Wdyt?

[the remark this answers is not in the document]
leighmcculloch: I took this change to a new PR since it is independent to the change this PR is introducing. Could you review https://github.com/stellar/experimental-payment-channels/pull/123?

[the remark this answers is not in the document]
accordeiro: Sure, thanks!

leighmcculloch:
@acharb identified that this change has an added benefit for all payment channels, not channels holding regulated assets.

This change prevents a channel from becoming stuck if one participant, with nothing to lose, breaks the closing transactions by increasing their own holdings of an asset up to the asset limit before the closing transaction, containing a payment to that participant with that asset, executes. 🎉 

Related discussion here https://github.com/stellar/experimental-payment-channels/issues/93#issuecomment-870955992.

leighmcculloch:
Closing this after today's discussion because we've resolved to focus on single asset support which doesn't need this.

Also @stanford-scs shared an idea with me about how we can use recovery transactions to get similar behavior in a way that is extendable, so I think when we return to multiasset, if we do, we need to weigh that idea as well.

Issue #69, opened 2021-06-03
sdk/state: prevent reusing iteration number when payment not completely confirmed


leighmcculloch:
The channel stores the last confirmed payment but nothing about a payment that is partially confirmed. This means it's possible for the state machine to sign multiple different incomplete payments that contain different balances within. This is a problem because once a participant has signed for one state at a specific iteration number, if they sign a new state at the same iteration number, there will be no way for them to contest the old state.

The channel should probably store the last incomplete confirmed payment and for simplicity not allow signing a new payment if there is another payment that hasn't been fully confirmed. The only course of action if a payment never fully confirms should be to close the channel.

PR #81, opened 2021-06-05
sdk/state: store latestUnconfirmedPayment, add unit test

acharb:
this stores the latestUnconfirmedPayment on the channel, and uses it for validation in `ConfirmPayment`

This is to prevent the case of a receiver receiving two payments with different data but using the same next iteration number.

Also adds a unit test to test this

closes https://github.com/stellar/experimental-payment-channels/issues/69

### DIFF
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -41,8 +41,8 @@ type Channel struct {
 	localSigner  *keypair.Full
 	remoteSigner *keypair.FromAddress
 
-	latestCloseAgreement     *CloseAgreement
-	latestUnconfirmedPayment *Payment
+	latestCloseAgreement     CloseAgreement
+	latestUnconfirmedPayment Payment
 }
 
 type Config struct {
@@ -74,26 +74,21 @@ func NewChannel(c Config) *Channel {
 }
 
 func (c *Channel) NextIterationNumber() int64 {
-	if c.latestUnconfirmedPayment != nil {
+	if !c.latestUnconfirmedPayment.isEmpty() {
 		return c.latestUnconfirmedPayment.IterationNumber
-	} else if c.latestCloseAgreement != nil {
-		return c.latestCloseAgreement.IterationNumber + 1
 	}
-	return 1
+	return c.latestCloseAgreement.IterationNumber + 1
 }
 
 // Balance returns the amount owing from the initiator to the responder, if positive, or
 // the amount owing from the responder to the initiator, if negative.
 func (c *Channel) Balance() Amount {
-	if c.latestCloseAgreement == nil {
-		return Amount{NativeAsset{}, 0}
-	}
 	return c.latestCloseAgreement.Balance
 }
 
 // newBalance is a hlper method for computing what the new channel balance will be if
 // the input payment is submitted successfully.
-func (c *Channel) newBalance(p *Payment) Amount {
+func (c *Channel) newBalance(p Payment) Amount {
 	var amountFromInitiator, amountFromResponder int64
 	if p.FromInitiator {
 		amountFromInitiator = p.Amount.Amount
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -41,8 +41,7 @@ type Channel struct {
 	localSigner  *keypair.Full
 	remoteSigner *keypair.FromAddress
 
-	latestCloseAgreement *CloseAgreement
-	// TODO - set this, probably use different name
+	latestCloseAgreement     *CloseAgreement
 	latestUnconfirmedPayment *Payment
 }
 
@@ -75,15 +74,12 @@ func NewChannel(c Config) *Channel {
 }
 
 func (c *Channel) NextIterationNumber() int64 {
-	var latestI int64
 	if c.latestUnconfirmedPayment != nil {
-		latestI = c.latestUnconfirmedPayment.IterationNumber
+		return c.latestUnconfirmedPayment.IterationNumber
 	} else if c.latestCloseAgreement != nil {
-		latestI = c.latestCloseAgreement.IterationNumber
-	} else {
-		latestI = 0
+		return c.latestCloseAgreement.IterationNumber + 1
 	}
-	return latestI + 1
+	return 1
 }
 
 // Balance returns the amount owing from the initiator to the responder, if positive, or
--- a/sdk/state/update.go
+++ b/sdk/state/update.go
@@ -10,6 +10,13 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+// The high level steps for creating a channel update should be as follows, where the returned payments
+// flow to the next step:
+// 1. Sender calls ProposePayment()
+// 2. Receiver calls ConfirmPayment
+// 3. Sender calls ConfirmPayment
+// 4. Receiver calls ConfirmPayment
+
 type Payment struct {
 	IterationNumber       int64
 	Amount                Amount
@@ -103,6 +110,9 @@ func (c *Channel) PaymentTxs(p Payment) (close, decl *txnbuild.Transaction, err 
 	return
 }
 
+// ConfirmPayment confirms a payment. The original proposer should only have to call this once, and the
+// receiver should call twice. First to sign the payments and store signatures, second to just store the new signatures
+// from the other party's confirmation.
 func (c *Channel) ConfirmPayment(p Payment) (payment Payment, fullySigned bool, err error) {
 	// at the end of this method if a fully signed payment, create a close agreement and clear latest latestUnconfirmedPayment to
 	// prepare for the next update. If not fully signed, save latestUnconfirmedPayment, as we are still in the process of confirming.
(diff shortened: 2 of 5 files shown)


sdk/state/state.go
```diff
@@ -75,15 +74,12 @@ func NewChannel(c Config) *Channel {
 }
 
 func (c *Channel) NextIterationNumber() int64 {
```
leighmcculloch, line 76: This function is a bit subtle to use correctly because in the case we're abandoning a precious unconfirmed payment it returns the wrong value. Can we make this function not exported? Or is it being used in the test now?

leighmcculloch (commented):
I think there's a couple things to address about how we're using and testing the `Payment`. Comments inline.

sdk/state/update.go
```diff
@@ -97,6 +99,10 @@ func (c *Channel) ConfirmPayment(p *Payment) (payment *Payment, fullySigned bool
 		return nil, fullySigned, errors.New(fmt.Sprintf("invalid payment iteration number, got: %s want: %s",
 			strconv.FormatInt(p.IterationNumber, 10), strconv.FormatInt(c.NextIterationNumber(), 10)))
 	}
+	if c.latestUnconfirmedPayment != nil && !reflect.DeepEqual(c.latestUnconfirmedPayment, p) {
```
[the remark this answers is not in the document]
acharb: ya  good catch, they should have different signatures. Test was passing because storing the pointer and then passing it around. I changed it to pass around payment as a value, and to store as a value. Also changed the `CloseAgreement` to store as value to prevent a similar test passing there when we add it

https://github.com/stellar/experimental-payment-channels/pull/81/commits/68688f7161180713daa554e55f57e4db37e89dc1

sdk/state/update.go
```diff
@@ -143,6 +149,7 @@ func (c *Channel) ConfirmPayment(p *Payment) (payment *Payment, fullySigned bool
 		return p, fullySigned, fmt.Errorf("verifying declaration signed by remote: %w", err)
 	}
 	if !signed {
+		c.latestUnconfirmedPayment = p
```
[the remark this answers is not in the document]
acharb: > I think we're missing storing the payment in other cases above. We need to store it once we have signed any transaction for it, because we need to make sure that once we've signed anything for it, we don't ever sign those same types of transactions with different details.

my thinking was that we only want to store the unconfirmedPayment when we return from the method w/o an error. Eg. if for whatever reason signing a transaction returns an error, I didn't think we'd want to store the "bad" input payment that triggered that. Then if a "correct" payment is sent to `ConfirmPayment` because we've stored the "bad" one and the method will fail since the payments don't match.

I added the `defer func`, but to return early if an error was returned
https://github.com/stellar/experimental-payment-channels/pull/81/commits/c887334385256a4a4c1d845e83fb166e455c7c3c

sdk/state/state.go
acharb: > This function is a bit subtle to use correctly because in the case we're abandoning a precious unconfirmed payment it returns the wrong value

in which case would that happen? When `latestUnconfirmedPayment` gets reset the function defers to `latestCloseAgreement` for the correct next i, which should always be valid 🤔 .

But ya  it's being used in `state_test.go` currently to make sure the channel iteration numbers are aligned/correct

sdk/state/update.go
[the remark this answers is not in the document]
acharb: > Another thought is by overwriting the entire payment each time there is a chance we lose signatures. It's possible we sign a payment, record the signature, then the payment passed in doesn't have that signature, and then we lose the signature internally. It's important that once we sign a transaction its signature is stored. Maybe we shouldn't set it, but instead merge it. It might be useful if we add a Clone function to Payment too so that we can do deep clones at these points that are well defined. Wdyt?

that's a good point. After thinking about it probably there will be refactoring so I'd like to merge this in and then address that in a separate ticket: https://github.com/stellar/experimental-payment-channels/issues/82

sdk/state/state.go
acharb: I'm thinking of leaving exported for now, but we can circle back to this if you'd like when you come back online

sdk/state/update.go
```diff
@@ -18,16 +18,24 @@ type Payment struct {
 	FromInitiator         bool
 }
 
+func (p Payment) isEqual(p2 Payment) bool {
```
accordeiro, line 21: I find the `isEqual` function name a bit misleading here, since you're not fully comparing the objects (and you're actually expecting the signatures to be different). Could we either:

- Use a different name, such as `isEquivalent`? or
- Document why we're not checking for all fields in a comment?

sdk/state/update.go
```diff
@@ -92,11 +101,28 @@ func (c *Channel) PaymentTxs(p *Payment) (close, decl *txnbuild.Transaction, err
 	return
 }
 
-func (c *Channel) ConfirmPayment(p *Payment) (payment *Payment, fullySigned bool, err error) {
+func (c *Channel) ConfirmPayment(p Payment) (payment Payment, fullySigned bool, err error) {
+	defer func() {
```
accordeiro, line 105: I like the simplicity of adding a deferred function call here for state cleanup, but I do think this affects readability. I know you mentioned refactoring this later on, but could you leave a few comment lines explaining why this is here?

sdk/state/update_test.go
```diff
@@ -0,0 +1,75 @@
...
+		IterationNumber: 1,
+		Amount: Amount{
+			Asset:  NativeAsset{},
+			Amount: 400,
+		},
+		CloseSignatures: p.CloseSignatures,
+	}
+	_, fullySigned, err = receiverChannel.ConfirmPayment(pDifferent)
+	assert.False(t, fullySigned)
+	require.Error(t, err)
```
accordeiro, line 62: Should we also assert that `receiverChannel.latestUnconfirmedPayment` remains unchanged?

```suggestion
	require.Error(t, err)
	assert.Equal(t, p, receiverChannel.latestUnconfirmedPayment)
```

sdk/state/update_test.go
```diff
@@ -0,0 +1,75 @@
...
+	require.Equal(t, "a different unconfirmed payment exists", err.Error())
```
accordeiro, line 63: Should we also validate that the latestCloseAgreement remains empty?

sdk/state/update_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Confirming a payment with same sequence number and same amount should pass
+	p, fullySigned, err = sendingChannel.ConfirmPayment(p)
+	assert.True(t, fullySigned)
+	require.NoError(t, err)
+	assert.Equal(t, Payment{}, sendingChannel.latestUnconfirmedPayment)
+
+	p, fullySigned, err = receiverChannel.ConfirmPayment(p)
```
accordeiro, line 71: What are you testing here? That confirming payments is an idempotent operation? Just wonder if it's intentional that we're not triggering an error here because the tx has already been confirmed.

accordeiro (approved):
This looks great! Left a few comments and questions inline.

sdk/state/update.go
acharb: good point. `isEquivalent` makes sense to me. I also added a comment that hopefully makes it clearer

sdk/state/update.go
acharb: added a comment here that I think make the intent clearer. Lmk if it is still not clear
https://github.com/stellar/experimental-payment-channels/pull/81/commits/69c1b651a70d8a91162de5c6f943b26ceb82b6ab

sdk/state/update_test.go
acharb: 👍  https://github.com/stellar/experimental-payment-channels/pull/81/commits/0affc1f5abf6532953244e61a06078f91700e2f2

sdk/state/update_test.go
acharb: 👍  https://github.com/stellar/experimental-payment-channels/pull/81/commits/23f87e013c43bc3fb906b7940915bb3ab9ceca85

also looks like [testify uses DeepEqual](https://github.com/stretchr/testify/blob/acba37e5db06f0093b465a7d47822bf13644b66c/assert/assertions.go#L58) for comparing structs, so asserting equal empty structs should be fine here

sdk/state/update_test.go
acharb: ya  this isn't obvious but the receiver actually has to call `ConfirmPayment` twice, so this is testing that. The first time called it adds its signatures. The second time called it essentially is just grabbing the signatures the sender gave. So essentially:

1. Sender calls ProposePayment
2. Receiver calls ConfirmPayment
3. Sender calls ConfirmPayment
4. Receiver calls ConfirmPayment

This was a design decision to use the same method for steps 2,3,4 since they're similar

acharb: I added some comments that explains that:
https://github.com/stellar/experimental-payment-channels/pull/81/commits/df2993648747a22ee09d8f2ba9f356e3c431b0b4

Issue #70, opened 2021-06-03
sdk/state: write tests for open, payment, close


leighmcculloch:
We need to write tests for:
- [ ] Open
- [ ] Payment
- [ ] Close

This is probably going to be a pretty large amount of work, but it's pretty important we come up with a methodical approach to adding test coverage.

Most of the state machine we weren't sure how it should be designed, so we started with a single integration test at a high-level that dictated the shape of the API and both the happy path and the path where a participant closes at the wrong point. This has served us well to ensure the major parts are working while we figure out the API of the state machine, but it might be too verbose for covering all the edge cases we still need to code for.

Writing these tests should highlight many edge cases we aren't covering, and will require more than just writing code. As we identify edge cases we probably need to break this up into more issues.

We should probably start writing tests that are in the form:
1. Create channel in memory
2. Call function to modify state
3. Check outputs and new state is valid

leighmcculloch:
We've been filling a lot of gaps as we go during other work. I'm going to close this. We should continue to fill gaps as we go.

PR #83, opened 2021-06-09
sdk/state: close channel

acharb:
this adds functionality to the state machine for closing a channel

closes https://github.com/stellar/experimental-payment-channels/issues/47

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -1,0 +1,13 @@
+package close
+
+func (c *Channel) CloseStart(iterationNumber int) error {
+	return nil
+}
+
+func (c *Channel) CloseCoordinated(id string) (newStatus string, err error) {
+	return "", nil
+}
+
+func (c *Channel) CloseUncoordinated(id string) error {
+	return nil
+}
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -82,7 +82,7 @@ func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (CoordinatedClose
 		return CoordinatedClose{}, err
 	}
 	if !signed {
-		return CoordinatedClose{}, fmt.Errorf("verifying coordinated close signed by remote: not signed by remote")
+		return CoordinatedClose{}, fmt.Errorf("verifying coordinated close: not signed by remote")
 	}
 
 	// If local has not signed, sign.
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -9,8 +9,9 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-// The steps for a channel coordinated close are as followed (initiator and responder are interchangeable):
-// 1. Initiator submits latest declaration tx
+// The steps for a channel coordinated close are as followed:
+// 1. Initiator or Responder submits latest declaration tx
+// (in steps 2-4 Initiator and Responder are interchangeable, as long as they alternate)
 // 2. Initiator calls ProposeCoordinatedClose
 // 3. Responder calls ConfirmCoordinatedClose
 // 4. Initiator calls ConfirmCoordinatedClose
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -9,6 +9,12 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+// The steps for a channel coordinated close are as followed (initiator and responder are interchangeable):
+// 1. Initiator submits latest declaration tx
+// 2. Initiator calls ProposeCoordinatedClose
+// 3. Responder calls ConfirmCoordinatedClose
+// 4. Initiator calls ConfirmCoordinatedClose
+
 type CoordinatedClose struct {
 	observationPeriodTime      time.Duration
 	observationPeriodLedgerGap int64
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -1,13 +1,41 @@
 package close
 
-func (c *Channel) CloseStart(iterationNumber int) error {
-	return nil
+import (
+	"github.com/stellar/experimental-payment-channels/sdk/txbuild"
+	"github.com/stellar/go/txnbuild"
+)
+
+// TODO - does submitting happen in this method or outside of it? (probably in orchestration)
+func (c *Channel) StartClose() (*txnbuild.Transaction, error) {
+	// submit latest decl
+	decl, err := txbuild.Declaration(txbuild.DeclarationParams{
+		InitiatorEscrow:         c.initiatorEscrowAccount().Address,
+		StartSequence:           c.startingSequence,
+		IterationNumber:         c.latestCloseAgreement.IterationNumber,
+		IterationNumberExecuted: 0,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	decl, err = decl.AddSignatureDecorated(c.latestCloseAgreement.DeclarationSignatures...)
+	if err != nil {
+		return nil, err
+	}
+
+	return decl, nil
 }
 
 func (c *Channel) CloseCoordinated(id string) (newStatus string, err error) {
+	// modify txClose to be able to submit immediately
+	// sign
+	// send to other
 	return "", nil
 }
 
 func (c *Channel) CloseUncoordinated(id string) error {
+	// submit latest txDecl
+	// wait observation period O
+	// subit latest txClose
 	return nil
 }
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -73,7 +73,7 @@ func (c *Channel) CoordinatedCloseTx() (*txnbuild.Transaction, error) {
 	return txClose, nil
 }
 
-// ProposeClose proposes parameters for a close transaction to be submitted earlier.
+// ProposeCoordinatedClose proposes parameters for a close transaction to be submitted earlier.
 // This should be used when participants are in agreement on the final txClose parameters, but would
 // like to submit earlier than the original observation time.
 func (c *Channel) ProposeCoordinatedClose(observationPeriodTime time.Duration, observationPeriodLedgerGap int64) (CoordinatedClose, error) {
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -1,41 +1,141 @@
-package close
+package state
 
 import (
+	"fmt"
+	"time"
+
 	"github.com/stellar/experimental-payment-channels/sdk/txbuild"
 	"github.com/stellar/go/txnbuild"
+	"github.com/stellar/go/xdr"
 )
 
-// TODO - does submitting happen in this method or outside of it? (probably in orchestration)
-func (c *Channel) StartClose() (*txnbuild.Transaction, error) {
-	// submit latest decl
-	decl, err := txbuild.Declaration(txbuild.DeclarationParams{
+type CoordinatedClose struct {
+	observationPeriodTime      time.Duration
+	observationPeriodLedgerGap int64
+	closeSignatures            []xdr.DecoratedSignature
+}
+
+func (cc CoordinatedClose) CloseSignatures() []xdr.DecoratedSignature {
+	return cc.closeSignatures
+}
+
+func (c *Channel) CloseTxs() (*txnbuild.Transaction, *txnbuild.Transaction, error) {
+	txDecl, err := txbuild.Declaration(txbuild.DeclarationParams{
 		InitiatorEscrow:         c.initiatorEscrowAccount().Address,
 		StartSequence:           c.startingSequence,
 		IterationNumber:         c.latestCloseAgreement.IterationNumber,
 		IterationNumberExecuted: 0,
 	})
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
+	txClose, err := txbuild.Close(txbuild.CloseParams{
+		ObservationPeriodTime:      c.observationPeriodTime,
+		ObservationPeriodLedgerGap: c.observationPeriodLedgerGap,
+		InitiatorSigner:            c.initiatorSigner(),
+		ResponderSigner:            c.responderSigner(),
+		InitiatorEscrow:            c.initiatorEscrowAccount().Address,
+		ResponderEscrow:            c.responderEscrowAccount().Address,
+		StartSequence:              c.startingSequence,
+		IterationNumber:            c.latestCloseAgreement.IterationNumber,
+		AmountToInitiator:          c.initiatorClaimAmount(),
+		AmountToResponder:          c.responderClaimAmount(),
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+	return txDecl, txClose, nil
+}
 
-	decl, err = decl.AddSignatureDecorated(c.latestCloseAgreement.DeclarationSignatures...)
+func (c *Channel) CoordinatedCloseTx() (*txnbuild.Transaction, error) {
+	txClose, err := txbuild.Close(txbuild.CloseParams{
+		ObservationPeriodTime:      c.coordinatedClose.observationPeriodTime,
+		ObservationPeriodLedgerGap: c.coordinatedClose.observationPeriodLedgerGap,
+		InitiatorSigner:            c.initiatorSigner(),
+		ResponderSigner:            c.responderSigner(),
+		InitiatorEscrow:            c.initiatorEscrowAccount().Address,
+		ResponderEscrow:            c.responderEscrowAccount().Address,
+		StartSequence:              c.startingSequence,
+		IterationNumber:            c.latestCloseAgreement.IterationNumber,
+		AmountToInitiator:          c.initiatorClaimAmount(),
+		AmountToResponder:          c.responderClaimAmount(),
+	})
 	if err != nil {
 		return nil, err
 	}
-
-	return decl, nil
+	return txClose, nil
 }
 
-func (c *Channel) CloseCoordinated(id string) (newStatus string, err error) {
-	// modify txClose to be able to submit immediately
-	// sign
-	// send to other
-	return "", nil
+// ProposeClose proposes parameters for a close transaction to be submitted earlier.
+// This should be used when participants are in agreement on the final txClose parameters, but would
+// like to submit earlier than the original observation time.
+func (c *Channel) ProposeCoordinatedClose(observationPeriodTime time.Duration, observationPeriodLedgerGap int64) (CoordinatedClose, error) {
+	txCoordinatedClose, err := txbuild.Close(txbuild.CloseParams{
+		ObservationPeriodTime:      observationPeriodTime,
+		ObservationPeriodLedgerGap: observationPeriodLedgerGap,
+		InitiatorSigner:            c.initiatorSigner(),
+		ResponderSigner:            c.responderSigner(),
+		InitiatorEscrow:            c.initiatorEscrowAccount().Address,
+		ResponderEscrow:            c.responderEscrowAccount().Address,
+		StartSequence:              c.startingSequence,
+		IterationNumber:            c.latestCloseAgreement.IterationNumber,
+		AmountToInitiator:          c.initiatorClaimAmount(),
+		AmountToResponder:          c.responderClaimAmount(),
+	})
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+	txCoordinatedClose, err = txCoordinatedClose.Sign(c.networkPassphrase, c.localSigner)
+	if err != nil {
+		return CoordinatedClose{}, nil
+	}
+	return CoordinatedClose{
+		observationPeriodTime:      observationPeriodTime,
+		observationPeriodLedgerGap: observationPeriodLedgerGap,
+		closeSignatures:            txCoordinatedClose.Signatures(),
+	}, nil
 }
 
-func (c *Channel) CloseUncoordinated(id string) error {
-	// submit latest txDecl
-	// wait observation period O
-	// subit latest txClose
-	return nil
+func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (CoordinatedClose, error) {
+	txCoordinatedClose, err := txbuild.Close(txbuild.CloseParams{
+		ObservationPeriodTime:      cc.observationPeriodTime,
+		ObservationPeriodLedgerGap: cc.observationPeriodLedgerGap,
+		InitiatorSigner:            c.initiatorSigner(),
+		ResponderSigner:            c.responderSigner(),
+		InitiatorEscrow:            c.initiatorEscrowAccount().Address,
+		ResponderEscrow:            c.responderEscrowAccount().Address,
+		StartSequence:              c.startingSequence,
+		IterationNumber:            c.latestCloseAgreement.IterationNumber,
+		AmountToInitiator:          c.initiatorClaimAmount(),
+		AmountToResponder:          c.responderClaimAmount(),
+	})
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+
+	// If remote has not signed coordinated close, error as is invalid.
+	signed, err := c.verifySigned(txCoordinatedClose, cc.closeSignatures, c.remoteSigner)
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+	if !signed {
+		return CoordinatedClose{}, fmt.Errorf("verifying coordinated close signed by remote: not signed by remote")
+	}
+
+	// If local has not signed, sign.
+	signed, err = c.verifySigned(txCoordinatedClose, cc.closeSignatures, c.localSigner)
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+	if !signed {
+		txCoordinatedClose, err = txCoordinatedClose.Sign(c.networkPassphrase, c.localSigner)
+		if err != nil {
+			return CoordinatedClose{}, err
+		}
+		cc.closeSignatures = append(cc.closeSignatures, txCoordinatedClose.Signatures()...)
+	}
+
+	// TODO - merge instead of overwrite, similar to ConfirmProposal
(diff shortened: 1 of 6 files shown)


sdk/txbuild/declaration.go
```diff
@@ -19,8 +19,9 @@ func Declaration(p DeclarationParams) (*txnbuild.Transaction, error) {
 			AccountID: p.InitiatorEscrow.Address(),
 			Sequence:  startSequenceOfIteration(p.StartSequence, p.IterationNumber) + 0, // Declaration is the first transaction in an iteration's transaction set.
 		},
-		BaseFee:           0,
-		Timebounds:        txnbuild.NewTimeout(300),
+		BaseFee: 0,
+		// TODO - Timebounds needs to be explicit
+		Timebounds:        txnbuild.NewInfiniteTimeout(),
 		MinSequenceNumber: &minSequenceNumber,
```
acharb, line 25: this change was already made in `txbuild/close.go` , and needed to be copied here as well

sdk/state/state_test.go
```diff
@@ -391,6 +414,177 @@ func Test(t *testing.T) {
...
+		require.True(t, fullySigned)
+		ci, di, err := sendingChannel.PaymentTxs(payment)
+		require.NoError(t, err)
+		ci, err = ci.AddSignatureDecorated(payment.CloseSignatures...)
+		require.NoError(t, err)
+		di, err = di.AddSignatureDecorated(payment.DeclarationSignatures...)
+		require.NoError(t, err)
+	}
+
+	// Coordinated Close
```
acharb, line 529: there are a lot of changes in this test file. But most of it is copying code/refactoring. The notable changes are in this section (testing a coordinated close).

Uncoordinated close is already being tested in `TestOpenUpdatesUncoordinatedClose`

sdk/state/state.go
```diff
@@ -150,36 +174,6 @@ func (c *Channel) verifySigned(tx *txnbuild.Transaction, sigs []xdr.DecoratedSig
 	return false, nil
 }
 
-func (c *Channel) CloseStart(iterationNumber int) error {
```
acharb, line 153: this is removing functions that were originally added as part of the skeleton sdk. Now can be removed as the actual functions are being implemented

sdk/state/state_test.go
```diff
@@ -17,29 +18,34 @@ import (
...
+		os.Exit(1)
 	}
+	networkPassphrase = networkDetails.NetworkPassphrase
 
-	// Setup initiator.
-	initiator := Participant{
+	os.Exit(m.Run())
+}
+
+func initAccounts(t *testing.T) (initiator Participant, responder Participant) {
```
nikhilsaraf, line 47: style: I'm generally discouraging of naming the return types in the signature (would rather use a comment), because that requires you to do a `=` assignment instead of a `:=` assignment which requires a code reader to check if that variable has been assigned anywhere above and whether it is used etc, essentially taking away from the "don't reuse variables" philosophy.

(another code style pattern that we don't have to adopt but thought I'd bring up in case we find it useful and choose to adopt)

sdk/state/state_test.go
```diff
@@ -391,6 +414,177 @@ func Test(t *testing.T) {
...
+		_, err = client.SubmitFeeBumpTransaction(fbtx)
+		require.NoError(t, err)
+	}
+
+	// Perform a number of iterations, much like two participants may.
+	// Exchange signed C_i and D_i for each
+	t.Log("Subsequent agreements...")
+	rBalanceCheck := responder.Contribution
+	iBalanceCheck := initiator.Contribution
+	for i < 20 {
```
nikhilsaraf, line 476: why `20`?
can we put in a named variable?

sdk/state/state_test.go
```diff
@@ -391,6 +414,177 @@ func Test(t *testing.T) {
...
+		i++
+		require.Equal(t, i, initiatorChannel.NextIterationNumber())
+		require.Equal(t, i, responderChannel.NextIterationNumber())
+		amount := randomPositiveInt64(t, 100_0000000)
```
nikhilsaraf, line 480: similarly here too, can we put the number in a named variable. helps with readability

sdk/state/state_test.go
nikhilsaraf, line 476: @acharb this seems like non-deterministic testing, which doesn't make it very easy to replicate if it every fails.

Is there a way to make this into a deterministic testing pattern?
If yes, then can we add a TODO to convert this to a deterministic testing pattern?

tagging @leighmcculloch to get your thoughts as well.

sdk/state/state_test.go
```diff
@@ -391,6 +414,177 @@ func Test(t *testing.T) {
 	assert.EqualValues(t, initiatorEscrowResponse.Balances[0].Balance, fmt.Sprintf("%.7f", float64(iBalanceCheck)/float64(1_000_0000)))
 }
 
+func TestOpenUpdatesCoordinatedClose(t *testing.T) {
```
nikhilsaraf, line 417: @acharb this test seems very long which makes it hard to follow, and much much harder to maintain.

not a blocker, but if we can break it up in a future PR soon then that would make it a little easier to follow/maintain. It may be a little more complicated to do and we may need to invest in developing a good test framework, but something we should discuss with @leighmcculloch once he's back.

sdk/state/state_test.go
```diff
@@ -391,6 +414,177 @@ func Test(t *testing.T) {
...
+	require.NoError(t, err)
+	fbtx, err = fbtx.Sign(networkPassphrase, initiator.KP)
+	require.NoError(t, err)
+	_, err = client.SubmitFeeBumpTransaction(fbtx)
+	require.NoError(t, err)
+
+	t.Log("Initiator proposes a coordinated close")
+	cc, err := initiatorChannel.ProposeCoordinatedClose(0, 0)
+	require.NoError(t, err)
+	cc, err = responderChannel.ConfirmCoordinatedClose(cc)
```
nikhilsaraf, line 551: 💯  on this style of testing remote/local channels without requiring networking logic. 🎉

sdk/state/state_test.go
```diff
@@ -391,6 +414,177 @@ func Test(t *testing.T) {
...
+		BaseFee:    txnbuild.MinBaseFee,
+	})
+	require.NoError(t, err)
+	fbtx, err = fbtx.Sign(networkPassphrase, initiator.KP)
+	require.NoError(t, err)
+	_, err = client.SubmitFeeBumpTransaction(fbtx)
+	if err != nil {
+		hErr := horizonclient.GetError(err)
+		t.Log("Submitting Close:", txCoordinated.SourceAccount().Sequence, "Error:", err)
+		t.Log(hErr.ResultString())
```
nikhilsaraf, line 572: do we need to return if this errors out, or is it meaningful to continue through the end of this test?

sdk/state/state_test.go
```diff
@@ -391,6 +414,177 @@ func Test(t *testing.T) {
...
+	}
+	t.Log("Coordinated close successful")
+
+	// check final escrow fund amounts are correct
+	accountRequest := horizonclient.AccountRequest{AccountID: responder.Escrow.Address()}
+	responderEscrowResponse, err := client.AccountDetail(accountRequest)
```
nikhilsaraf, line 578: just noticed that we are making a call to horizon. It is best if tests can avoid making calls to an external network, because we don't know if that network will be available during the running of this test.

do we need to make this call or can we check if the value is equal to what we should have initialized it to in the beginning of the test, without making the call to horizon?

nikhilsaraf (approved):
Added a bunch of comments.

None of them are blocking so marked as approved. But let's discuss in person, may be quickest.

sdk/state/close.go
```diff
@@ -0,0 +1,148 @@
...
+	observationPeriodTime      time.Duration
+	observationPeriodLedgerGap int64
+	closeSignatures            []xdr.DecoratedSignature
+}
+
+func (cc CoordinatedClose) CloseSignatures() []xdr.DecoratedSignature {
+	return cc.closeSignatures
+}
+
+func (c *Channel) CloseTxs() (*txnbuild.Transaction, *txnbuild.Transaction, error) {
```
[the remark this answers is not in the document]
acharb: good idea. I haven't seen comment vars used in the sdf go code before, but I added named return variables. I'd be happy to discuss further though
https://github.com/stellar/experimental-payment-channels/pull/83/commits/a08e853171efce60fac09a2fa59077578749756b

sdk/state/close.go
```diff
@@ -0,0 +1,148 @@
...
+	txClose, err := txbuild.Close(txbuild.CloseParams{
+		ObservationPeriodTime:      c.observationPeriodTime,
+		ObservationPeriodLedgerGap: c.observationPeriodLedgerGap,
+		InitiatorSigner:            c.initiatorSigner(),
+		ResponderSigner:            c.responderSigner(),
+		InitiatorEscrow:            c.initiatorEscrowAccount().Address,
+		ResponderEscrow:            c.responderEscrowAccount().Address,
+		StartSequence:              c.startingSequence,
+		IterationNumber:            c.latestCloseAgreement.IterationNumber,
+		AmountToInitiator:          c.initiatorClaimAmount(),
```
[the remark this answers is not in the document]
acharb: ya  you're right there will most likely need to be locking. I think it would be best to add that when we start thinking about networking, since that's when it will most be needed.

I added a ticket to track that thought: https://github.com/stellar/experimental-payment-channels/issues/84

sdk/state/close.go
```diff
@@ -0,0 +1,148 @@
...
+		AmountToResponder:          c.responderClaimAmount(),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return txClose, nil
+}
+
+// ProposeClose proposes parameters for a close transaction to be submitted earlier.
```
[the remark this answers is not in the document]
acharb: https://github.com/stellar/experimental-payment-channels/pull/83/commits/e48d2ad592400d201bf2a94a78ff2db5af4cf5fd

sdk/state/close.go
```diff
@@ -0,0 +1,148 @@
...
+		ResponderEscrow:            c.responderEscrowAccount().Address,
+		StartSequence:              c.startingSequence,
+		IterationNumber:            c.latestCloseAgreement.IterationNumber,
+		AmountToInitiator:          c.initiatorClaimAmount(),
+		AmountToResponder:          c.responderClaimAmount(),
+	})
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+	txCoordinatedClose, err = txCoordinatedClose.Sign(c.networkPassphrase, c.localSigner)
```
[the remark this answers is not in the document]
acharb: good point. Personally I think changing names for the same struct adds some reading friction, eg. `txCoordinatedClose` vs `txCoordiantedCloseSigned`. I'd be happy to discuss though. We've been using the same name vars so far

sdk/state/close.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return CoordinatedClose{
+		observationPeriodTime:      observationPeriodTime,
+		observationPeriodLedgerGap: observationPeriodLedgerGap,
+		closeSignatures:            txCoordinatedClose.Signatures(),
+	}, nil
+}
+
+func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (CoordinatedClose, error) {
+	txCoordinatedClose, err := txbuild.Close(txbuild.CloseParams{
```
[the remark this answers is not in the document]
acharb: I think that will help with readability 👍  added
https://github.com/stellar/experimental-payment-channels/pull/83/commits/a5df1ce03c8f422f0320f31ed6eafc3b69cd33ac

sdk/state/close.go
```diff
@@ -0,0 +1,148 @@
...
+		return CoordinatedClose{}, err
+	}
+
+	// If remote has not signed coordinated close, error as is invalid.
+	signed, err := c.verifySigned(txCoordinatedClose, cc.closeSignatures, c.remoteSigner)
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+	if !signed {
+		return CoordinatedClose{}, fmt.Errorf("verifying coordinated close signed by remote: not signed by remote")
```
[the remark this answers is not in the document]
acharb: this is currently the pattern we've adopted for errors so far (eg. if you look in `update.go`), but I can see it being confusing 🤔  basically we're trying to mimic wrapped errors. Probably near the finish line we can look at verification errors more holistically to come up with a better/clearer pattern. I changed this one to read clearer though
https://github.com/stellar/experimental-payment-channels/pull/83/commits/37cfbc4afdde35326c55276de00cbeeb10996d7d

sdk/state/close.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// If local has not signed, sign.
+	signed, err = c.verifySigned(txCoordinatedClose, cc.closeSignatures, c.localSigner)
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+	if !signed {
```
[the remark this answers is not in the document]
acharb: So the general steps are:

```
// 1. Initiator or Responder submits latest declaration tx
// (in steps 2-4 Initiator and Responder are interchangeable, as long as they alternate)
// 2. Initiator calls ProposeCoordinatedClose
// 3. Responder calls ConfirmCoordinatedClose
// 4. Initiator calls ConfirmCoordinatedClose
```
So in step 4 initiator (local) has already signed. They sign it when they propose it.

This shouldn't be a bug though. If you're the initiator, when you propose the new `txClose` you sign it and send to the responder. If the responder submits it, great that's what you wanted. If responder doesn't, then you can still just wait for `O` to pass and submit the original `txClose` that you already have fully signed.

the coordinated close has the same data as the latest `txClose` just with the shortened `O` period

happy to discuss further

sdk/state/state.go
```diff
@@ -133,6 +143,20 @@ func (c *Channel) responderSigner() *keypair.FromAddress {
 	}
 }
 
+func (c *Channel) initiatorClaimAmount() int64 {
+	if c.latestCloseAgreement.Balance.Amount < 0 {
+		return c.latestCloseAgreement.Balance.Amount * -1
```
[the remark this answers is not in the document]
acharb: correct if the channel balance is positive it means owed to responder. If it's negative it means owed to initiator. We added a comment for that here: https://github.com/stellar/experimental-payment-channels/blob/a946b6b9b45735f7b4270542ec0f699c06a5c699/sdk/state/state.go#L83

sdk/state/state_test.go
acharb: I see your point thx for pointing out 🤔 . I'd be happy to discuss but the comment solution feels weird to me. I might be just not used to it though. I think when Leigh gets back who has more golang experience than I he can give his input here.

I think I'll leave as is for now if alright 😬

sdk/state/state_test.go
acharb: 20 just felt like a good number of iterations to make sure the back and forth updates is working correctly. I made it a named var though:
https://github.com/stellar/experimental-payment-channels/pull/83/commits/e08f796e1f3771b5b549865c01737ba702d9cba7

sdk/state/state_test.go
acharb: added a comment, I think that should be clearer 😄 
https://github.com/stellar/experimental-payment-channels/pull/83/commits/d48f3be031f4512fbed2f78979b6259366b4a724

sdk/state/state_test.go
acharb: i don't feel strongly here. The only non-deterministic parts of these integration tests are the payment amounts and that we're calling a local standalone network. I think for this use case and stage of the project that makes sense, but I'd be happy to discuss further. 

We have a ticket [here](https://github.com/stellar/experimental-payment-channels/issues/70) for adding unit testing which won't rely on a network and can have non-random payment amounts to be deterministic

sdk/state/state_test.go
acharb: I agree. This state_test.go can be cleaned up. I couldn't figure out a good way to write this second test w/o duplicating a lot of the code in the original test. 

I think once the majority of the sdf/state code is written this integration test can be broken up / looked at holistically and made more modular so there's less dup code. I didn't think now is the right time to do that though since stuff is changing frequently because early in the project

sdk/state/state_test.go
acharb: oops good catch, we should end it
https://github.com/stellar/experimental-payment-channels/pull/83/commits/5fa3ee4c71d32837169639010cf138ec38b71988

sdk/state/state_test.go
acharb: for this test file we're doing an integration test with a standalone network that includes CAP-21. This makes sure the transactions are formatted correct with correct ops/values and can be submitted successfully, among other things like making sure sequence numbers are correct.

We have a [ticket](https://github.com/stellar/experimental-payment-channels/issues/70) for adding unit tests, which won't depend on a network call

acharb:
@nikhilsaraf got over eager and just responded to all your comments now 😄 . But happy to discuss still, lmk if you want to. Going to merge this in now though

sdk/state/close.go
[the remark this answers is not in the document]
nikhilsaraf: haha I like having those two names being separated like in your example :)
I'm ok with the way it is and don't want to push too hard if we're reused across the codebase.

sdk/state/close.go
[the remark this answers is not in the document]
nikhilsaraf: that makes sense since it's ok for responder to submit earlier since this is a coordinated close.

wondering if it's already signed in step 2, then why do we need to check if it is !signed in step 4?
(basically I'm wondering if we are checking for a state that isn't possible. and if that state occurs then it's a programmer error). can discuss on call.

sdk/state/close.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/stellar/experimental-payment-channels/sdk/txbuild"
+	"github.com/stellar/go/txnbuild"
+	"github.com/stellar/go/xdr"
+)
+
+// The steps for a channel coordinated close are as followed:
+// 1. Initiator or Responder submits latest declaration tx
+// (in steps 2-4 Initiator and Responder are interchangeable, as long as they alternate)
+// 2. Initiator calls ProposeCoordinatedClose
```
leighmcculloch, line 15: Why does the initiator have to call this step? Either participant should be able to take the first step.

sdk/state/close.go
```diff
@@ -0,0 +1,120 @@
...
+// 3. Responder calls ConfirmCoordinatedClose
+// 4. Initiator calls ConfirmCoordinatedClose
+
+type CoordinatedClose struct {
+	observationPeriodTime      time.Duration
+	observationPeriodLedgerGap int64
```
leighmcculloch, line 21: These periods of time should be zero for a coordinated close. Why are they required?

sdk/state/close.go
```diff
@@ -0,0 +1,120 @@
...
+	closeSignatures            []xdr.DecoratedSignature
```
leighmcculloch, line 22: The other objects we have for Payment, Open, keep all their fields exported so that they're inspectable, and serializable. We should do the same here too I think, at least for consistency, but also so we have the same benefits and pattern for use.

sdk/state/close.go
```diff
@@ -0,0 +1,120 @@
...
+func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (CoordinatedClose, error) {
+	txCoordinatedClose, err := c.makeCloseTx(cc.observationPeriodTime, cc.observationPeriodLedgerGap)
+	if err != nil {
+		return CoordinatedClose{}, err
+	}
+
+	// If remote has not signed coordinated close, error as is invalid.
+	signed, err := c.verifySigned(txCoordinatedClose, cc.closeSignatures, c.remoteSigner)
+	if err != nil {
+		return CoordinatedClose{}, err
```
leighmcculloch, line 82: We should wrap `err` here with `fmt.Errorf` like we do in other places so that if an error occurs we know what was happening when it occurred.

E.g.: https://github.com/stellar/experimental-payment-channels/blob/c30a604c216a5d8015e20309f3aad4d20ef67fb6/sdk/state/open.go#L120

sdk/state/close.go
```diff
@@ -0,0 +1,120 @@
...
+		return CoordinatedClose{}, nil
+	}
+	return CoordinatedClose{
+		observationPeriodTime:      observationPeriodTime,
+		observationPeriodLedgerGap: observationPeriodLedgerGap,
+		closeSignatures:            txCoordinatedClose.Signatures(),
+	}, nil
+}
+
+func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (CoordinatedClose, error) {
```
leighmcculloch, line 73: Can we return a fullySigned bool like we do in other propose/confirm function pairs?

E.g.:
https://github.com/stellar/experimental-payment-channels/blob/c30a604c216a5d8015e20309f3aad4d20ef67fb6/sdk/state/open.go#L109

sdk/state/state.go
```diff
@@ -43,6 +43,8 @@ type Channel struct {
 
 	latestCloseAgreement     CloseAgreement
 	latestUnconfirmedPayment Payment
+
+	coordinatedClose CoordinatedClose
```
leighmcculloch, line 47: Coordinating a close is really just modifying the previous close agreement. I think it would be helpful because of that if we store the signatures for it in a CloseAgreement in the latestCloseAgreement. This way the last close is still there in the same place, it can just be executed early.

I think this would require adding the observation periods to the close agreement, and so merging the CoordinatedClose and CloseAgreement types.

Wdyt?

sdk/state/state.go
```diff
@@ -133,6 +143,20 @@ func (c *Channel) responderSigner() *keypair.FromAddress {
 	}
 }
 
+func (c *Channel) initiatorClaimAmount() int64 {
```
leighmcculloch, line 146: We should name these functions `initiatorBalanceAmount` using the term Balance instead of Claim, since we don't use the term claim anywhere else.

sdk/state/state_test.go
leighmcculloch: > w/o duplicating a lot of the code

We don't need to keep our tests DRY. Let's keep tests isolated from each other. It makes it easier to delete them when we don't need them, or for them to evolve independently of each other.

Also, this is likely to get better over time because the more functionality we add to the SDK, the tests will get naturally smaller as functionality moves out of these integration tests into the SDK.

leighmcculloch (commented):
Looks great. I left a few comments below.

sdk/state/close.go
acharb: ya  it can be either, above step 2 in parentheses it says that. I'll move it down a line so that sticks out more

sdk/state/close.go
acharb: will a coordinated close always have a zero observation time? I wasn't sure if a shorter O (not necessarily zero) should be allowed

looking at the sep again though that it what is says. I can change

sdk/state/state.go
acharb: But there would be two kinds of closeSignatures in that case right? eg:
```
type CloseAgreement struct {
	IterationNumber       int64
	Balance               Amount

        // new:
	CoordinatedCloseSignature []xdr.DecoratedSignature
}
```
That felt awkward to me when I tried that, so created another type for separation of concerns.  🤔

sdk/state/close.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/95/commits/7b8efe8cb40017fd90aca14577e7effca343155b

sdk/state/close.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/95/commits/ac3ee05b8c4e42b0d3eec6522f5d4339dd26c6d2

sdk/state/state.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/95/commits/02f9d38f37cd5e88b3d0c739b0bb9debb6753faf

sdk/state/close.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/95/commits/66759128404aae82555bbdc864396e7448c13b65

sdk/state/close.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/95/commits/e00b88157d4faf1510792a74ce340487cdec34fc

sdk/state/close.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/95/commits/b5083d0128fb1a48da68137f412f009fe198d3ee

sdk/state/state.go
leighmcculloch: Hmm, good point. We'd also be mutating an object that conceptually was locked in previously.

I think the thing I'm stuck on is that the channel has only two states we should be tracking:
- The most recent fully signed agreement.
- The agreement we're considering right now, and that agreement is either a new agreement (a new payment) of the adjustment of some attribute of the most recent agreement.

If the `lastUnconfirmedAgreement` was a `CloseAgreement` rather than a `Payment`, and if the `CloseAgreement` stored the observation periods then that would probably work. 

Storing them separately works too, but I worry we've now go to check in a few places if a coordinate close or a payment is in progress, because if either is in progress we shouldn't let the other begin, that could be dangerous.

acharb: > Storing them separately works too, but I worry we've now go to check in a few places if a coordinate close or a payment is in progress, because if either is in progress we shouldn't let the other begin, that could be dangerous.

that's true. I'm changing the close code to work without a `CoordinatedClose` type to see how it looks. I'll have a PR out

PR #87, opened 2021-06-11
sdk/state: move integration tests into integration folder, move functions into helper file

acharb:
No code changes here, only moving code to different files/folders for clearer organization. Integration tests will be in the `sdk/state/integration` folder. Unit tests in the `sdk/state` folder (right now `update_test.go` is the only one)

also moves helper functions into their own `integration/helpers.go` file.


nikhilsaraf (approved):
LGTM, thanks! 🎉

Issue #82, opened 2021-06-07
sdk/state: instead of overwriting latestUnconfirmedPayment clone/merge so we don't lose signatures


acharb:
problem with overwriting is if we're given a payment with missing signatures we will lose the ones we have stored. Need a way to extract the new signatures then merge into our existing ones instead of overwriting the whole object.

see comment here: https://github.com/stellar/experimental-payment-channels/pull/81#discussion_r645923209

PR #88, opened 2021-06-14
sdk/state: merge new signatures into stored signature slice instead of overwriting

acharb:
adds functionality for merging new signatures into stored signatures instead of overwriting. This process is done in the close and update steps, when a user receives signatures from the other party and needs to store them locally

closes https://github.com/stellar/experimental-payment-channels/issues/82

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -91,7 +91,7 @@ func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (coordinatedClose
 	}
 	fullySigned = true
 
-	c.coordinatedClose.CloseSignatures = appendNewSignatures(c.coordinatedClose.CloseSignatures, cc.CloseSignatures)
+	c.coordinatedClose = CoordinatedClose{CloseSignatures: appendNewSignatures(c.coordinatedClose.CloseSignatures, cc.CloseSignatures)}
 	return c.coordinatedClose, fullySigned, nil
 }
 
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -20,16 +20,6 @@ type CoordinatedClose struct {
 	CloseSignatures []xdr.DecoratedSignature
 }
 
-// mergeCoordinatedCloseData merges the data from a new coordinated close into an existing one. The signatures of the existing
-// coordinated close are appended to so that existing signatures are not lost.
-func mergeCoordinatedCloseData(cc CoordinatedClose, newCoordinatedClose CoordinatedClose) CoordinatedClose {
-	return CoordinatedClose{
-		observationPeriodTime:      newCoordinatedClose.observationPeriodTime,
-		observationPeriodLedgerGap: newCoordinatedClose.observationPeriodLedgerGap,
-		closeSignatures:            appendNewSignatures(cc.closeSignatures, newCoordinatedClose.closeSignatures),
-	}
-}
-
 func (c *Channel) CloseTxs() (txDecl *txnbuild.Transaction, txClose *txnbuild.Transaction, err error) {
 	txDecl, err = txbuild.Declaration(txbuild.DeclarationParams{
 		InitiatorEscrow:         c.initiatorEscrowAccount().Address,
@@ -101,8 +91,8 @@ func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (coordinatedClose
 	}
 	fullySigned = true
 
-	c.coordinatedClose = mergeCoordinatedCloseData(c.coordinatedClose, cc)
-	return cc, nil
+	c.coordinatedClose.CloseSignatures = appendNewSignatures(c.coordinatedClose.CloseSignatures, cc.CloseSignatures)
+	return c.coordinatedClose, fullySigned, nil
 }
 
 // makeCloseTx is a helper method for creating a close transaction with custom observation values.
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -26,6 +26,16 @@ func (cc CoordinatedClose) CloseSignatures() []xdr.DecoratedSignature {
 	return cc.closeSignatures
 }
 
+// mergeCoordinatedCloseData merges the data from a new coordinated close into an existing one. The signatures of the existing
+// coordinated close are appended to so that existing signatures are not lost.
+func mergeCoordinatedCloseData(cc CoordinatedClose, newCoordinatedClose CoordinatedClose) CoordinatedClose {
+	return CoordinatedClose{
+		observationPeriodTime:      newCoordinatedClose.observationPeriodTime,
+		observationPeriodLedgerGap: newCoordinatedClose.observationPeriodLedgerGap,
+		closeSignatures:            appendNewSignatures(cc.closeSignatures, newCoordinatedClose.closeSignatures),
+	}
+}
+
 func (c *Channel) CloseTxs() (txDecl *txnbuild.Transaction, txClose *txnbuild.Transaction, err error) {
 	txDecl, err = txbuild.Declaration(txbuild.DeclarationParams{
 		InitiatorEscrow:         c.initiatorEscrowAccount().Address,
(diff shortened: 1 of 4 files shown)


sdk/state/close.go
```diff
@@ -26,6 +26,16 @@ func (cc CoordinatedClose) CloseSignatures() []xdr.DecoratedSignature {
 	return cc.closeSignatures
 }
 
+// mergeCoordinatedCloseData merges the data from a new coordinated close into an existing one. The signatures of the existing
+// coordinated close are appended to so that existing signatures are not lost.
+func mergeCoordinatedCloseData(cc CoordinatedClose, newCoordinatedClose CoordinatedClose) CoordinatedClose {
```
acharb, line 31: I'm not tied to this naming using "merge". Some of the data fields are overwritten, and then the signatures are the ones actually merged so none are lost. But I couldn't think of a better name for the method

sdk/state/close.go
```diff
@@ -98,8 +108,7 @@ func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (CoordinatedClose
 		cc.closeSignatures = append(cc.closeSignatures, txCoordinatedClose.Signatures()...)
 	}
 
-	// TODO - merge instead of overwrite, similar to ConfirmProposal
-	c.coordinatedClose = cc
+	c.coordinatedClose = mergeCoordinatedCloseData(c.coordinatedClose, cc)
 	return cc, nil
```
leighmcculloch, line 112: We should return the merged result so that we're returning the final and complete state that we have.

sdk/state/close.go
leighmcculloch: I think what you highlight isn't an issue with the function name but the function responsibilities. I think we should call `appendNewSignatures` directly.

sdk/state/close.go
```diff
@@ -98,8 +108,7 @@ func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (CoordinatedClose
 		cc.closeSignatures = append(cc.closeSignatures, txCoordinatedClose.Signatures()...)
 	}
 
-	// TODO - merge instead of overwrite, similar to ConfirmProposal
-	c.coordinatedClose = cc
+	c.coordinatedClose = mergeCoordinatedCloseData(c.coordinatedClose, cc)
```
leighmcculloch, line 111: We need this same behavior on all the other confirm functions too. Not sure if you're planning that for a separate PR, but I think you could do it here too.

leighmcculloch (commented):
Couple requests inline.

sdk/state/close.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/88/commits/ec2b6306997dc887ef74c30809a0a1c40d822b5a

sdk/state/close.go
acharb: removed the merge methods here: https://github.com/stellar/experimental-payment-channels/pull/88/commits/ec2b6306997dc887ef74c30809a0a1c40d822b5a

definitely makes it cleaner since we removed the observation fields from the `CoordinatedClosed` struct

sdk/state/close.go
acharb: I added for the `Update` and `Close` confirm methods. I believe those are the only confirm methods it's needed atm

For the `Open` steps we're actually not storing signatures atm. I added this ticket: https://github.com/stellar/experimental-payment-channels/issues/92 for changing that to match the `Update` and `Close` steps. For that ticket I can add this `appendNewSignatures` functionality (assuming you agree the `Open` should start storing sigs, if you could confirm in that ticket that would be great 😄 )

sdk/state/close.go
```diff
@@ -91,9 +91,8 @@ func (c *Channel) ConfirmCoordinatedClose(cc CoordinatedClose) (coordinatedClose
 	}
 	fullySigned = true
 
-	// TODO - merge instead of overwrite, similar to ConfirmProposal
-	c.coordinatedClose = cc
-	return cc, fullySigned, nil
+	c.coordinatedClose.CloseSignatures = appendNewSignatures(c.coordinatedClose.CloseSignatures, cc.CloseSignatures)
```
leighmcculloch, line 94: We never seem to set `c.coordinatedClose` in the participant who doesn't call propose. We only appear to modify it. I guess it doesn't have any other fields, but I think we should still do the following, and I expect in the future we might need to store other things in it?

```suggestion
	c.coordinatedClose = CoordinatedClose{CloseSignatures: appendNewSignatures(c.coordinatedClose.CloseSignatures, cc.CloseSignatures)}
```

If you feel strongly against this feel free to move ahead without it.

sdk/state/update_test.go
```diff
@@ -80,3 +82,63 @@ func TestLastConfirmedPayment(t *testing.T) {
...
+}
+
+func randomByteArray(length int) ([]byte, error) {
+	arr := make([]byte, length)
+	_, err := rand.Read(arr)
+	if err != nil {
+		return nil, err
+	}
+	return arr, nil
+}
```
leighmcculloch, line 144: 🤔 It's not clear to me what this test is testing. There's a lot of setup relating to channels, but then the functions that are called are just `appendNewSignatures` on groups of signatures. If this function is only intended to test `appendNewSignatures` then you don't need any of the `Channel` setup.

Would the below achieve the same thing you're going for?

```suggestion
func TestAppendNewSignature(t *testing.T) {
	closeSignatures := []xdr.DecoratedSignature{
		{ Signature: randomByteArray(10) },
		{ Signature: randomByteArray(10) },
	}
	
	closeSignaturesToAppend := []xdr.DecoratedSignature{
		closeSignatures[0], // A duplicate signature is included.
		{ Signature: randomByteArray(10) },
	}

	newCloseSignatures := appendNewSignatures(closeSignatures, closeSignaturesToAppend)
	
	// Check that the final slice of signatures does not contain the duplicate.
	assert.ElementsMatch(
		t,
		newCloseSignatures,
		[]xdr.DecoratedSignature{
			closeSignatures[0],
			closeSignatures[1],
			closeSignaturesToAppend[1],
		},
	)
}

func randomByteArray(t *testing.T, length int) []byte {
	arr := make([]byte, length)
	_, err := rand.Read(arr)
	require.NoError(t, err)
	return arr
}
```

sdk/state/state.go
```diff
@@ -173,6 +173,20 @@ func (c *Channel) verifySigned(tx *txnbuild.Transaction, sigs []xdr.DecoratedSig
...
 }
 
+func appendNewSignatures(oldSignatures []xdr.DecoratedSignature, newSignatures []xdr.DecoratedSignature) []xdr.DecoratedSignature {
+	m := make(map[string]bool)
+	for _, os := range oldSignatures {
+		m[string(os.Signature)] = true
+	}
+
+	for _, ns := range newSignatures {
+		if _, found := m[string(ns.Signature)]; !found {
```
leighmcculloch, line 183: Because you've made the slice value a bool, you can simply do:
```suggestion
		if !m[string(ns.Signature)] {
```

leighmcculloch (commented):
I left a couple comments inline.

I think it would also be helpful if we added a test that illustrated what scenario we're trying to prevent with this change. The scenario is that if two parties are passing back and forth say a payment, it shouldn't be possible for one party to delete a prior signature in a later step and cause the client to lose a signature.

sdk/state/close.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/88/commits/899c9a46671dc43ca318575f74818f2d0d351a7c

probably this will be changed again in the PR I'm making for not using CoordinatedClose. But atm I agree this is more correct

sdk/state/update_test.go
acharb: ya  I think I was too tied to using the channel instead of focusing on the append method. This is a great simplification. Also `assert.ElementsMatch` 🙌 
https://github.com/stellar/experimental-payment-channels/pull/88/commits/39ddf70ab299af6f2f69e5e64739be0915a88405

sdk/state/state.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/88/commits/35496032fe6eb27173c1ac3ea4bd2ca6cb2823e5

acharb:
> I think it would also be helpful if we added a test that illustrated what scenario we're trying to prevent with this change. The scenario is that if two parties are passing back and forth say a payment, it shouldn't be possible for one party to delete a prior signature in a later step and cause the client to lose a signature.

I added in the unit test: https://github.com/stellar/experimental-payment-channels/pull/88/commits/39ddf70ab299af6f2f69e5e64739be0915a88405 to make sure appending signatures doesn't lose any. Imo that should suffice unless you think an integration test covering this adds value

leighmcculloch (approved):

Issue #71, opened 2021-06-03
sdk/state: support non-native asset


leighmcculloch:
The channel state machine supports a single native asset balance but should a non-native asset. The txbuild functions may also need expanding to support this change. The tests should be updated to use a non-native asset as well.

acharb:
was completed with this PR: https://github.com/stellar/experimental-payment-channels/pull/87

PR #89, opened 2021-06-15
sdk/state: form channel and close channel with non-native currency

acharb:
this is the core change for [this ticket](https://github.com/stellar/experimental-payment-channels/issues/71) of supporting non-native currencies.

Steps in this PR:
(basically the initial changes in `txbuild`, which will be used by `sdk/state`)
1. change txbuild/close to create a closing channel transaction that has payments in non-native asset.
2. change txbuild/formation to configure the channel to support non-native asset

Next steps (not in this PR):
1. Escrow account creation needs trustlines for non-native assets
2. Change the open channel functionality to be able to custom choose a non-native asset
2. The `latestCloseAgreement` `lastUnconfirmedPayment` need to store what the channel asset is
3. When closing the channel the payments need to be able to be non-native assets

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -116,5 +116,6 @@ func (c *Channel) makeCloseTx(observationPeriodTime time.Duration, observationPe
 		IterationNumber:            c.latestCloseAgreement.IterationNumber,
 		AmountToInitiator:          c.initiatorClaimAmount(),
 		AmountToResponder:          c.responderClaimAmount(),
+		Asset:                      c.latestCloseAgreement.Balance.Asset,
 	})
 }
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -26,6 +26,8 @@ func (c *Channel) OpenTxs() (txClose, txDecl, formation *txnbuild.Transaction, e
 		IterationNumber:            1,
 		AmountToInitiator:          0,
 		AmountToResponder:          0,
+		// TODO - change to non native
+		Asset: NativeAsset{},
 	})
 	if err != nil {
 		return
@@ -45,6 +47,8 @@ func (c *Channel) OpenTxs() (txClose, txDecl, formation *txnbuild.Transaction, e
 		InitiatorEscrow: c.initiatorEscrowAccount().Address,
 		ResponderEscrow: c.responderEscrowAccount().Address,
 		StartSequence:   c.startingSequence,
+		// TODO - change to non native
+		Asset: NativeAsset{},
 	})
 	return
 }
--- a/sdk/txbuild/close.go
+++ b/sdk/txbuild/close.go
@@ -19,6 +19,7 @@ type CloseParams struct {
 	IterationNumber            int64
 	AmountToInitiator          int64
 	AmountToResponder          int64
+	Asset                      txnbuild.Asset
 }
 
 func Close(p CloseParams) (*txnbuild.Transaction, error) {
@@ -55,7 +56,7 @@ func Close(p CloseParams) (*txnbuild.Transaction, error) {
 		tp.Operations = append(tp.Operations, &txnbuild.Payment{
 			SourceAccount: p.ResponderEscrow.Address(),
 			Destination:   p.InitiatorEscrow.Address(),
-			Asset:         txnbuild.NativeAsset{},
+			Asset:         p.Asset,
 			Amount:        amount.StringFromInt64(p.AmountToInitiator),
 		})
 	}
@@ -63,7 +64,7 @@ func Close(p CloseParams) (*txnbuild.Transaction, error) {
 		tp.Operations = append(tp.Operations, &txnbuild.Payment{
 			SourceAccount: p.InitiatorEscrow.Address(),
 			Destination:   p.ResponderEscrow.Address(),
-			Asset:         txnbuild.NativeAsset{},
+			Asset:         p.Asset,
 			Amount:        amount.StringFromInt64(p.AmountToResponder),
 		})
 	}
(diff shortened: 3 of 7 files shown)


sdk/txbuild/formation.go
```diff
@@ -35,6 +37,13 @@ func Formation(p FormationParams) (*txnbuild.Transaction, error) {
 		SourceAccount: p.InitiatorEscrow.Address(),
 		Signer:        &txnbuild.Signer{Address: p.InitiatorSigner.Address(), Weight: 1},
 	})
+	if !p.Asset.IsNative() {
```
acharb, line 40: the trustlines for escrow accounts is created in the Formation transaction.

see the [SEP setup steps](https://github.com/stellar/experimental-payment-channels/blob/main/specifications/sep-payment-channel-transactions.md#setup) under the formation transaction part

sdk/txbuild/close.go
```diff
@@ -55,15 +56,15 @@ func Close(p CloseParams) (*txnbuild.Transaction, error) {
 		tp.Operations = append(tp.Operations, &txnbuild.Payment{
 			SourceAccount: p.ResponderEscrow.Address(),
 			Destination:   p.InitiatorEscrow.Address(),
-			Asset:         txnbuild.NativeAsset{},
+			Asset:         p.Asset,
```
acharb, line 59: now we can choose a non-native asset by selecting the field in `CloseParams`

nikhilsaraf (approved):
This is extremely digestible and makes a lot of sense. Thanks for the stripped down version. 🎉 

Added one non-blocking suggestion inline. Besides that it's good to go and can be merged in too. Approving, feel free to take the suggestion if you feel it makes sense.

Good to go ahead on the next PR in your list. 💯 

PS: seems like there is a test failing, can you look into that. Good to merge once that is passing of course!

sdk/txbuild/formation.go
```diff
@@ -35,6 +37,13 @@ func Formation(p FormationParams) (*txnbuild.Transaction, error) {
 		SourceAccount: p.InitiatorEscrow.Address(),
 		Signer:        &txnbuild.Signer{Address: p.InitiatorSigner.Address(), Weight: 1},
```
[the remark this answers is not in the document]
acharb: ya  good question I noticed that too. Looks like bc the protocol is designed that way
https://stellarfoundation.slack.com/archives/C02B04RMK/p1623866928302000

sdk/txbuild/formation.go
```diff
@@ -35,6 +37,13 @@ func Formation(p FormationParams) (*txnbuild.Transaction, error) {
...
 	})
+	if !p.Asset.IsNative() {
+		tp.Operations = append(tp.Operations, &txnbuild.ChangeTrust{
+			Line:          p.Asset,
+			Limit:         p.AssetLimit,
+			SourceAccount: p.InitiatorEscrow.Address(),
+		})
+	}
 	tp.Operations = append(tp.Operations, &txnbuild.EndSponsoringFutureReserves{SourceAccount: p.InitiatorEscrow.Address()})
 	tp.Operations = append(tp.Operations, &txnbuild.BeginSponsoringFutureReserves{SourceAccount: p.ResponderSigner.Address(), SponsoredID: p.ResponderEscrow.Address()})
```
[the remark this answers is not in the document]
acharb: my view is avoiding over-modularizing code at this early stage in the project. Right now it may make sense to group this code into a helper method, but things are changing pretty frequently. Imo a helper method here (called just twice, doesn't really make it that much more readable) right now may end up being more work in the long run trying to fit helper methods to use cases when things change, or then removing the helper methods since they don't make sense anymore

if alright with you I'd like to leave as is. As the project becomes more set in stone is when I think it would make sense to make these changes.

lmk if you disagree though happy to discuss / adjust my coding methodology 😃

sdk/state/close.go
```diff
@@ -116,5 +116,6 @@ func (c *Channel) makeCloseTx(observationPeriodTime time.Duration, observationPe
 		IterationNumber:            c.latestCloseAgreement.IterationNumber,
 		AmountToInitiator:          c.initiatorClaimAmount(),
 		AmountToResponder:          c.responderClaimAmount(),
+		Asset:                      c.latestCloseAgreement.Balance.Asset,
```
acharb, line 119: @nikhilsaraf fyi I actually needed to include this other change so the test doesn't break, sorry for adding late. Because the closing transaction created in state/close.go needs the asset now, we're pulling it from the `latestCloseAgreement`. Similarly how the amount is pulled from the `latestCloseAgreement`

going to merge now but I can make changes in next PR if needed

sdk/txbuild/formation.go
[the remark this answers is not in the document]
nikhilsaraf: Yeah, that makes sense. I'm generally thinking in terms of descriptive logical units but it's equally good to leave as is so I'm fine with that. Thanks for the context!

sdk/state/close.go
nikhilsaraf: thanks for highlighting.

One thought is that the asset should probably live on the Channel rather than pulling from the latestCloseAgreement, but we don't need to worry about that since that's merely a design/interface optimization that we can improve later if we feel the need. I'm ok with leaving as-is. 

Thanks!

sdk/state/open.go
```diff
@@ -26,6 +26,8 @@ func (c *Channel) OpenTxs() (txClose, txDecl, formation *txnbuild.Transaction, e
 		IterationNumber:            1,
 		AmountToInitiator:          0,
 		AmountToResponder:          0,
+		// TODO - change to non native
+		Asset: NativeAsset{},
```
nikhilsaraf, line 30: like the TODOs !

PR #91, opened 2021-06-17
sdk/state: finish adding non-native asset support

acharb:
this is the final changes for supporting non-native assets.

1. Changes the Open functions to take asset type as an input, so that we can create channels with non-native
2. Changes the Update functions to support non-native
3. Add non-native functionality to tests

(note the closing functions are already supporting non-native from previous PRs)

### DIFF
--- a/sdk/state/integration/state_test.go
+++ b/sdk/state/integration/state_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stellar/experimental-payment-channels/sdk/state"
 	"github.com/stellar/go/clients/horizonclient"
 	"github.com/stellar/go/keypair"
+	"github.com/stellar/go/protocols/horizon"
 	"github.com/stellar/go/txnbuild"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -24,7 +25,7 @@ type Participant struct {
 	KP                   *keypair.Full
 	Escrow               *keypair.Full
 	EscrowSequenceNumber int64
-	Contribution         int64
+	Contribution         int64 // The contribution of the asset that will be used for payments
 }
 
 // Setup
@@ -42,6 +43,7 @@ func TestMain(m *testing.M) {
 
 func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
 	asset := txnbuild.NativeAsset{}
+	// native asset has no asset limit
 	assetLimit := ""
 	rootResp, err := client.Root()
 	require.NoError(t, err)
@@ -60,7 +62,7 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
 
 	// Open
 	t.Log("Open...")
-	open, err := initiatorChannel.ProposeOpen()
+	open, err := initiatorChannel.ProposeOpen(state.OpenParams{asset, assetLimit})
 	require.NoError(t, err)
 	for {
 		var fullySignedR bool
@@ -79,7 +81,7 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
 	}
 
 	{
-		ci, di, fi, err := initiatorChannel.OpenTxs()
+		ci, di, fi, err := initiatorChannel.OpenTxs(state.OpenParams{asset, assetLimit})
 		require.NoError(t, err)
 
 		ci, err = ci.AddSignatureDecorated(open.CloseSignatures...)
@@ -283,12 +285,8 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
 }
 
 func TestOpenUpdatesCoordinatedClose(t *testing.T) {
-	// TODO - test nonnative asset
-	asset := txnbuild.NativeAsset{}
-	assetLimit := ""
-	rootResp, err := client.Root()
-	require.NoError(t, err)
-	distributor := keypair.Master(rootResp.NetworkPassphrase).(*keypair.Full)
+	asset, distributor := initAsset(t, client)
+	assetLimit := "5000"
 	initiator, responder := initAccounts(t, client, asset, assetLimit, distributor)
 	initiatorChannel, responderChannel := initChannels(t, client, initiator, responder)
 
@@ -299,7 +297,7 @@ func TestOpenUpdatesCoordinatedClose(t *testing.T) {
 
 	// Open
 	t.Log("Open...")
-	open, err := initiatorChannel.ProposeOpen()
+	open, err := initiatorChannel.ProposeOpen(state.OpenParams{Asset: asset, AssetLimit: assetLimit})
 	require.NoError(t, err)
 	for {
 		var fullySignedR bool
@@ -318,7 +316,7 @@ func TestOpenUpdatesCoordinatedClose(t *testing.T) {
 	}
 
 	{
-		ci, di, fi, err := initiatorChannel.OpenTxs()
+		ci, di, fi, err := initiatorChannel.OpenTxs(state.OpenParams{Asset: asset, AssetLimit: assetLimit})
 		require.NoError(t, err)
 
 		ci, err = ci.AddSignatureDecorated(open.CloseSignatures...)
@@ -375,7 +373,7 @@ func TestOpenUpdatesCoordinatedClose(t *testing.T) {
 		t.Log("Proposal: ", i, paymentLog, amount/1_000_0000)
 
 		// Sender: creates new Payment, sends to other party
-		payment, err := sendingChannel.ProposePayment(state.Amount{Asset: state.NativeAsset{}, Amount: amount})
+		payment, err := sendingChannel.ProposePayment(state.Amount{Asset: asset, Amount: amount})
 		require.NoError(t, err)
 
 		var fullySigned bool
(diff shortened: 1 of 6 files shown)


sdk/state/open.go
```diff
@@ -2,19 +2,29 @@ package state
...
 type Open struct {
 	CloseSignatures       []xdr.DecoratedSignature
 	DeclarationSignatures []xdr.DecoratedSignature
 	FormationSignatures   []xdr.DecoratedSignature
+
+	Asset      Asset
+	AssetLimit string
+}
+
+type OpenParams struct {
```
acharb, line 22: this is the main change of this PR. We're now passing the Asset to `ProposeOpen`, so that we can form the channel with the non-native asset

note: I added an input Params struct `OpenParams`, I'm not tied to the name bc sounds pretty similar to `Open`. But maybe different enough 🤔

sdk/state/open.go
```diff
@@ -47,18 +56,23 @@ func (c *Channel) OpenTxs() (txClose, txDecl, formation *txnbuild.Transaction, e
...
+		AssetLimit:      p.AssetLimit,
 	})
 	return
 }
 
 // ProposeOpen proposes the open of the channel, it is called by the participant
 // initiating the channel.
-func (c *Channel) ProposeOpen() (Open, error) {
+func (c *Channel) ProposeOpen(p OpenParams) (Open, error) {
+	if !p.Asset.IsNative() {
```
acharb, line 68: need to validate that asset limit is a string that can be parsed as an int. Note it's a type string because `txnbuild.ChangeTrust` operation takes `Limit` as a string input

sdk/state/open.go
```diff
@@ -168,7 +184,7 @@ func (c *Channel) ConfirmOpen(m Open) (open Open, fullySigned bool, err error) {
 	fullySigned = true
 	c.latestCloseAgreement = CloseAgreement{
 		IterationNumber:       1,
-		Balance:               Amount{},
+		Balance:               Amount{Asset: m.Asset},
```
acharb, line 187: the balance on the stored `latestClosAgreement` now stores the asset type and doesn't assume native

sdk/state/open.go
```diff
@@ -2,19 +2,29 @@ package state
...
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/stellar/experimental-payment-channels/sdk/txbuild"
 	"github.com/stellar/go/txnbuild"
 	"github.com/stellar/go/xdr"
 )
 
+// TODO - should store on channel like Update and Close proposals?
```
acharb, line 12: added this TODO for another PR. created a ticket for it: https://github.com/stellar/experimental-payment-channels/issues/92

sdk/state/open_test.go
```diff
@@ -0,0 +1,47 @@
+package state
```
acharb, line 1: added this single unit test to test this new OpenParams input

sdk/state/state.go
```diff
@@ -22,7 +22,6 @@ type Amount struct {
 type EscrowAccount struct {
 	Address        *keypair.FromAddress
 	SequenceNumber int64
-	Balances       []Amount
```
acharb, line 25: not being used and we don't need this right now. When we add support for multiple assets we might add back

sdk/state/update.go
```diff
@@ -46,6 +46,10 @@ func (c *Channel) ProposePayment(amount Amount) (Payment, error) {
 	if amount.Amount <= 0 {
 		return Payment{}, errors.New("payment amount must be greater than 0")
 	}
+	if amount.Asset != c.latestCloseAgreement.Balance.Asset {
```
acharb, line 49: need to validate the input asset is the one we're expecting since not always native now

sdk/state/update.go
```diff
@@ -132,14 +136,20 @@ func (c *Channel) ConfirmPayment(p Payment) (payment Payment, fullySigned bool,
...
 
+	// validate payment
 	if p.IterationNumber != c.NextIterationNumber() {
 		return p, fullySigned, errors.New(fmt.Sprintf("invalid payment iteration number, got: %s want: %s",
 			strconv.FormatInt(p.IterationNumber, 10), strconv.FormatInt(c.NextIterationNumber(), 10)))
 	}
 	if !c.latestUnconfirmedPayment.isEmpty() && !c.latestUnconfirmedPayment.isEquivalent(p) {
 		return p, fullySigned, errors.New("a different unconfirmed payment exists")
 	}
+	if p.Amount.Asset != c.latestCloseAgreement.Balance.Asset {
```
acharb, line 147: same validation, can't assume native asset now

sdk/state/update_test.go
```diff
@@ -38,8 +38,12 @@ func TestLastConfirmedPayment(t *testing.T) {
 		RemoteEscrowAccount: localEscrowAccount,
 	})
 
+	// latest close agreement should be set during open steps
```
acharb, line 41: this would normally be set during the `open` tests. Mimicking that for this unit test

sdk/state/integration/state_test.go
```diff
@@ -283,12 +285,8 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
 }
 
 func TestOpenUpdatesCoordinatedClose(t *testing.T) {
-	// TODO - test nonnative asset
-	asset := txnbuild.NativeAsset{}
-	assetLimit := ""
-	rootResp, err := client.Root()
-	require.NoError(t, err)
-	distributor := keypair.Master(rootResp.NetworkPassphrase).(*keypair.Full)
+	asset, distributor := initAsset(t, client)
```
acharb, line 288: now we're creating a new asset instead of testing with the native one

sdk/state/integration/state_test.go
```diff
@@ -454,10 +452,19 @@ func TestOpenUpdatesCoordinatedClose(t *testing.T) {
...
+	assert.EqualValues(t, fmt.Sprintf("%.7f", float64(rBalanceCheck)/float64(1_000_0000)), assetBalance(asset, responderEscrowResponse))
 
 	accountRequest = horizonclient.AccountRequest{AccountID: initiator.Escrow.Address()}
 	initiatorEscrowResponse, err := client.AccountDetail(accountRequest)
 	require.NoError(t, err)
-	assert.EqualValues(t, fmt.Sprintf("%.7f", float64(iBalanceCheck)/float64(1_000_0000)), initiatorEscrowResponse.Balances[0].Balance)
+	assert.EqualValues(t, fmt.Sprintf("%.7f", float64(iBalanceCheck)/float64(1_000_0000)), assetBalance(asset, initiatorEscrowResponse))
+}
+
+func assetBalance(asset txnbuild.Asset, account horizon.Account) string {
```
acharb, line 463: helper function that gets the balance for a specific asset from a `horizon.Account` response

nikhilsaraf: super intuitive to read this function used in the code above 💯 🎉

sdk/state/open.go
nikhilsaraf: we could call it `AssetParams` or `AssetLimit` since it's encapsulating the Asset info like asset type and limit.
OpenParams is a little confusing, but adding a comment on the struct explaining it's intended use case allows us to be a little more flexible in naming it.

sdk/state/open.go
nikhilsaraf: Generally I prefer to keep things strongly typed until the point of translation where possible. Puts more work on the compiler so less context to keep in the programmer's mind while coding.

That would also eliminate us needing to parse as a String here.
But fine the way it is as a string for now so no need to change.

sdk/state/update_test.go
```diff
@@ -38,8 +38,12 @@ func TestLastConfirmedPayment(t *testing.T) {
...
+	sendingChannel.latestCloseAgreement.Balance = Amount{Asset: NativeAsset{}}
+	receiverChannel.latestCloseAgreement.Balance = Amount{Asset: NativeAsset{}}
```
nikhilsaraf, line 43: out of curiosity -- why does the balance have on limit? Does that mean if someone pays the payment channel account with a native asset then that by default will be claimable by the initiator or remote based on who ends up with that escrow account?
and that this issue does not apply to non-native assets because of the trust limit that we set on the trustline?

nikhilsaraf (approved):
looks great, thanks! 🎉 

have the green light to merge.

sdk/state/open.go
acharb: left a comment. I think the params may be expanding to more than just asset later on so left as `OpenParams` for now
https://github.com/stellar/experimental-payment-channels/pull/91/commits/bafd563bb751648bfa3c240ee6415bc1790a820e

sdk/state/update_test.go
acharb: correct the `assetLimit` is just needed for the adding the trustlines to the escrow accounts. But that's a good point that we're not checking if a payment surpasses the escrow account `assetLimit`. Which could cause the closing transaction to fail if surpassing that assetLimit 🤔 

I'll add a ticket for that, because it needs some additional thinking through.

It may tie in with [this ticket](https://github.com/stellar/experimental-payment-channels/issues/92) for storing the open object on channel. So we can use it to check the channel assetLimit. Or maybe we should store on `Balance` as you said

acharb: https://github.com/stellar/experimental-payment-channels/issues/93

PR #96, opened 2021-06-23
sdk/state: fix TestProposePayment_validAsset test

leighmcculloch:
### What

Small changes to the `TestProposePayment_valid_asset` test:

- Rename to `TestProposePayment_validAsset`.
- Change the setup of the invalid asset test so that a valid asset limit is provided.
- Use EqualError instead of separate Error and Equal assertions.

### Why

Camel case for test descriptions is more common after the first underscore that separates the test entity from the description.

The invalid asset test had an invalid asset limit so it was testing the same thing as the invalid asset limit test and not actually testing that invalid assets provided to the function will be caught.

We can use EqualError to check that an error occurred and that its error string matches the given value.

### DIFF
--- a/sdk/state/open_test.go
+++ b/sdk/state/open_test.go
@@ -8,7 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestProposePayment_valid_asset(t *testing.T) {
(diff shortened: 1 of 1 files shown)


acharb (approved):
nice catch lgtm!

PR #98, opened 2021-06-23
sdk/state: fix linter errors

leighmcculloch:
### What
Replace `errors.New(fmt.Sprintf(...))` with `fmt.Errorsf(...)`.

### Why
One of the default Go linters that are configured in VSCode is warning about it.

### DIFF
--- a/sdk/state/update.go
+++ b/sdk/state/update.go
@@ -138,15 +138,15 @@ func (c *Channel) ConfirmPayment(p Payment) (payment Payment, fullySigned bool, 
 
 	// validate payment
 	if p.IterationNumber != c.NextIterationNumber() {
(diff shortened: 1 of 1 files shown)


acharb (approved):
lgtm 👍 

for my own knowledge is the reason your linter is catching this because fmt.Errorf is generally advised over errors.New?

leighmcculloch:
The linter error that is showing up is `S1028` from the `staticcheck` linter: https://staticcheck.io/docs/checks#S1028. It's just a simplification along similar lines of some of the simplifications that go fmt will apply in simplification mode. We use `staticcheck` in the stellar/go repo as part of CI, but we don't have it setup on this repo to run as part of CI.

PR #99, opened 2021-06-23
sdk: add golangci-lint ci run

leighmcculloch:
### What
Add golangci-lint linter to run on PRs.

### Why
To get some of the basic lint checks that we use in other repos, like staticcheck, to reduce warnings that some IDEs like VSCode will display when viewing the code in this repo.

We don't actually use golangci-lint in the stellar/go repo, but it is a popular and simple way to run a variety of linters including staticcheck, and if it goes well here I think I'll propose we replace staticcheck there with it too.

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -8,6 +8,22 @@ on:
 
 jobs:
 
+  lint:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


acharb (approved):
lgtm 👍  looks like it already caught some stuff 😎

PR #101, opened 2021-06-24
sdk/txbuild: remove comment about time bounds

leighmcculloch:
### What
Remove the todo comment saying we need to make the time bounds explicit.

### Why
Channels are long lived and we actually want the closes to be valid for a long time, indefinitely. There's no need to restrict the window within which they are valid. We're using the new relative bounds mechanisms introduced in CAP-21 to limit when they can be executed instead.


acharb (approved):
makes sense 👍 

I also added in [txbuild/declaration.go](https://github.com/stellar/experimental-payment-channels/blob/4adc3cf5ccac03259cc1f84535a91d0c5610e709/sdk/txbuild/declaration.go#L23) if you could remove there as well 😄

PR #102, opened 2021-06-25
sdk/state: use CloseAgreement instead of Payment and CoordinatedClose structs

acharb:
from [this comment](https://github.com/stellar/experimental-payment-channels/pull/83#discussion_r657387696) I removed `CoordinatedClose` and changed `channel.latestUnconfirmedPayment` to `channel.latestUnconfirmedCloseAgreement`. 

In doing so it became apparent we don't need the `Payment` struct, and can just pass around / store a `CloseAgreement` 😄 . So made those changes here

### DIFF
--- a/sdk/state/update.go
+++ b/sdk/state/update.go
@@ -31,7 +31,7 @@ func (ca CloseAgreement) isEquivalent(ca2 CloseAgreement) bool {
 }
 
 func (ca CloseAgreement) isEmpty() bool {
-	return ca.IterationNumber == 0 && ca.Amount == (Amount{}) && len(p.CloseSignatures) == 0 && len(p.DeclarationSignatures) == 0
+	return ca.IterationNumber == 0 && ca.Balance == (Amount{}) && len(ca.CloseSignatures) == 0 && len(ca.DeclarationSignatures) == 0
 }
 
 func (c *Channel) ProposePayment(amount Amount) (CloseAgreement, error) {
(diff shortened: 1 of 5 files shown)


sdk/state/update.go
```diff
@@ -17,37 +17,29 @@ import (
...
-	IterationNumber       int64
-	Balance               Amount
-	CloseSignatures       []xdr.DecoratedSignature
-	DeclarationSignatures []xdr.DecoratedSignature
+func (ca CloseAgreement) isEmpty() bool {
+	return ca.IterationNumber == 0 && ca.Balance == (Amount{}) && len(ca.CloseSignatures) == 0 && len(ca.DeclarationSignatures) == 0
 }
 
-func (c *Channel) ProposePayment(amount Amount) (Payment, error) {
+func (c *Channel) ProposePayment(amount Amount) (CloseAgreement, error) {
```
acharb, line 37: I left the method names to be `ProposePayment / ConfirmPayment`. I think it still makes more sense than `ProposeCloseAgreement / ConfirmCloseAgreement` for the user, but open to suggestions 🤔

leighmcculloch: I think this is the right call, at least with what we know in this moment. I'm suspicious that we might still want to return a Payment type that wraps a CloseAgreement and annotates what the CloseAgreement is, where the CloseAgreement is always determinable from the Payment annotations. We wouldn't store the Payment, we'd still only store the CloseAgreement, but we'd use the Payment in communication.

Happy to run with what you've got here for now though.

sdk/state/update.go
```diff
@@ -97,9 +87,9 @@ func (c *Channel) PaymentTxs(p Payment) (close, decl *txnbuild.Transaction, err
 		ResponderEscrow:            c.responderEscrowAccount().Address,
 		StartSequence:              c.startingSequence,
 		IterationNumber:            c.NextIterationNumber(),
-		AmountToInitiator:          maxInt64(0, newBalance.Amount*-1),
-		AmountToResponder:          maxInt64(0, newBalance.Amount),
-		Asset:                      p.Amount.Asset,
+		AmountToInitiator:          maxInt64(0, ca.Balance.Amount*-1),
```
leighmcculloch, line 90: FYI you can also do the following (e.g. https://play.golang.org/p/Qo8v2N_PVN7):
```suggestion
		AmountToInitiator:          maxInt64(0, -ca.Balance.Amount),
```

leighmcculloch (approved):
🎉

sdk/state/update.go
acharb: til! I actually think it reads easier with the `*-1` though? Unless in golang that's normal. Going to leave for now can change later if needed

PR #108, opened 2021-06-25
sdk/state: make close agreement details comparable

### LINKED ISSUE #105: sdk/state: shift close agreement details that are comparable into their own type
Close agreements are two sets of data, details about what the close agreement is, and signatures with participants authorizing the agreement. This is similar to Stellar transactions which have transaction data and then signatures, and both are grouped together in a transaction envelope.

Separating the values into the types as so makes it easier to do things like compare whether agreements are the same independent of their signatures, and will be easier to understand than having a single type with a soup of all params. This isn't a huge need at the current scale of the data that is stored in the close agreements, but it will be mildly better than what we have and moderately useful if we add more data.

leighmcculloch:
### What
Make the `CloseAgreement` fields that define what the agreement is, comparable.

### Why
In some places we're wanting to compare if two close agreements are the same and we are using a function, isEquivalent, to find out. This works, but there are some advantages to us using simple Go value comparisons for this and we can use them if we group the fields that define the details or term of the agreement together:
- The data structure communicates what parts of the agreement make up the actual agreement vs sigs.
- We can use the agreement details as map keys if we need to.
- It makes copying an agreement independent of signatures simpler.
- This follows the same pattern Stellar transactions use.

Close #105

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -19,7 +19,7 @@ func (c *Channel) CloseTxs() (txDecl *txnbuild.Transaction, txClose *txnbuild.Tr
 	txDecl, err = txbuild.Declaration(txbuild.DeclarationParams{
 		InitiatorEscrow:         c.initiatorEscrowAccount().Address,
 		StartSequence:           c.startingSequence,
-		IterationNumber:         c.latestCloseAgreement.IterationNumber,
+		IterationNumber:         c.latestCloseAgreement.Details.IterationNumber,
 		IterationNumberExecuted: 0,
 	})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


acharb (approved):
nifty trick lgtm 👍 

I'm wondering if we'll ever have to add something non-comparable in the `CloseAgreementDetails` that both parties agree on tho. But probably a cross that bridge when we get there thought

leighmcculloch:
> I'm wondering if we'll ever have to add something non-comparable in the `CloseAgreementDetails` that both parties agree on tho. But probably a cross that bridge when we get there thought

Possibly. At which point we can add a function back but instead of it being a function that returns partial equality it will be a function that returns absolute equality for the details object.

PR #110, opened 2021-06-25
sdk: split unit and integration tests

### LINKED ISSUE #109: sdk: split up unit and integration tests on ci
We've started to see integration tests failing (#100) and it'd be helpful if we ran them separately to unit tests. Over time we should rely on unit tests more so, and integration tests might flake from time to time. While we will work on making them flake less, it'd be helpful to be able to tell at a glance what failures CI has discovered and splitting unit and integration tests to run separately would do that.

leighmcculloch:
### What
Split unit and integration tests on CI, so they run separately. This requires using the `INTEGRATION_TESTS=1` env var when running tests locally to have integration tests run. Also moved the build steps to their own step and also builds tests too.

### Why
Integration tests are slow relative to the unit tests and it'd be helpful if we get immediate feedback on unit test failures, or build failures immediately without having to wait for them all to run.

Close #109


sdk/state/integrationtests/helpers_test.go
```diff
@@ -17,7 +17,7 @@ import (
 
 // functions to be used in the state_test integration tests
 
-func initAccounts(t *testing.T, client horizonclient.ClientInterface, asset txnbuild.Asset, assetLimit string, distributorKP *keypair.Full) (initiator Participant, responder Participant) {
+func initAccounts(t *testing.T, asset txnbuild.Asset, assetLimit string, distributorKP *keypair.Full) (initiator Participant, responder Participant) {
```
leighmcculloch, line 20: This is a bit of a side change but I noticed we were passing around the `client` but using the global `networkPassphrase`, so I stopped passing the client around for consistency.

.github/workflows/sdk.yml
```diff
@@ -39,9 +68,8 @@ jobs:
...
       with:
         go-version: 1
-    - name: Build SDK
-      working-directory: sdk
-      run: go build -v ./...
-    - name: Test SDK
+    - name: Run SDK Integration Tests
       working-directory: sdk
-      run: go test -v ./...
+      run: for p in ./**/integrationtests; do go test -v $p; done
```
leighmcculloch, line 73: `go test` will stream logs when using `-v` and testing a single package, which is why I'm finding the integration tests packages and then running go test on each one individually. Otherwise go test will batch the logs and show them at the end.

.github/workflows/sdk.yml
```diff
@@ -24,7 +24,36 @@ jobs:
...
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1
+    - name: Build SDK
+      working-directory: sdk
+      run: go build ./...
+    - name: Build SDK Tests
+      working-directory: sdk
+      run: go test -exec=echo ./...
```
acharb, line 41: I'm not 100% on how these configs work. Should we do `go test -v` here to see logs of the unit tests?

acharb (approved):
lgtm 👍  1 non-blocking question

.github/workflows/sdk.yml
leighmcculloch: This line with the `-exec=echo` is a hack because there is no way with the Go tooling to tell it to build all the tests in a module without running them. So instead I'm running the tests, but telling it to run the `echo` command instead of the actual tests.

There's an issue open about this limitation here: https://github.com/golang/go/issues/15513.

PR #111, opened 2021-06-25
sdk/state: use authorized/unauthorized language

### LINKED ISSUE #104: sdk/state: replace the term 'unconfirmed' and 'fully signed' with something better
In some places we still refer to close agreements to be unconfirmed and elsewhere we refer to them being fully signed. We should use a single term for all these things, as they're referring to the same concept.

leighmcculloch:
### What
Use authorized and unauthorized language to refer to close agreements that have been signed by all participants or not.

### Why
The code uses two terms `fullySigned` and `unconfirmed` to refer to if a close agreement is respectively authorized or unauthorized. It's unhelpful we use two terms that aren't clearly opposites of one another as the reader must have this tribal knowledge or infer it from logic in the SDK. It would be helpful to use a single term that have simple opposites. Unfortunately `fullySigned` is not a great choice because `notFullySigned` is long and wordy. Unfortunately `unconfirmed` is not a great choice because `confirmed` is an ambiguous term to use given the `ConfirmX` functions in the SDK do not result in a completely confirmed agreement, but in many cases a partially confirmed agreement.

The term authorized is being preferred because it is consistent with language we use when we talk about the state of signatures on a transaction for the Stellar network. On the Stellar network a transaction is authorized if the signers of the account holder(s) have signed the transaction. If a transaction is only partially signed, it is not authorized. We also use this same language when discussing [pre-authorized transactions](https://developers.stellar.org/docs/glossary/multisig/#pre-authorized-transaction), and pre-authorized transaction is authorized ahead of time.

Close #104

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -62,36 +62,36 @@ func (c *Channel) ProposeCoordinatedClose() (CloseAgreement, error) {
 	return c.latestUnauthorizedCloseAgreement, nil
 }
 
-func (c *Channel) ConfirmCoordinatedClose(ca CloseAgreement) (closeAgreement CloseAgreement, fullySigned bool, err error) {
+func (c *Channel) ConfirmCoordinatedClose(ca CloseAgreement) (closeAgreement CloseAgreement, authorized bool, err error) {
 	txCoordinatedClose, err := c.makeCloseTx(0, 0)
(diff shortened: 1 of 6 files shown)


acharb (approved):
lgtm!

PR #113, opened 2021-06-25
sdk/state: keep amounts in int64

leighmcculloch:
### What
Change the asset limit to be an int64 to be consistent with the Amount Balance being stored as an int64.

### Why
For consistency, so all amounts are stored the same way. This also makes it easier to work with the value since we can use simple math operations on it if we need to.

### DIFF
--- a/sdk/state/integrationtests/helpers_test.go
+++ b/sdk/state/integrationtests/helpers_test.go
@@ -3,7 +3,6 @@ package integrationtests
 import (
 	"crypto/rand"
 	"encoding/binary"
-	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


acharb (approved):
lgtm. I left as string to be consistent with the `txnbuild` package, but if you think we might have to alter the value then int64 makes sense

leighmcculloch:
> I left as string to be consistent with the `txnbuild` package, but if you think we might have to alter the value then int64 makes sense

The txnbuild package looks like it was inconsistent too, using int64 for balances, and strings for asset limits. I just fixed that as well in 69737c6. I'd missed that on a first pass.

PR #114, opened 2021-06-25
sdk/state: remove unnecessary use of strconv

leighmcculloch:
### What
Remove use of strconv in fmt strings that can be done with `%d`.

### Why
It's the common way to include numbers in strings, and is simpler.


acharb (approved):
lgtm, good to know

PR #115, opened 2021-06-28
sdk: add delays to retries in integration tests

### LINKED ISSUE #100: sdk/state: integration tests are intermittently failing
The integration tests appear to be failing intermittently.

This is an example of a failure:
https://github.com/stellar/experimental-payment-channels/actions/runs/966218090

leighmcculloch:
### What
Add delays to retries in integration tests.

### Why
The integration tests run in isolation from each other and so can't share data about the state of Horizon. Sometimes they attempt to grab data from Horizon too quickly after a previous test runs and Horizon doesn't know the sequence number of the root account has changed. This causes the next test to fail. We use retries on any root account operations to address this, but there's no sleep on the retries and so the retries run immediately without delay. A short delay should ensure Horizon has a moment to ingest the changes in account state.

Close #100

### DIFF
--- a/sdk/state/integrationtests/helpers_test.go
+++ b/sdk/state/integrationtests/helpers_test.go
@@ -214,6 +214,7 @@ func retry(maxAttempts int, f func() error) (err error) {
 		if err == nil {
 			return
 		}
+		time.Sleep(time.Second)
 	}
 	return err
 }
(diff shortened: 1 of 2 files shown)


acharb (approved):
lgtm

leighmcculloch:
Oops, I merged #116 before this PR. This one was supposed to be merged first. Oh well, this change was merged in #116 now. Closing.

PR #116, opened 2021-06-28
sdk/txbuild: move integration test helpers into another file

leighmcculloch:
### What
Move integration test helpers into a `helpers_test.go`.

### Why
@acharb introduced this pattern in the `state` package and it has worked pretty well in that package.

### Note
This PR contains commits from #115 that should disappear once that PR is merged.


acharb (approved):
lgtm

PR #119, opened 2021-06-28
sdk: run integration tests in ci in a simpler way

leighmcculloch:
### What
Change the CI integration test run to use the `-p` flag to limit the number of package tests that run at same time, instead of using a for loop to manually run the package tests one at a time.

### Why
By default Go runs each packages tests in parallel which causes the logs not to stream which is why we for loop over the packages and run them one by one. Turns out we don't need to bash script it, we can just tell Go to run one package at a time and we'll get the logs streamed.

Dealing with the logs not streaming was a learning for me with Go, so I opened an issue about this on the Go repo (golang/go#46959) because this seems like something that isn't well documented about the go test toolchain.


acharb (approved):
lgtm

PR #123, opened 2021-06-28
sep: use more specific language about when signatures can be exchanged

leighmcculloch:
### What
Use more specific language about exactly when signatures can be exchanged safely in the channel open and update processes, and refocus on language on that exchange rather than on the actual signing.

### Why
@accordeiro pointed out in https://github.com/stellar/experimental-payment-channels/pull/78/files#r660133764 that it isn't when the signing occurs, but when signatures are exchanged, that is the critical aspect of the order of exchange that's required.

This makes the text more accurate.

The SDK we're developing doesn't sign transactions until they're safe to exchange to ensure that we don't ever have a signed agreement exchanged too early, but there could be valid reasons why an implementer wants to sign all components of an agreement upfront even if they don't plan to share all components upfront.


acharb (commented):
makes sense to me. Leaving for Alex to approve since his change though

accordeiro (approved):
Looks great, thank you!

Issue #40, opened 2021-05-21
examples/benchmarks: one-way payments


leighmcculloch:
We need to build a demo that operates over a network to coordinate a payment channel and that demonstrates what sort of naive transactions-per-second could be achieved with the way the protocol is currently designed.

The demo should result in us having some benchmarks and numbers that could be shareable.

It is also expected that in the process of writing this demo we will collect learnings about what could be improved. We are likely to find bottlenecks or inconveniences in the SDK for high-performance applications that we haven't yet discovered in our existing tests and examples. If any of these bottlenecks are roadblocks we may want to look into those issues while working on this demo.

shanzzam:
Assigning to Backlog as there isn't any assigned. We can also make this an Epic as there are multiple tasks working toward this issue as a deliverable.

PR #124, opened 2021-06-28
examples/console: add a lightweight example

### LINKED ISSUE #122: examples/console: add a lightweight demo using the sdk state package
A lightweight demo using the sdk state package outside tests that allows a user to open a channel directly with a network connected party running the same software, and to issue payments, and to receive payments, and to close the channel.

This demo should be really lightweight and won't likely have a long-term use beyond exercising what we have to get a feel for what is missing and to demonstrate what we have right now.

This example demo may become the foundation of #40.

leighmcculloch:
### What
Add a lightweight example using the sdk's state package.

### Why
To learn how easy the API is to use now in lightweight application. To learn how other problems we know need solving get solved if they aren't implemented as part of the SDK. To demonstrate what's working at the moment.

This will also likely be foundation for #40 in next milestone.

For #122


leighmcculloch:
This example code is worthy of merging at this point I think.

The PR has some code in it that we could migrate into the SDK, like the `Submitter` and `ChannelSubmitter` classes, but we can circle back to doing that later. Those classes are helpful utilities.

The example is truly an example and optimizes for simplicity rather than completeness. However, it adequately demonstrates some major use case: 1) happy path, 2) other participant disappears. It doesn't demonstrate contesting or a bad actor.

acharb:
@leighmcculloch still taking a look at this but first thought is could we add a README? trying to run it I'm getting an error, but I'm definitely doing something wrong

ran this command:
```
go run . -account GDA47MAGFUB4XDZMWAVHUAU6KUU7JOSX5PDURBZ2MWTF7WWPWAWE56W3 -signer SALRWV7PJUHBZRW27POF5JN7IKKQHKKWLCSWYUJAKURASR2IJHAU45HB
```
with a fresh local Horizon so makes sense account doesn't exist, but then getting this error
```
account GDA47MAGFUB4XDZMWAVHUAU6KUU7JOSX5PDURBZ2MWTF7WWPWAWE56W3 does not exist, attempting to create using network root key
error: Failed to parse account sequence number: strconv.ParseInt: parsing "": invalid syntax
```

leighmcculloch:
Thanks, I added a readme and pushed a fix to the bug you were seeing.

examples/console/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Follow the [Getting Started](../../Getting%20Started.md) instructions to run a
+stellar-core and horizon with CAP-21 capabilities.
+
+Run the example providing a key and its secret key.
+
+```
+git clone https://github.com/stellar/experimental-payment-channels
+cd examples/console
+go run . -horizon=http://localhost:8000 -account=G... -signer=S...
```
acharb, line 21: it looks like the way it's set up actual keys need to be passed

Just use default random pub/secret keys?
```suggestion
go run . -horizon=http://localhost:8000 -account= GCIFBFIALD3GDBNHAA7QHOYMYFEXVZTRY6ICHWDDB5HELJRGVJZ4HB5V -signer= SDD7AE7ZFFXWC4LNAILC5XQCV7J5BHGUJULB2MAHQ4SQS5SCUYG2YFUW
```

examples/console/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+The application will create the account passed in if needed.
+
+The application will randomly generate and create an escrow account as well.
+
+Type `help` once in to discover commands to use.
```
acharb, line 28: 💭  we could add instructions for quick testing as well, something like

```
to test locally:
in first terminal: 
$ go run . ...
$ listen localhost:8001

in second terminal: 
$ go run . ...
$ connect localhost:8001
```

examples/console/main.go
```diff
@@ -0,0 +1,218 @@
...
+		if len(params) == 0 {
+			continue
+		}
+		switch params[0] {
+		case "help":
+			fmt.Fprintf(os.Stdout, "listen [addr]:<port> - listen for a peer to connect\n")
+			fmt.Fprintf(os.Stdout, "connect <addr>:<port> - connect to a peer\n")
+			fmt.Fprintf(os.Stdout, "open - open a channel with asset\n")
+			fmt.Fprintf(os.Stdout, "deposit <amount> - deposit asset into escrow account\n")
+			fmt.Fprintf(os.Stdout, "pay <amount> - pay amount of asset to peer\n")
```
acharb, line 143: since balances are all in stroops except for here

```suggestion
			fmt.Fprintf(os.Stdout, "deposit <lumen amount> - deposit asset into escrow account\n")
			fmt.Fprintf(os.Stdout, "pay <lumen amount> - pay amount of asset to peer\n")
```

examples/console/agent.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+func (a *Agent) loop() {
+	var err error
+	recv := json.NewDecoder(io.TeeReader(a.Conn(), a.LogWriter))
+	send := json.NewEncoder(io.MultiWriter(a.Conn(), a.LogWriter))
+	for {
+		m := message{}
+		err = recv.Decode(&m)
+		if err != nil {
```
acharb, line 227: ❓ is every log that comes up in the console from the other party? Reason I'm wondering is the logs are pretty hard to mentally parse in the console while going through the steps of open/close/payment.

Could we add a log here that says something like `fmt.Println("receiving message:")`? I believe in here is where it should go

examples/console/agent.go
```diff
@@ -0,0 +1,370 @@
...
+		err = a.handle(m, send)
+		if err != nil {
+			fmt.Fprintf(a.LogWriter, "error handling message: %v\n", err)
+		}
+	}
+}
+
+func (a *Agent) handle(m message, send *json.Encoder) error {
+	if m.Hello != nil {
+		return a.handleHello(*m.Hello, send)
```
acharb, line 240: we could also add in each one of these cases: `handling hello`, `handling open` etc. to help with reading the logs

examples/console/agent.go
```diff
@@ -0,0 +1,370 @@
...
+			if err != nil {
+				return fmt.Errorf("building formation tx: %w", err)
+			}
+			err = a.Submitter.SubmitFeeBumpTx(formationTx)
+			if err != nil {
+				return fmt.Errorf("submitting formation tx: %w", err)
+			}
+		}
+	}
+	if !open.Equal(openIn) {
```
acharb, line 315: 💭  for the `handleOpen` and `handleUpdate` cases were comparing the input object with the updated one to decide if we should return a message

in reality the participant would send the object one last time for the other party to have the final signatures right? Should we do that here?

Might not matter bc we're going to change to the CAP-40 which would change this code anyways, so maybe add a `TODO` to keep track of?

examples/console/agent.go
```diff
@@ -0,0 +1,370 @@
...
+	if err != nil {
+		return fmt.Errorf("confirming payment: %w", err)
+	}
+	if !update.Equal(updateIn) {
+		err = send.Encode(message{Update: &update})
+		if err != nil {
+			return fmt.Errorf("encoding payment to send back: %w", err)
+		}
+	}
+	if authorized {
```
acharb, line 351: when sending a payment both participants get the `payment successfully received` log, even the participant that initially sent it

once again this might not be worth addressing now bc CAP-40 will change this code. So a `TODO` might make sense

examples/console/agent.go
```diff
@@ -0,0 +1,370 @@
...
+		fmt.Fprintf(a.LogWriter, "payment successfully received\n")
+	}
+	return nil
+}
+
+func (a *Agent) handleClose(closeIn state.CloseAgreement, send *json.Encoder) error {
```
acharb, line 357: 💭  after closing we're still able to send payments (though can't close them). Once the close is successful should we quit the console? Or throw an error that the channel is closed

acharb (commented):
this is awesome 🙌  left a couple comments/ 💭  inline. `TODO`s might make sense for some of these since this code will likely change soon with CAP-40

💡 one other idea is could we add a command to print the current participant's channel state? Eg. seeing the balance, iteration number, and if I'm the initiator or not would be a helpful feature

examples/console/main.go
leighmcculloch: Good point. I've changed all the amounts to be consistent. d0d7b94

examples/console/agent.go
leighmcculloch: I think this is a gap in the sdk rather than the example and we should fix it there. I created #186 to address this.

examples/console/agent.go
leighmcculloch: > in reality the participant would send the object one last time for the other party to have the final signatures right? Should we do that here?

You're correct, the participant who signs last needs to send their signatures to the other participant. That's what this code does. It checks if the open received is the same as the open after confirming. If they're the same, no signatures were added to the open, and there's no point sharing it with the other participant since it is identical to the open received. If they're different, signatures must have been added, and the client should send a copy back to the other participant.

Lmk if I'm missing something.

examples/console/agent.go
leighmcculloch: Fixed in 8a8965f.

examples/console/agent.go
leighmcculloch: The logging needs redoing to be able to do this. Right now logs are sent to stderr in between the network connection and the json decoder and if we add the println here it will print out after the message rather than before it. I think we can defer improving this for now and iterate on it in another PR later.

acharb: sgtm 👍

examples/console/agent.go
acharb: 👍 nope I misread looks good

acharb (approved):
lgtm! 🙌

Issue #92, opened 2021-06-17
sdk/state: store the "open" object on channel similar to the update and close channel steps do


acharb:
Right now we're storing the "payment" object `latestUnconfirmedPayment`  during the update steps, and the "closing" object `latestCloseAgreement` and `coordinatedClose` on channel. I'm wondering if we should also be storing the "open" object on the channel as well.

1. Keeps that same pattern
2. Right now when User A proposes an open, and signs txC, they're not storing that signature. So technically User B can return that "open" w/o User A's signature. This is the reason we stored payment and close on the channel so we don't lose signatures

I'm thinking we should do the same here. I'm also wondering if there's a nifty trick we could implement since open, update, close all use this same pattern of storing the message before sending and verifying it when the message is returned

wdyt? @nikhilsaraf @leighmcculloch

leighmcculloch:
I think we should do the same.

> I'm also wondering if there's a nifty trick we could implement since open, update, close all use this same pattern

My opinion is we should not try and find a nifty trick to reduce code duplication for this. There isn't enough code here to benefit from that yet, and it'll be simpler to maintain, delete, and evolve this code if we don't.

PR #125, opened 2021-06-28
sdk/state: store an open agreement on the channel, similar to the `close` and `update` steps

acharb:
during the `open` steps save an `openAgreement` on the channel. This primarily is for ensuring that a channel's signatures aren't lost, where new signatures are appended to the channel's saved `openAgreement` 

this makes the `open` steps consistent with the `update` and `close` steps

closes https://github.com/stellar/experimental-payment-channels/issues/92


sdk/state/integrationtests/state_test.go
```diff
@@ -42,37 +42,46 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
 
 	// Open
 	t.Log("Open...")
+	// I signs txClose
 	open, err := initiatorChannel.ProposeOpen(state.OpenParams{Asset: asset, AssetLimit: assetLimit})
 	require.NoError(t, err)
-	for {
```
acharb, line 47: changed this to do the steps explicitly. I found it difficult to understand how many times each channel calls confirmOpen with the `for loop`

leighmcculloch: Do you think this suggests the approach we are taking where we call a single function to do confirmation is a mistake? Maybe we do need separate functions for each step of we want the caller to know what's happening each each step. Originally the goal was to abstract that so a caller wouldn't need to know.

acharb: 🤔  it does make it clearer if we split it. But I agree that the caller shouldn't need to know which of the transactions are confirmed or not confirmed at a point in time. This change was more for me/us while we're developing to make it clearer.

In practice I think a user calling `ConfirmOpen` until the authorized flag turns true, is a better interface.

Otherwise it could become something like `ConfirmOpen/ReConfirmOpen/ConfirmFormation` or something similarly ugly. Also if we split it then it gets complicated in what if someone calls `ReConfirmOpen` before `ConfirmOpen` or another out of order step etc.

leighmcculloch: Got it. This probably makes the test a bit more predictable too.

leighmcculloch: We don't have to do this in this PR, but now that this is changed to be more explicit we could add some assertions to this test after each call to check the right number of signers are present, or the right signers are present.

sdk/state/integrationtests/state_test.go
```diff
@@ -42,37 +42,46 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
...
+		require.NoError(t, err)
+		require.True(t, authorizedI)
 	}
 
 	{
 		ci, di, fi, err := initiatorChannel.OpenTxs(state.OpenParams{Asset: asset, AssetLimit: assetLimit})
 		require.NoError(t, err)
 
-		ci, err = ci.AddSignatureDecorated(open.CloseSignatures...)
+		ci, err = ci.AddSignatureDecorated(initiatorChannel.OpenAgreement().CloseSignatures...)
```
leighmcculloch, line 76: 💭 I don't think this and the similar changes are necessary since `open` is in scope here and you just need to set it on line 67 above instead of the `_`. You don't have to change it if you prefer to do it this way, but just noticing this change seems unnecessary.
```suggestion
		ci, err = ci.AddSignatureDecorated(open.CloseSignatures...)
```

sdk/state/open.go
```diff
@@ -8,8 +8,7 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-// TODO - should store on channel like Update and Close proposals?
-type Open struct {
+type OpenAgreement struct {
```
leighmcculloch, line 11: 🎉 +1 this is a great change. We can now say that "Agreements" are a concept. I think this is a concept we should name in the SEP too.

sdk/state/open.go
```diff
@@ -100,7 +100,31 @@ func (c *Channel) ProposeOpen(p OpenParams) (Open, error) {
...
+			return
+		}
+		asset := m.Asset
+		assetLimit := m.AssetLimit
+		if c.openAgreement.Asset != nil {
+			asset = c.openAgreement.Asset
+		}
+		if c.openAgreement.AssetLimit != 0 {
+			assetLimit = c.openAgreement.AssetLimit
+		}
```
leighmcculloch, line 117: ❓ It looks like this has the values in `c` override the values in `m`, but throughout the rest of the functions we use the value for `m`, like at line 130. We should never allow us to get to the point where there are different values, so this selection shouldn't be possible, and we should always error if an open is attempted to be confirmed.

leighmcculloch (commented):
One thing I want to understand better before we merge this (❓), and one thought that I defer to you on (💭). Otherwise looks great. I really like the idea to use the term Agreement for all these objects 👏🏻.

sdk/state/open.go
acharb: good catch. I added an `OpenAgreementDetails` similar to the `CloseAgreementDetails`, and then validate the input in `ConfirmOpen` so it's safe to always pull from `m.Details`
https://github.com/stellar/experimental-payment-channels/pull/125/commits/f4dcd5d2b54aad8c753ca70b34b1be11a9bff40a

sdk/state/integrationtests/state_test.go
acharb: I'd like to leave as using the `.OpenAgreement()` when submitting the open transaction to be really sure that the saved open agreement is correct (maybe down the road other code modifies the saved open agreement in an intermediate step). Even though the in scope `open` object should definitely be the same, just to be sure

sdk/state/integrationtests/state_test.go
acharb: added :)
https://github.com/stellar/experimental-payment-channels/pull/125/commits/01c3803648356e523978d5f15d00efc8e316384e

sdk/state/integrationtests/state_test.go
```diff
@@ -42,37 +42,69 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
...
 	// Open
 	t.Log("Open...")
+	// I signs txClose
 	open, err := initiatorChannel.ProposeOpen(state.OpenParams{Asset: asset, AssetLimit: assetLimit})
 	require.NoError(t, err)
-	for {
+	assert.ElementsMatch(t,
+		[]int{1, 0, 0},
+		[]int{len(open.CloseSignatures), len(open.DeclarationSignatures), len(open.FormationSignatures)},
+	)
```
leighmcculloch, line 51: ❗ `ElementsMatch` checks that the elements of the array are the same, but doesn't require them to be in the same order which means this assertion will still pass even if `len(open.CloseSignatures)` is zero and one of the others have a value of one. I believe you need to use `Equal`. Or, to improve the readability of this, simply write three simple assertions.
```suggestion
	assert.Len(t, open.CloseSignatures, 1)
	assert.Len(t, open.DeclarationSignatures, 0)
	assert.Len(t, open.FormationSignatures, 0)
```

sdk/state/integrationtests/state_test.go
```diff
@@ -42,37 +42,69 @@ func TestOpenUpdatesUncoordinatedClose(t *testing.T) {
...
+			[]int{len(open.CloseSignatures), len(open.DeclarationSignatures), len(open.FormationSignatures)},
+		)
 	}
 
 	{
-		ci, di, fi, err := initiatorChannel.OpenTxs(state.OpenParams{Asset: asset, AssetLimit: assetLimit})
+		ci, di, fi, err := initiatorChannel.OpenTxs(state.OpenParams{
+			Asset:      initiatorChannel.OpenAgreement().Details.Asset,
+			AssetLimit: initiatorChannel.OpenAgreement().Details.AssetLimit,
+		})
```
leighmcculloch, line 96: 💭 I think we need to improve how outside code gets at the the txs, but we don't need to solve that here. I opened a separate issue about it: https://github.com/stellar/experimental-payment-channels/issues/127.

sdk/state/open.go
```diff
@@ -64,22 +70,25 @@ func (c *Channel) OpenTxs(p OpenParams) (txClose, txDecl, formation *txnbuild.Tr
...
 	if err != nil {
-		return Open{}, err
+		return OpenAgreement{}, err
 	}
-	open := Open{
+	open := OpenAgreement{
 		CloseSignatures: txClose.Signatures(),
-		Asset:           p.Asset,
-		AssetLimit:      p.AssetLimit,
+		Details: OpenAgreementDetails{
```
leighmcculloch, line 86: The golint-ci linter warning is telling you that you can do `OpenAgreementDetails(p)`.

sdk/state/open.go
```diff
@@ -100,10 +109,32 @@ func (c *Channel) ProposeOpen(p OpenParams) (Open, error) {
...
+				Asset:      m.Details.Asset,
+				AssetLimit: m.Details.AssetLimit,
+			},
+		}
+	}()
+
+	// validate openAgreement details
+	if !c.openAgreement.isEmpty() && m.Details != c.openAgreement.Details {
+		return m, authorized, fmt.Errorf("input open agreement details do not match the saved open agreement details")
+	}
```
leighmcculloch, line 133: 💡 The comment to validate isn't very specific, how about:
```suggestion
	// If the open agreement details don't match the open agreement in progress, error.
	if !c.openAgreement.isEmpty() && m.Details != c.openAgreement.Details {
		return m, authorized, fmt.Errorf("open agreement details do not match the open agreement in progress")
	}
```

leighmcculloch: ❗ Could you move this check above the defer? I think it's a little more defensive to place code that protects the rest of the function ahead of any code, and we don't need the defer to run if this errors.

leighmcculloch (commented):
Looking great, a couple things on the changes below.
❗ important, 💡 suggestion.

sdk/state/integrationtests/state_test.go
acharb: oops thanks, changed
https://github.com/stellar/experimental-payment-channels/pull/125/commits/286d5b91c92ef1af0e40a266d144667660c09134

sdk/state/open.go
acharb: thx
https://github.com/stellar/experimental-payment-channels/pull/125/commits/302589e3a63055ada8fefc8fdfa44a7e088a234f

sdk/state/open.go
acharb: sounds good, I moved the update validation step above the defer too to keep the same pattern
https://github.com/stellar/experimental-payment-channels/pull/125/commits/9a1e0b0da100dd44f31815753c5d143c1f947907

leighmcculloch (approved):

Issue #127, opened 2021-06-29
sdk/state: improve how tx's are accessed outside the package


leighmcculloch:
We need to improve how code outside the channel gets transactions for submission.

We have functions `OpenTxs` and `CloseTxs` that are used by outside the package and internally but they require input params which requires the caller to figure out what the inputs should be from the saved agreements. This makes it harder to get the usage of state.Channel correctly because the caller must select the correct values.

PR #130, opened 2021-06-29
sdk/state: move the tx funcs so they are all in the same place in each file

leighmcculloch:
### What
Move the functions that return transactions into the same place in each file.

### Why
The open, update, close Go code files organize most of their logic in the same way which makes looking at them side by side rather convenient. I'm currently working on #127 and noticed that the functions that return transactions are in different places, and this makes them look more similar.


acharb (approved):
lgtm

PR #131, opened 2021-06-30
sdk/state: consolidate the building of close txs

leighmcculloch:
### What
Consolidate the building of close transactions so that there is only one function and one place that occurs.

### Why
Building the close transactions happens in three places currently, in `CloseTxs`, `PaymentTxs`, and in `ProposePayment`. Each location collect the input data in different ways. There should be one owner responsible for close tx building. Ideally we are always building close txs during the update or final steps the exact same way.

For #127, although does not resolve it completely.

### Merging
This PR should be merged after #130.


acharb (approved):
lgtm, I like the simplification 💯

Issue #80, opened 2021-06-04
sdk/state: update payment/close to use separated disbursements for regulated assets


leighmcculloch:
Add support for disbursements that are separated from the closing transaction, as described in #78.

Will require changes to all steps that build or sign a closing transaction. Will require modifications to the txbuild package too.

This change is blocked on #78 being merged.

PR #132, opened 2021-06-30
sdk/state: remove CoordinatedCloseTxs and build observation into closing agreements

leighmcculloch:
### What
Remove `CoordinatedCloseTxs` and build the observation parameters into the `ClosingAgreement`.

### Why
Coordinated close is the happy path and so we should make it the default way to close, with closing without the additional signatures being the fallback. There's a few changes we need to make for this to happen, and this is the first, that there aren't separate functions, like `CoordinatedCloseTxs` for getting the coordinated closing transaction. The user should be able to just get the closing transactions, via `CloseTxs`, and if the agreement is a coordinated close agreement, it should just work.

There was another reason we were thinking of moving observation parameters into the closing agreement, but I don't remember what they were.

Related to #127 

### Merging

This should be merged after #131 and contains commits from that PR until it is merged.

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -15,6 +15,23 @@ import (
 // 3. Responder calls ConfirmCoordinatedClose
 // 4. Initiator calls ConfirmCoordinatedClose
 
+// makeCloseTx returns a close transaction with observation values.
+func (c *Channel) makeCloseTx(observationPeriodTime time.Duration, observationPeriodLedgerGap int64) (*txnbuild.Transaction, error) {
+	return txbuild.Close(txbuild.CloseParams{
+		ObservationPeriodTime:      observationPeriodTime,
+		ObservationPeriodLedgerGap: observationPeriodLedgerGap,
+		InitiatorSigner:            c.initiatorSigner(),
+		ResponderSigner:            c.responderSigner(),
+		InitiatorEscrow:            c.initiatorEscrowAccount().Address,
+		ResponderEscrow:            c.responderEscrowAccount().Address,
+		StartSequence:              c.startingSequence,
+		IterationNumber:            c.latestAuthorizedCloseAgreement.Details.IterationNumber,
+		AmountToInitiator:          c.initiatorBalanceAmount(),
+		AmountToResponder:          c.responderBalanceAmount(),
+		Asset:                      c.latestAuthorizedCloseAgreement.Details.Balance.Asset,
+	})
+}
+
 func (c *Channel) CloseTxs() (txDecl *txnbuild.Transaction, txClose *txnbuild.Transaction, err error) {
 	txDecl, err = txbuild.Declaration(txbuild.DeclarationParams{
 		InitiatorEscrow:         c.initiatorEscrowAccount().Address,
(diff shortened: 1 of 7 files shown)


sdk/state/close.go
```diff
@@ -63,21 +57,15 @@ func amountToResponder(balance int64) int64 {
...
 // ProposeCoordinatedClose proposes a close transaction to be submitted immediately.
 // This should be used when participants are in agreement on the final txClose parameters, but would
 // like to submit earlier than the original observation time.
 func (c *Channel) ProposeCoordinatedClose() (CloseAgreement, error) {
 	d := c.latestAuthorizedCloseAgreement.Details
+	d.ObservationPeriodTime = 0
+	d.ObservationPeriodLedgerGap = 0
 
-	txCoordinatedClose, err := c.makeCloseTx(d, 0, 0)
+	_, txCoordinatedClose, err := c.CloseTxs(d)
```
leighmcculloch, line 68: It probably seems a bit non-performant to build all transactions here instead of only the close transaction, but that won't matter soon when we implement #80, and even if we weren't doing that, a coordinated close only happens once in a channels lifetime and so the extra work here isn't really that important.

sdk/state/update.go
```diff
@@ -16,8 +17,10 @@ import (
 
 // CloseAgreementDetails contains the details that the participants agree on.
 type CloseAgreementDetails struct {
-	IterationNumber int64
-	Balance         Amount
+	IterationNumber            int64
+	Balance                    Amount
+	ObservationPeriodTime      time.Duration
```
acharb, line 22: ❓ it feels weird to me that we have the observation periods on both the `CloseAgreementDetails` and the `Channel`. Eg. this creates the chance that `channel.latestAuthorizedCloseAgreement.observationPeriodTime` is different than `channel.observationPeriodTime`. Or when trying to find the current channel `O` period do you look at the channel or the latest close agreement?

when proposing and confirming a coordinated close both parties should know to use zero for the observation periods, and instead of needing to validate the observation periods match the validation is done when verifying the signatures.

what if we made the `CloseTxs` func:
`func (c *Channel) CloseTxs(d CloseAgreementDetails, observationPeriod time.Duration, observationPeriodLedgerGap int64)` ? Or something similar

sdk/state/update.go
```diff
@@ -74,6 +79,9 @@ func (c *Channel) ConfirmPayment(ca CloseAgreement) (closeAgreement CloseAgreeme
 	if ca.Details.IterationNumber != c.NextIterationNumber() {
 		return ca, authorized, fmt.Errorf("invalid payment iteration number, got: %d want: %d", ca.Details.IterationNumber, c.NextIterationNumber())
 	}
+	if ca.Details.ObservationPeriodTime != c.observationPeriodTime || ca.Details.ObservationPeriodLedgerGap != c.observationPeriodLedgerGap {
+		return ca, authorized, fmt.Errorf("invalid payment observation period")
```
acharb, line 83: 💡  to make error more specific
```suggestion
		return ca, authorized, fmt.Errorf("invalid payment observation period: different than channel state")
```

acharb (commented):
overall sgtm for cleaning up the close transaction methods. Had a ❓   about the `O` period though, and a 💡

sdk/state/update.go
leighmcculloch: > what if we made the `CloseTxs` func:
> `func (c *Channel) CloseTxs(d CloseAgreementDetails, observationPeriod time.Duration, observationPeriodLedgerGap int64)`

I don't think we should have the observation periods on CloseTxs as separate parameters because the values are part of the agreement and so the calling code should expect all the data it needs for the agreement to either be stored inside the Channel or on the agreement object. It'd be more odd to expect calling code to pass in the observation period.

> it feels weird to me that we have the observation periods on both the `CloseAgreementDetails` and the `Channel`

We could remove the observation period from the `Channel` so it only lives inside the agreements, in the same way that the balances live only inside agreements. I think that'd work. I just made that change in bc1c96c. Wdyt?

acharb: >We could remove the observation period from the Channel so it only lives inside the agreements, in the same way that the balances live only inside agreements. I think that'd work. I just made that change in bc1c96c. Wdyt?

I like it a lot 🙌

acharb (approved):
lgtm!

sdk/state/update.go
leighmcculloch: Full disclosure, part of me worries that this introduces some complexity that is in some ways unnecessary, even though it is useful, and so there's a trade-off here. But let's see how it goes, because having it in the open is better than having it in the Channel. Thanks for leaving this comment 😄.

PR #134, opened 2021-06-30
sdk/state: fix coordinated close agreement declaration signatures

leighmcculloch:
### What
When constructing the coordinated close agreement include declaration signatures of the original close agreement.

### Why
The coordinated `CloseAgreement` doesn't contain declaration signatures. It has a slice for them but it is empty. The coordinated close process replaces the current close agreement, and so for users that choose to coordinate before closing, they will lose signatures for the declaration transaction when the new agreement is promoted to authorized.


acharb (approved):
good point, lgtm

PR #135, opened 2021-06-30
sdk/state: fix coordinated close for confirming participant

leighmcculloch:
### What
Build the final authorized close agreement for a confirmed coordinated close off the last authorized rather than last unauthorized.

### Why
The confirmer can't execute the coordinated close because it has the signatures from the unauthorized transaction added in, but the unauthorized agreement might be a payment that was abandoned and for the confirmer it definitely isn't the close that is being amended. For the confirmer of the coordinated close the agreement should match their authorized, we check that at the beginning, the only thing that is different is the close signatures so we can base it off the rest and then adopt the signatures if they are valid. 

This is a little odd in that the proposer sets it as their unauthorized, but that doesn't really matter because they also have the same agreement, except for the close signatures, set in their authorized so nothing is lost.


sdk/state/close.go
```diff
@@ -122,9 +124,8 @@ func (c *Channel) ConfirmCoordinatedClose(ca CloseAgreement) (closeAgreement Clo
 	authorized = true
 	c.latestAuthorizedCloseAgreement = CloseAgreement{
 		Details:               ca.Details,
-		CloseSignatures:       appendNewSignatures(c.latestUnauthorizedCloseAgreement.CloseSignatures, ca.CloseSignatures),
-		DeclarationSignatures: c.latestUnauthorizedCloseAgreement.DeclarationSignatures,
+		CloseSignatures:       ca.CloseSignatures,
+		DeclarationSignatures: c.latestAuthorizedCloseAgreement.DeclarationSignatures,
 	}
-	c.latestUnauthorizedCloseAgreement = CloseAgreement{}
```
leighmcculloch, line 128: I don't know why I removed this, it should stay, or rather there's little benefit to removing it but it is cleaner for the proposers side.

acharb (approved):
lgtm. Is this also a problem with the `ConfirmPayment` method? A past payment is abandoned and the latestUnauthorizedCloseAgreement signatures aren't cleared

leighmcculloch:
> Is this also a problem with the `ConfirmPayment` method? A past payment is abandoned and the latestUnauthorizedCloseAgreement signatures aren't cleared

If that's the case, yeah. An abandoned payment is also more complex, we _must_ skip over that iteration number. We can't reuse an iteration number so there's probably a bigger problem there.

PR #136, opened 2021-06-30
sdk/state: remove Coordinated term from func names

leighmcculloch:
### What
Remove the term `Coordinated` from function names and code.

### Why
The default happy path is to do a coordinated close, or rather, it should just be part of the standard close process which involves declaring, trying to coordinate the immediate close, and falling back to using the existing close transactions. While we call this coordinated close in the SEP, I don't think we get much value of using the term in the SDK because in practice there will be one close process not two.

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -71,7 +71,7 @@ func (c *Channel) ProposeClose() (CloseAgreement, error) {
 
 	_, txClose, err := c.CloseTxs(d)
 	if err != nil {
-		return CloseAgreement{}, fmt.Errorf("making coordianted close transactions: %w", err)
+		return CloseAgreement{}, fmt.Errorf("making close transactions: %w", err)
 	}
 	txClose, err = txClose.Sign(c.networkPassphrase, c.localSigner)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


sdk/state/close.go
```diff
@@ -57,32 +60,37 @@ func amountToResponder(balance int64) int64 {
...
+// than the original observation time.
+func (c *Channel) ProposeClose() (CloseAgreement, error) {
 	d := c.latestAuthorizedCloseAgreement.Details
 	d.ObservationPeriodTime = 0
 	d.ObservationPeriodLedgerGap = 0
 
-	_, txCoordinatedClose, err := c.CloseTxs(d)
+	_, txClose, err := c.CloseTxs(d)
 	if err != nil {
 		return CloseAgreement{}, fmt.Errorf("making coordianted close transactions: %w", err)
```
acharb, line 74: the only reason I found this is bc I misspelled coordinated when doing global search again 😅  
```suggestion
		return CloseAgreement{}, fmt.Errorf("making close transactions: %w", err)
```

acharb (approved):

PR #138, opened 2021-06-30
sdk/state: reject payments from remote if they change balance in their favor

### LINKED ISSUE #112: sdk/state: ConfirmPayment should error if other participant requests reverse payment
The `Channel` is intended to only support payments, and not invoicing or requesting assets from the other participant. Therefore, it shouldn't be possible for participant A to ask participant B for funds by proposing a negative payment, or by proposing a `CloseAgreement` that increases the agreement in their own favor. Payments should always be a push of assets to the other participant that the other participant can always accept. This simplifies validation for the `Channel` object as the `ConfirmPayment` function can blindly accept any payment that doesn't leave its participant in a worse of state.

It may be possible in the future that we want to support exchanges which necessarily will require a participant to validate the close agreement themselves to decide if what they will receive is acceptable, but that's out of scope of what we're building right now.

This is worth doing as it will make the `Channel` safer to use.

leighmcculloch:
### What
Change the `ConfirmPayment` function so it rejects payments from the remote if the remote is proposing that the local pay the remote.

### Why
This is a payment channel where either party can send money to the other party, and it isn't designed fro requesting/invoicing the other party for funds. There's several reasons for this. It is safer and easier to write confirm functionality that enforces this rather than require that the user always check the balance is an okay change. The channel is designed to always move forward and in disagreement it becomes unclear if the channel should close or stay open and skip a generation. Mostly this is for simplicity for the major use case.

Close #112

### DIFF
--- a/sdk/state/update.go
+++ b/sdk/state/update.go
@@ -127,12 +127,16 @@ func (c *Channel) ConfirmPayment(ca CloseAgreement) (closeAgreement CloseAgreeme
 		return ca, authorized, fmt.Errorf("verifying close signed by remote: not signed by remote")
 	}
 
-	// If local has not signed close, sign.
+	// If local has not signed close, check that the payment is in locals favor, then sign.
 	signed, err = c.verifySigned(txClose, ca.CloseSignatures, c.localSigner)
 	if err != nil {
 		return ca, authorized, fmt.Errorf("verifying close signed by local: %w", err)
 	}
 	if !signed {
+		if (c.initiator && ca.Details.Balance.Amount > c.latestAuthorizedCloseAgreement.Details.Balance.Amount) ||
+			(!c.initiator && ca.Details.Balance.Amount < c.latestAuthorizedCloseAgreement.Details.Balance.Amount) {
+			return ca, authorized, fmt.Errorf("close agreement from remote changes the balance in their favor")
+		}
 		txClose, err = txClose.Sign(c.networkPassphrase, c.localSigner)
 		if err != nil {
 			return ca, authorized, fmt.Errorf("signing close with local: %w", err)
--- a/sdk/state/update.go
+++ b/sdk/state/update.go
@@ -127,7 +127,7 @@ func (c *Channel) ConfirmPayment(ca CloseAgreement) (closeAgreement CloseAgreeme
 		return ca, authorized, fmt.Errorf("verifying close signed by remote: not signed by remote")
 	}
 
-	// If local has not signed close, check that the payment is not from local to remote, then sign.
+	// If local has not signed close, check that the payment is not to the proposer, then sign.
 	signed, err = c.verifySigned(txClose, ca.CloseSignatures, c.localSigner)
 	if err != nil {
 		return ca, authorized, fmt.Errorf("verifying close signed by local: %w", err)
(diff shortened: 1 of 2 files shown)


sdk/state/update.go
```diff
@@ -127,12 +127,16 @@ func (c *Channel) ConfirmPayment(ca CloseAgreement) (closeAgreement CloseAgreeme
...
-	// If local has not signed close, sign.
+	// If local has not signed close, check that the payment is in locals favor, then sign.
 	signed, err = c.verifySigned(txClose, ca.CloseSignatures, c.localSigner)
 	if err != nil {
 		return ca, authorized, fmt.Errorf("verifying close signed by local: %w", err)
 	}
 	if !signed {
+		if (c.initiator && ca.Details.Balance.Amount > c.latestAuthorizedCloseAgreement.Details.Balance.Amount) ||
+			(!c.initiator && ca.Details.Balance.Amount < c.latestAuthorizedCloseAgreement.Details.Balance.Amount) {
+			return ca, authorized, fmt.Errorf("close agreement from remote changes the balance in their favor")
```
acharb, line 138: `changes the balance in their favor` sounds a bit ambiguous to me (could paying money to someone be in my favor?). What about something like:
```suggestion
			return ca, authorized, fmt.Errorf("close agreement is a payment to the proposer")
```

sdk/state/update_test.go
```diff
@@ -85,6 +85,108 @@ func TestChannel_ConfirmPayment_rejectsDifferentObservationPeriod(t *testing.T)
...
+	})
+
+	// A close agreement from the remote participant should be rejected if the
+	// payment changes the balance in the favor of the remote.
+	channel.latestAuthorizedCloseAgreement = CloseAgreement{
+		Details: CloseAgreementDetails{
+			IterationNumber: 1,
+			Balance: Amount{
+				Asset:  NativeAsset{},
+				Amount: 100, // Local (responder) owes remote (initiator) 100.
```
acharb, line 168: technically the positive amount means initiator owes responder. Can we make these values negative? So instead: originally `-100` then new close agreement `-110`

acharb (approved):
sgtm. Had 2 non-critical comments

sdk/state/update.go
leighmcculloch: I like it 💯

sdk/state/update_test.go
leighmcculloch: Ah, good catch. The comment is wrong. It doesn't matter if it is negative or positive, what matters is the diff between this and the new agreement a few lines down has to shift the balance so the local is losing out.
```suggestion
				Amount: 100, // Remote (initiator) owes local (responder) 100.
```

sdk/state/update_test.go
```diff
@@ -85,6 +85,108 @@ func TestChannel_ConfirmPayment_rejectsDifferentObservationPeriod(t *testing.T)
...
+			},
+		},
+	}
+	ca := CloseAgreementDetails{
+		IterationNumber: 2,
+		Balance: Amount{
+			Asset:  NativeAsset{},
+			Amount: 90, // Local (responder) owes remote (initiator) 110, payment of 10 from ❌ local to remote.
```
leighmcculloch, line 176:
```suggestion
			Amount: 90, // Remote (initiator) owes local (responder) 90, payment of 10 from ❌ local to remote.
```

sdk/state/update.go
```diff
@@ -127,12 +127,16 @@ func (c *Channel) ConfirmPayment(ca CloseAgreement) (closeAgreement CloseAgreeme
 		return ca, authorized, fmt.Errorf("verifying close signed by remote: not signed by remote")
 	}
 
-	// If local has not signed close, sign.
+	// If local has not signed close, check that the payment is in locals favor, then sign.
```
leighmcculloch, line 130:
```suggestion
	// If local has not signed close, check that the payment is not from local to remote, then sign.
```

PR #140, opened 2021-07-01
sdk: enable formatting checks

leighmcculloch:
### What
Enable formatting checks in CI with `golangci-lint`.

### Why
I noticed when editing code that there's some code that isn't formatted with common Go formatting tools `go fmt`/`goimports`/`gofumpt`. `golangci-lint` supports format checking and a lot of linters, but it only has a small subset turned on by default and that doesn't include any format checks.

In terms of why `gofumpt`: `gofmt` does formatting and simplification. `goimports` is a superset of `gofmt` without simplification, but keeps import directives up-to-date and is conventionally almost always preferred over `gofmt` now for everything except simplification. `gofumpt` is a superset of `goimports` and is also common in some IDEs like VSCode and makes code more the same, so more code looks the same.

Like the other linters these checks can be run locally by simply running `golangci-lint run`, and formatting can be updated with `golangci-lint run --fix`.

This is somewhat unrelated to this PR, but there's one linter already enabled, govet, that we should enable an optional feature for, the shadow checks. I think we should enable it because stellar/go uses it and if we move this code into stellar/go it will be easier if our code is passing the same linters and is at an equal or higher standard. I have commented it out because it doesn't pass at the moment and I don't want to spend the time fixing that atm, but will at some point when I have a moment.

### DIFF
--- a/sdk/.golangci.yml
+++ b/sdk/.golangci.yml
@@ -1,9 +1,12 @@
 linters:
   # Enable these linters in addition to the default linters that golangci-lint starts with.
   enable:
+    - gofmt
     - gofumpt
 
 linters-settings:
   govet:
     # TODO: Enable shadow checks, because stellar/go has them enabled, but they're a bit painful for us to enable now.
     # check-shadowing: true
+  gofmt:
+    simplify: true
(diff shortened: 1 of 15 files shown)


acharb (approved):
lgtm! thx for the explanation in the description

PR #147, opened 2021-07-03
sep: add notes about trustline limit

### LINKED ISSUE #145: sep: simplify the trustline limit
Let's remove the trustline limit as a configuration and always assume the limit will be the maximum int64.

@stanford-scs made a good point today that trustline limits are inconvenient and has not proven to be particularly useful on chain either. 

If someone opens a payment channel with a small asset limit there are some edge cases that can get tricky if someone deposits into the asset limit and breaks the closing transaction. It's not particularly troublesome, but since the asset limit isn't proving useful to people let's hardcode it and leave it out until someone identifies a need for it.

cc @acharb

leighmcculloch:
### What
Restrict the trustline limit that non-native assets use to the max.

### Why
To simplify the protocol. See the change for more details.

Close #145

### DIFF
--- a/specifications/sep-payment-channel-transactions.md
+++ b/specifications/sep-payment-channel-transactions.md
@@ -216,7 +216,8 @@ multisig accounts. F has source account E, and sequence number set to s.
   trustline between escrow account setup and formation causing the presigned
   closing transaction to become invalid.
 
-  The `CHANGE_TRUST` operations configure the trustlines with the maximum limit, which is the maximum value of an `int64`, `0x7FFFFFFFFFFFFFFF`.
+  The `CHANGE_TRUST` operations configure the trustlines with the maximum limit,
+  which is the maximum value of an `int64`, `0x7FFFFFFFFFFFFFFF`.
 
 - C_i, see [Update](#Update) process.
 
@@ -634,9 +635,12 @@ Implementations that allow lower asset limits may produce closing transactions
 that could fail if the final state makes a payment that would exceed the
 destination account's trustline limit.
 
-Implementations that are intended for use with assets that have excessive supply may also produce closing transactions that could fail if trustline limits would be exceeded because of excessive deposits.
+Implementations that are intended for use with assets that have excessive supply
+may also produce closing transactions that could fail if trustline limits would
+be exceeded because of excessive deposits.
 
-In both cases a party who is not a participant can deposit an amount into the escrow accounts to cause the closing transaction's payment to fail.
+In both cases a party who is not a participant can deposit an amount into the
+escrow accounts to cause the closing transaction's payment to fail.
 
 ### Clawback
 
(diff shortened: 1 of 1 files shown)


specifications/sep-payment-channel-transactions.md
```diff
@@ -565,6 +568,24 @@ process if its payment operation is dependent on amounts frozen.
 There is nothing participants can do to prevent this, other than using only auth
 immutable assets.
 
+### Trustline Limits
+
+Trustlines on the escrow accounts are defined as always having a maximum asset
```
nikhilsaraf, line 573: If my understanding is correct, escrow accounts will always have max limits, and closing transactions are made as claimable balances while retaining access of the escrow account to the initiator.

In none of these situations are we making a payment that can fail because of a lack of available limit, correct?

specifications/sep-payment-channel-transactions.md
```diff
@@ -565,6 +568,24 @@ process if its payment operation is dependent on amounts frozen.
...
+limit. This restriction makes the behavior of the closing transaction as
+predictable as possible and simplifies implementations that are designed for
+operating on common assets that do not have excessive supply.
+
+Implementations that allow lower asset limits may produce closing transactions
```
nikhilsaraf, line 578: my confusion arises from this part. The closing transaction will be paid as a claimable balance so therefore should not be able to fail as a result of the destination account's trustline limit, correct?

nikhilsaraf (approved):
some questions inline, but approving to unblock if these don't require a change.

specifications/sep-payment-channel-transactions.md
leighmcculloch: Escrow accounts always have max limits, but there are no claimable balances in use. Payments occur directly between the two escrow accounts. Payments shouldn't fail unless they involve a trust line that has been auth revoked, or some of the asset in the sending trust line has clawed back.

Setting the asset limit to max should for the most part mean implementations don't need to check if the asset limit would be exceeded, assuming the accounts are not interacting with assets that have excessive supply on chain.

specifications/sep-payment-channel-transactions.md
leighmcculloch: Ah, so we don't use claimable balances. The closing transaction contains these operations:
https://github.com/stellar/experimental-payment-channels/blob/f421bace8d0917a9fd7e7ae9935f492a158bda62/specifications/sep-payment-channel-transactions.md#L264-L270

leighmcculloch: We didn't use claimable balances because they require sponsorship. Creation of new ledger entries can fail if sponsorship fails which makes the protocol more complicated. It requires us to store reserves in accounts. It resulted in us needing two escrow accounts so that the accounts are completely separated at close, because otherwise reserves and sponsorship mixing could result in the accounts being bound together after close. Creating a claimable balance is a new ledger entry which is why reserves/sponsorship are involved.

So instead we we use simple payments between two escrow accounts. 🎉

specifications/sep-payment-channel-transactions.md
nikhilsaraf: ok that makes sense -- yeah I forgot we changed it to a two-escrow system

PR #148, opened 2021-07-06
sdk/state: support single asset only

### LINKED ISSUE #72: sdk/state: support single assets/balances
The channel supports a single asset, but should be updated to support multiple assets and balances.

leighmcculloch:
### What
Change the SDK back to supporting a single asset only.

### Why
We resolved to focus on single asset payment channels because someone made a good point that if we aren't supporting exchange/swaps, then there isn't huge value in having multiple assets in the one channel vs multiple single-asset channels.

Incorporates the changes proposed by SEP changes #146 and #147.

Close #72

### DIFF
--- a/sdk/state/integrationtests/helpers_test.go
+++ b/sdk/state/integrationtests/helpers_test.go
@@ -21,10 +21,9 @@ import (
 type AssetParam struct {
 	Asset       txnbuild.Asset
 	Distributor *keypair.Full
-	AssetLimit  int64
 }
 
-func initAccounts(t *testing.T, assets []AssetParam) (initiator Participant, responder Participant) {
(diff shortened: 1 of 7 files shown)


acharb (commented):
lgtm! Had 1 question. 

It looks like we're keeping the `Trustline` struct, but if we follow the sep about always using the max assetLimit (https://github.com/stellar/experimental-payment-channels/pull/147), then it would be simpler to just use `txnbuild.Asset` instead? Or could that change be in a separate PR

leighmcculloch:
> It looks like we're keeping the `Trustline` struct

It looks like I missed removing a ton more functionality 😄. I just pushed 1601e42 that should remove the remainder of the multi asset support from the formation and open code.

acharb (approved):
so much simpler 🙌  lgtm!

PR #149, opened 2021-07-07
sdk/state: replace txnbuild.Asset with an internal simpler asset type

leighmcculloch:
### What
Replace txnbuild.Asset in the state package with an internal simpler asset type, a string that follows [SEP-11 Txrep]'s format.

### Why
While experimenting with the state package (#122) I found the interface type txnbuild.Asset inconvenient. Interfaces can't be used to deserialize. While working around that issue I realized we don't really need an interface for what we're doing, we only need a simple value and it is more convenient to be able to specify a concrete value in the simplest form and to convert it to what the txnbuild package needs at the point it needs it.

I chose a string because we have a specification for assets as string in [SEP-11 Txrep] which has been used in Horizon for new endpoints that reference assets.

[SEP-11 Txrep]: https://stellar.org/protocol/sep-11

### DIFF
--- a/sdk/state/asset.go
+++ b/sdk/state/asset.go
@@ -1,0 +1,34 @@
+package state
+
+import (
+	"strings"
+
+	"github.com/stellar/go/txnbuild"
+)
+
+type Asset string
+
+const NativeAsset = Asset("native")
+
+func (a Asset) Native() bool {
+	return a.Asset().IsNative()
+}
+
+func (a Asset) Code() string {
+	return a.Asset().GetCode()
+}
+
+func (a Asset) Issuer() string {
+	return a.Asset().GetIssuer()
+}
+
+func (a Asset) Asset() txnbuild.Asset {
+	parts := strings.SplitN(string(a), ":", 2)
+	if len(parts) == 1 {
+		return txnbuild.NativeAsset{}
+	}
+	return txnbuild.CreditAsset{
+		Code:   parts[0],
+		Issuer: parts[1],
+	}
+}
--- a/sdk/state/asset.go
+++ b/sdk/state/asset.go
@@ -10,7 +10,7 @@ type Asset string
 
 const NativeAsset = Asset("native")
 
-func (a Asset) Native() bool {
+func (a Asset) IsNative() bool {
 	return a.Asset().IsNative()
 }
 
--- a/sdk/state/asset_test.go
+++ b/sdk/state/asset_test.go
@@ -1,0 +1,34 @@
+package state_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stellar/experimental-payment-channels/sdk/state"
+	"github.com/stellar/go/txnbuild"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAsset(t *testing.T) {
+	testCases := []struct {
+		Asset             state.Asset
+		WantTxnbuildAsset txnbuild.Asset
+		WantNative        bool
+		WantCode          string
+		WantIssuer        string
+	}{
+		{state.Asset(""), txnbuild.NativeAsset{}, true, "", ""},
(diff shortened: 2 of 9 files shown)


sdk/state/asset_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/stellar/experimental-payment-channels/sdk/state"
+	"github.com/stellar/go/txnbuild"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAsset(t *testing.T) {
+	testCases := []struct {
+		Asset             state.Asset
+		WantTxnbuildAsset txnbuild.Asset
+		WantNative        bool
```
acharb, line 16: nit: is `WantNative` needed? If `WantTxnbuildAsset` is already checking that

sdk/state/asset_test.go
```diff
@@ -0,0 +1,34 @@
...
+		WantCode          string
+		WantIssuer        string
+	}{
+		{state.Asset(""), txnbuild.NativeAsset{}, true, "", ""},
+		{state.Asset("native"), txnbuild.NativeAsset{}, true, "", ""},
+		{state.NativeAsset, txnbuild.NativeAsset{}, true, "", ""},
+		{state.Asset(":"), txnbuild.CreditAsset{}, false, "", ""},
+		{state.Asset("ABCD:GABCD"), txnbuild.CreditAsset{Code: "ABCD", Issuer: "GABCD"}, false, "ABCD", "GABCD"},
```
acharb, line 24: can we add a test case for something like `ABCD:GABCD:AB`?

sdk/state/asset.go
```diff
@@ -0,0 +1,34 @@
...
+	"strings"
+
+	"github.com/stellar/go/txnbuild"
+)
+
+type Asset string
+
+const NativeAsset = Asset("native")
+
+func (a Asset) Native() bool {
```
acharb, line 13: 💭 to me `Native()` just based off the name feels like it's converting an asset to a Native representation. I remember you saying `Is<foo>` is generally not used in go, but I wonder in this case if it makes sense to just mimic the `IsNative()` method since that's what's currently on txnbuild.Asset as well?

I don't feel terribly strongly about this though if you think `Native()` is correct

acharb (approved):
lgtm! I left a couple thoughts, but nothing blocking

sdk/state/asset_test.go
leighmcculloch: `WantNative` is the value we expect to get from the `.Native()` func, `WantTxnbuildAsset` is the value we expect to get from the `.Asset()` func.

sdk/state/asset_test.go
leighmcculloch: 👍🏻 f140df3

sdk/state/asset.go
leighmcculloch: +1. I think IsX would be clearer here. When I said that IsX isn't common in Go, I was wrong about that. It is true that GetX/SetX naming isn't common and is explicitly called out in the effective Go docs, but IsX is common, even in the Go stdlib it is present. Updated in 5456314.

acharb: > When I said that IsX isn't common in Go, I was wrong about that. It is true that GetX/SetX naming isn't common

oh I may have mixed those up 😅. good to know though thx 👍

leighmcculloch: No you had heard me correctly, I was just wrong. I think at some point I started to believe no-Get/Set meant no-Is, but that's just not the case, which is good because the IsX reads better here.

sdk/state/asset_test.go
leighmcculloch: I added back a test I deleted and added an additional test in 9933490.

acharb (approved):
lgtm! 👍

PR #153, opened 2021-07-08
sdk/state: make formation expiry deterministic

### LINKED ISSUE #150: sdk/state: make formation deterministic by making its time bounds deterministic
The formation transaction sets its time bounds relative to the now time, which requires both participants to have synchronized clocks and for the clients to generate the transaction at the exact moment. This is unlikely to happen. This behavior also makes it impossible to regenerate the formation transaction later.

We should either:
 - Set the time bounds to be infinite.
 - Set the time bounds max to a fixed value as part of the open agreement. (Preferred)

The declaration and close transactions use the first idea above, but that is because they explicitly should be able to be executed at any time. For the formation the second is preferred because the transaction is intended for executing soon.

leighmcculloch:
### What
Make the formation transaction that gets built deterministic by changing the time bounds to be based on time.Now to be based on a fixed time the proposer chooses.

### Why
The formation transaction sets its time bounds relative to the now time, which requires both participants to have synchronized clocks at a resolution of 1 second and for the clients to generate the transaction at the exact moment. This is unlikely to happen. This behavior also makes it impossible to regenerate the formation transaction later.

I discovered this problem while working on #122. Two participants running separately were producing different open agreement transactions because the time bound maximum was set to different values in the formation transaction.

Close #150

### DIFF
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -20,7 +20,7 @@ func (d OpenAgreementDetails) Equal(d2 OpenAgreementDetails) bool {
 	return d.ObservationPeriodTime == d2.ObservationPeriodTime &&
 		d.ObservationPeriodLedgerGap == d2.ObservationPeriodLedgerGap &&
 		d.Asset == d2.Asset &&
-		d.ExpiresAt == d2.ExpiresAt
+		d.ExpiresAt.Equal(d2.ExpiresAt)
 }
 
 type OpenAgreement struct {
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -126,6 +126,11 @@ func (c *Channel) ConfirmOpen(m OpenAgreement) (open OpenAgreement, authorized b
 		return m, authorized, fmt.Errorf("input open agreement details do not match the saved open agreement details")
 	}
 
+	// If the expiry of the agreement is past the max expiry the channel will accept, error.
+	if m.Details.ExpiresAt.After(time.Now().Add(c.maxOpenExpiry)) {
+		return m, authorized, fmt.Errorf("input open agreement expire too far into the future")
+	}
+
 	// at the end of this method, if no error, then save a new channel openAgreement. Use the
 	// channel's saved open agreement details if present, to prevent other party from changing.
 	defer func() {
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -13,10 +13,14 @@ type OpenAgreementDetails struct {
 	ObservationPeriodTime      time.Duration
 	ObservationPeriodLedgerGap int64
 	Asset                      Asset
+	ExpiresAt                  time.Time
 }
 
 func (d OpenAgreementDetails) Equal(d2 OpenAgreementDetails) bool {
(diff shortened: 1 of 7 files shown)


sdk/state/open.go
```diff
@@ -120,6 +126,11 @@ func (c *Channel) ConfirmOpen(m OpenAgreement) (open OpenAgreement, authorized b
 		return m, authorized, fmt.Errorf("input open agreement details do not match the saved open agreement details")
 	}
 
+	// If the expiry of the agreement is past the max expiry the channel will accept, error.
```
acharb, line 129: ❓ If my timezone is your timezone minus 1 hour. And you send me an open agreement, this check would fail right?

🤔 if so I'm wondering how we could resolve. Keep everything in perspective of utc, `time.Now().UTC()`?

acharb (commented):
overall sgtm. Just left two ❓ related to timezones

I'm also wondering if it's a good idea to create an integration test with two parties in different timezones. Or that could be done as part of the demo

sdk/txbuild/formation.go
```diff
@@ -24,7 +26,7 @@ func Formation(p FormationParams) (*txnbuild.Transaction, error) {
 			Sequence:  p.StartSequence,
 		},
 		BaseFee:    0,
-		Timebounds: txnbuild.NewTimeout(300),
+		Timebounds: txnbuild.NewTimebounds(0, p.ExpiresAt.Unix()),
```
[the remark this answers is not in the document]
leighmcculloch: You're right, it should be UTC.
```suggestion
		Timebounds: txnbuild.NewTimebounds(0, p.ExpiresAt.UTC().Unix()),
```

sdk/state/open.go
leighmcculloch: I don't think we need the UTC here because the time object will contain the local timezone and convert it appropriately in the comparison. However we do need to use it before converting to unix at the other spot because unix time doesn't encode the timezone.

acharb: 🤔 right because the time object contains the local timezone wouldn't the example I gave above stand? eg.
```
Initiator timezone: UTC - 0. time.Now() = 3:00
Responder timezone: UTC - 2. time.Now() = 1:00

Initiator proposes open with expiry at 3:05

Responder calls ConfirmOpen, and 3:05 > 1:05, so this would error.
```
Right? (I might be thinking about it wrong)

leighmcculloch: You are correct except those times all contain time zones so they all compare fine.

```
Initiator timezone: UTC - 0. time.Now() = 3:00 UTC-0
Responder timezone: UTC - 2. time.Now() = 1:00 UTC-2

Initiator proposes open with expiry at 3:05 UTC-0 (1:05 UTC-2)

Responder calls ConfirmOpen, and 3:05 UTC-0 == 1:05 UTC-2, so this would be fine.
```

acharb: yup you're right, took me a second to wrap my head around 👍

acharb (approved):
lgtm!

PR #154, opened 2021-07-08
sep, sdk/state: fix sponsorship operations during open

### LINKED ISSUE #151: sdk/state: make signer sponsorship consistent
It looks like the SEP and our implementation of who sponsors which signers has diverged.

In the code the owner of the escrow account sponsors all signers on that escrow account.

https://github.com/stellar/experimental-payment-channels/blob/7abcd861801d65c80fbb96a8aa83fc2bebc8f2a2/sdk/txbuild/formation.go#L30-L62

In the SEP the owner of the signer sponsors their own signers on all escrow accounts.

https://github.com/stellar/experimental-payment-channels/blob/e3fd9b8c2cdce5f8e1951bab8c3501999487e85a/specifications/sep-payment-channel-transactions.md#L497-L507

I don't think it really matters which approach we take at this point. Longer term it would be ideal to do what the SEP says because it allows an account with multiple signers to sponsor its own excessive signer capabilities. In the short term we are only implementing single signer so that is mute, and it makes it cheaper to not have to switch sponsors which is an additional four operations during formation. That fee saving isn't significant though, so I think we should change the SDK to match the SEP, then the SEP supports a future use and the SDK is consistent with it.

leighmcculloch:
### What
Make the participants sponsor only their own signers.

### Why

The SEP states that the participants are only supposed to sponsor their own signers, but then the open process has them sponsoring each other signers. This has little impact on single signer setups, but cleaning this up makes it a bit more pure on ownership of the respective ledger entries.

Close #151

### DIFF
--- a/sdk/txbuild/formation.go
+++ b/sdk/txbuild/formation.go
@@ -26,6 +26,8 @@ func Formation(p FormationParams) (*txnbuild.Transaction, error) {
 		BaseFee:    0,
 		Timebounds: txnbuild.NewTimeout(300),
 	}
+
+	// I sponsoring ledger entries on EI
 	tp.Operations = append(tp.Operations, &txnbuild.BeginSponsoringFutureReserves{SourceAccount: p.InitiatorSigner.Address(), SponsoredID: p.InitiatorEscrow.Address()})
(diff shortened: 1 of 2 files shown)


acharb (approved):
lgtm

PR #155, opened 2021-07-09
sdk/state: validate balances and handle deposits and balance changes

### LINKED ISSUE #48: sdk/state: validate balances and handle deposits and balance changes
The state machine doesn't need to support facilitating the actual deposit, because that will happen on network using regular simple transaction. The state machine needs to provide a way to update the balances within after a deposit has occurred.

The state machine does not currently validate that the close agreement payment that will be made will be satisfied by the balance of the paying participant's balance.

We need to add a check in the confirm payment, that checks if the known balance of the escrow account is sufficient.

This change needs to be aware of deposits and balance changes that may occur in the escrow account. A deposit will change the balances of the escrow account. We probably do not want to check the escrow accounts balance on every payment. Ideally we only check Horizon for the escrow balance when we receive a payment that won't be sufficient.

leighmcculloch:
### What

When receiving a proposed new close agreement/payment, error with `ErrUnderfunded` if the close agreement over commits the payer. Support updating the known balances of escrow accounts so the payment can be confirmed if on network the balance has been increased.

### Why

The state machine does not validate that a proposed close agreement will be satisfied by the balance of the paying participant's balance. This is a problem because the payer could over commit, breaking the channel. The receiver in a payment shouldn't accept a close agreement if it can't be satisfied by the current balances in the escrow accounts.

It's possible for participants to deposit more of an asset into their escrow account, so it should also be possible 
to update the known balances inside the channel when that happens. Deposits can happen at anytime without a participants involvement so the flow must support a way to move forward if a participant proposes a payment that based on stale knowledge is invalid, but would be valid with updated knowledge. The ConfirmPayment function returns an `ErrUnderfunded` error so that a caller can respond to that specific situation by checking the balance of the payers escrow account on network, updating it locally, and trying again.

Close #48

### DIFF
--- a/sdk/go.sum
+++ b/sdk/go.sum
@@ -154,9 +154,14 @@ github.com/stellar/throttled v2.2.3-0.20190823235211-89d75816f59d+incompatible/g
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/objx v0.1.1 h1:2vfRuCMp5sSVIDSqO8oNnWJq7mPa6KVP3iPIwFBuy8A=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.3.0 h1:NGXK3lHquSN08v5vWalVI/L8XU9hdzE/G6xsrze47As=
+github.com/stretchr/objx v0.3.0/go.mod h1:qt09Ya8vawLte6SNmTgCsAVtYtaKzEcn8ATUoHMkEqE=
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
+github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/stretchr/testify v1.5.1 h1:nOGnQDM7FYENwehXlg/kFVnos3rEvtKTjRvOWSzb6H4=
 github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
+github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
+github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/tyler-smith/go-bip39 v0.0.0-20180618194314-52158e4697b8/go.mod h1:sJ5fKU0s6JVwZjjcUEX2zFOnvq0ASQ2K9Zr6cf67kNs=
 github.com/valyala/bytebufferpool v1.0.0 h1:GqA5TC/0021Y/b9FG4Oi9Mr3q7XYx6KllzawFIhcdPw=
 github.com/valyala/bytebufferpool v1.0.0/go.mod h1:6bBcMArwyJ5K/AmCkWv1jt77kVWyCJ6HpOuEn7z0Csc=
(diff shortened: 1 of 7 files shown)


sdk/go.sum
```diff
@@ -154,9 +154,14 @@ github.com/stellar/throttled v2.2.3-0.20190823235211-89d75816f59d+incompatible/g
...
 github.com/stretchr/objx v0.1.1 h1:2vfRuCMp5sSVIDSqO8oNnWJq7mPa6KVP3iPIwFBuy8A=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.3.0 h1:NGXK3lHquSN08v5vWalVI/L8XU9hdzE/G6xsrze47As=
+github.com/stretchr/objx v0.3.0/go.mod h1:qt09Ya8vawLte6SNmTgCsAVtYtaKzEcn8ATUoHMkEqE=
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
+github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/stretchr/testify v1.5.1 h1:nOGnQDM7FYENwehXlg/kFVnos3rEvtKTjRvOWSzb6H4=
 github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
+github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
+github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
```
leighmcculloch, line 164: I updated the testify module to get access to the `assert.ErrorIs` func.

sdk/state/update.go
```diff
@@ -137,6 +139,9 @@ func (c *Channel) ConfirmPayment(ca CloseAgreement) (closeAgreement CloseAgreeme
 			(!c.initiator && ca.Details.Balance.Amount < c.latestAuthorizedCloseAgreement.Details.Balance.Amount) {
 			return ca, authorized, fmt.Errorf("close agreement is a payment to the proposer")
 		}
+		if c.amountToLocal(ca.Details.Balance.Amount) > c.remoteEscrowAccount.Balance {
+			return ca, authorized, fmt.Errorf("close agreement over commits: %w", ErrUnderfunded)
+		}
 		txClose, err = txClose.Sign(c.networkPassphrase, c.localSigner)
```
acharb, line 137: 💭  thoughts on also adding a check when proposing a payment. To check that the amount I'm sending is not greater than what I have available. More of a self-check for accidentally proposing too much.

Bc if I do propose too much, and the other party decides to accept it, then the channel is in trouble

sdk/state/state.go
```diff
@@ -86,6 +87,14 @@ func (c *Channel) LatestCloseAgreement() CloseAgreement {
 	return c.latestAuthorizedCloseAgreement
 }
 
+func (c *Channel) UpdateLocalEscrowAccountBalance(balance int64) {
```
acharb, line 85: 💭 `UpdateRemoteEscrowAccountBalance` is currently only used in the tests, so this is dependent on the orchestration code to do the updating? I'm wondering
1. should we be updating the balance w/in the state machine? Eg. in ConfirmPayment
2. if not 1. Then should this be living on the state machine since? ie should we make it a method in the test, or if you foresee the orchestration code using then makes sense to me keep it here

acharb (commented):
left a couple thoughts but overall lgtm

sdk/state/state.go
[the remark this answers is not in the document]
leighmcculloch: No not really, at least not yet. Removed.

sdk/state/update.go
leighmcculloch: Good idea. Let's address that separate to this. I created an issue for it: https://github.com/stellar/experimental-payment-channels/issues/157.

sdk/state/state.go
leighmcculloch: I don't think the state machine should update itself because that requires a call out to Horizon. The application or code that uses the state machine needs to take care of knowing how to call out to Horizon (or tracking balances some other way).

> should we make it a method in the test

I don't think I understand this suggestion. Would the test update the internal value of the escrow account balance? I think we need some way for the caller to update the balance since the state machine can't do it on its own.

acharb: > I don't think the state machine should update itself because that requires a call out to Horizon
true agreed, for a moment I forgot about the reason for this PR 😆 

>Would the test update the internal value of the escrow account balance?
true, disregard

sgtm to leave as is thx 👍

acharb (approved):
lgtm!

PR #156, opened 2021-07-09
sdk/state: remove assets from places only needed for multi asset support

leighmcculloch:
### What
Remove assets from `CloseAgreements` and some other places.

### Why

In #72 we rewound support for multiple assets and moved forward with single asset support only. `CloseAgreements` only need to include asset information if the channel is multi-asset. This simplifies several things. 

I realized this could be simplified when I worked on #155. In that change I needed to add a balance to `EscrowAccount`. It was going to be significantly harder to follow the existing pattern and include asset information in that object, so I didn't. I think it is worth bringing the same simplicity to close agreements and everywhere else.

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -37,9 +37,9 @@ func (c *Channel) CloseTxs(d CloseAgreementDetails) (txDecl *txnbuild.Transactio
 		ResponderEscrow:            c.responderEscrowAccount().Address,
 		StartSequence:              c.startingSequence,
 		IterationNumber:            d.IterationNumber,
-		AmountToInitiator:          amountToInitiator(d.Balance.Amount),
-		AmountToResponder:          amountToResponder(d.Balance.Amount),
-		Asset:                      d.Balance.Asset.Asset(),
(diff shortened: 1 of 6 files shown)


sdk/state/close.go
```diff
@@ -37,9 +37,9 @@ func (c *Channel) CloseTxs(d CloseAgreementDetails) (txDecl *txnbuild.Transactio
 		ResponderEscrow:            c.responderEscrowAccount().Address,
 		StartSequence:              c.startingSequence,
 		IterationNumber:            d.IterationNumber,
-		AmountToInitiator:          amountToInitiator(d.Balance.Amount),
-		AmountToResponder:          amountToResponder(d.Balance.Amount),
-		Asset:                      d.Balance.Asset.Asset(),
+		AmountToInitiator:          amountToInitiator(d.Balance),
+		AmountToResponder:          amountToResponder(d.Balance),
+		Asset:                      c.openAgreement.Details.Asset.Asset(),
```
leighmcculloch, line 42: The channel relies on the asset information stored in the open agreement now.

acharb (approved):
sweet lgtm 👍

PR #158, opened 2021-07-12
sdk/state: prevent proposing payment that over commits

### LINKED ISSUE #157: sdk/state: prevent proposing payment that over commits payer
Similar to logic we introduced in #155 we should make it difficult for the proposer of a payment to propose a payment that the other participant is going to reject immediately. Examples of this would be a payment that over commits a payer, and would require more in the payers balance than they have.

Also, it's not in the payers best interest to pay too much because doing so would break the channel which would likely also be bad for them.

Thanks @acharb for identifying this issue: https://github.com/stellar/experimental-payment-channels/pull/155#discussion_r667233215.

leighmcculloch:
### What

Prevent a participant from proposing a payment that would exceed the balance they supplied to the payment channel.

### Why

@acharb pointed out (https://github.com/stellar/experimental-payment-channels/pull/155#discussion_r667233215) that allowing a participant to propose a payment too large could break the payment channel for both participants because the payment in the closing transaction would fail with insufficient funds. This isn't good for either participant. It also makes little sense to allow a payment to be proposed if a sensible other participant will reject it, which they should.

Close #157

### DIFF
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -82,6 +82,10 @@ func (c *Channel) LatestCloseAgreement() CloseAgreement {
 	return c.latestAuthorizedCloseAgreement
 }
 
+func (c *Channel) UpdateLocalEscrowAccountBalance(balance int64) {
+	c.localEscrowAccount.Balance = balance
+}
+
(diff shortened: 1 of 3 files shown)


acharb (approved):
lgtm!

PR #162, opened 2021-07-12
sep: change reasoning behind setup checks

leighmcculloch:
### What
Change the reasoning behind setup checks to more generally focus on participants needing to verify the other participants escrow account state after formation.

### Why
The statements made in the SEP already about these checks are focused on clawback state, but that is one of many states that could change. There are much worse things that can change, such as additional signers could be added by the other participant.


specifications/sep-payment-channel-transactions.md
```diff
@@ -612,8 +614,9 @@ check if either participant has clawback enabled. Checking the state after
 formation is critical because there is no way for participants to guarantee
 trustline state until after formation has completed because the state can change
 prior to formation. For this reason participants should perform their initial
-deposit after formation, unless they trust the asset issuer not to clawback from
-payment channel escrow accounts, or unless the asset is auth immutable.
+deposit after formation, unless they have no requirement for the asset not to be
+clawback enabled, or they trust the asset issuer not to clawback from payment
```
acharb, line 618: i think. If not, the double negative is hard to digest
```suggestion
deposit after formation, unless they have no requirement for the asset to be
clawback enabled, or they trust the asset issuer not to clawback from payment
```

specifications/sep-payment-channel-transactions.md
```diff
@@ -170,9 +170,11 @@ without any payment.
...
 8. I or R submit F.
 
-Participants should defer deposits of initial contributions till after formation
-for channels that will hold trustlines to issuers that are not auth immutable,
-and could be clawback enabled. See [Security](#Security).
+Participants may prefer to make deposits of initial contributions after
+formation if the channel will hold a trustline to an issuer that is not auth
+immutable, the participants' trustlines are not clawback enabled, and the
+participants do not want the trustlines of the escrow accounts to be clawback
+enabled. See [Security](#Security).
```
acharb, line 178: I'm having trouble wrapping my head around this section, I think bc I don't fully understand clawback.

❓ why does it matter if the initial contribution is made after formation if the asset is a clawback enabled one?

acharb (commented):
overall lgtm, had a question for my own understanding I'd like to understand before approving

specifications/sep-payment-channel-transactions.md
acharb: I think I might be understanding my confusion. Should this be instead:
```
the participants' trustlines are clawback enabled, or the
participants do not want the trustlines of the escrow accounts to be clawback
enabled.
```

leighmcculloch: Yeah this is not particularly easy to read, but I think it is accurate. Let's see if we can make it clearer.

Clawback can be enabled on an issuer account at anytime, unless the issuer account is "auth immutable". When a trust line is created the trust line inherits the clawback enabled flag from the issuer account. This means that when the escrow account adds its trust line is the point in time that we can have confidence about whether the escrow account's trust line will be clawback enabled or not.

If the initial contribution is made before formation and the other participants escrow account's trust line is created during formation you don't know if the other escrow account's trust line will be clawback enabled. All of this is probably fine for the average user, but if a user specifically doesn't want for their assets, and their payment channel to be subject to the clawback enabled capability, they'd ideally want to see the formation take place and ensure the other escrow account is not clawback enabled before depositing any asset to the channel.

Writing all that out I it does occur to me now that this is a bit of a special case and esoteric. It applies to most assets – because most assets are not auth immutable – but it only applies to users who are particularly sensitive to clawback enabled.

I think maybe we should reduce this paragraph to simply saying that in some cases users wanting to avoid their asset and channel being subject to the clawback enabled capability should read the security section. In that section we can discuss it in more detail.

Wdyt? If you think that sounds good I can take a stab at updating this PR.

acharb: gotchya thank you for the explanation! I understand now. 

Ya I agree with you about putting the bulk of the clawback explanation in the security section. I think the extra space to explain this case fully is helpful (at least it took me a bit to wrap my head around what this paragraph was saying). Suggesting below for the rework:

```
Participants may prefer to make deposits of initial contributions after
formation to ensure clawback trustlines are set correctly. See [Security](#Security).
```
and then in the security section it's explained in more detail.

specifications/sep-payment-channel-transactions.md
acharb: changing the suggestion based off your explanation above. But if this is not the specific intent your trying to say then nvrm 

```suggestion
deposit after formation, unless they are okay with the asset potentially being clawback enabled, or they trust the asset issuer not to clawback from payment
```

leighmcculloch:
@acharb I adopted one of your changes, and then in the process of applying the other I realized we need to keep this requirement but for different reasons. I've rewritten the text in the security section to discuss the broader reason.

specifications/sep-payment-channel-transactions.md
```diff
@@ -171,9 +171,10 @@ without any payment.
 7. I and R sign and exchange the formation transaction F.
 8. I or R submit F.
 
-Participants should defer deposits of initial contributions till after formation
-for channels that will hold trustlines to issuers that are not auth immutable,
-and could be clawback enabled. See [Security](#Security).
+Participants should check the state of the other participant's escrow account
+after the formation transaction is submitted to ensure the state is as
+expected.Participants may wish to reduce their exposure to griefing by making
```
acharb, line 176:
```suggestion
expected. Participants may wish to reduce their exposure to griefing by making
```

acharb (approved):
lgtm! 🙌

PR #167, opened 2021-07-14
sep: make clear that C_i has zero or one payment ops

leighmcculloch:
### What
Make it clear that the closing transaction C_i may have zero or one payment operations.

### Why
The document says the operation may be omitted, but we should make it clearer that is the case by saying there can be zero or one, rather than saying there will always be one.


acharb (approved):
🚀

PR #168, opened 2021-07-14
all: adding gitignore file

acharb:
**WHAT**
adding .gitignore file

**WHY**
my .DS_Store keeps popping up :) adding gitignore to ignore this or other future files we don't want checked in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store


.gitignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
leighmcculloch, line 1: We can add this, but I would recommend setting up a global gitignore file instead. Something like [this](https://github.com/leighmcculloch/dotfiles/blob/master/files/gitignore_global) is relatively common.

You can have your global gitignore file live anywhere on your system, I have mine at `~/.gitignore_global`, and you tell git where to find it with:

```
git config --global core.excludesfile '~/.gitconfig'
```

That command will edit your `~/.gitconfig` file adding a line like this:
https://github.com/leighmcculloch/dotfiles/blob/93aaaf4/files/gitconfig#L4

leighmcculloch (approved):

.gitignore
acharb: oooh TIL! 💯  will set this up

Issue #142, opened 2021-07-01
sdk/state: error when agreements contain extra signatures


leighmcculloch:
Extra signatures in agreements can cause a transaction to fail when submitted, so we should at least discard extra signatures, or error if a participant tries to send them to us.

PR #169, opened 2021-07-14
sdk/state: handle extra signatures in transactions

acharb:
**WHAT**
check that open and close agreements have only 2 sigs max for declarationSigs, closeSigs, and formationSigs (open only)

also adds a unit test file `sdk/state/close_test.go` for `close.go` logic

**WHY**
we're not currently checking an agreement has too many sigs, which will cause the transaction to fail when submitted. Here we're checking that there's no more than 2 signatures.

After the CAP-40 change we're no longer supporting participants having multiple sigs, so each participant is allowed 1 signature (see comment [here](https://github.com/stellar/experimental-payment-channels/pull/176#discussion_r672389953)), for a total of 2 sigs allowed per agreement

closes https://github.com/stellar/experimental-payment-channels/issues/142

### DIFF
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -103,6 +103,26 @@ func (c *Channel) ProposeOpen(p OpenParams) (OpenAgreement, error) {
 	return open, nil
 }
 
+func (c *Channel) validateOpen(m OpenAgreement) error {
+	// If the open agreement details don't match the open agreement in progress, error.
+	if !c.openAgreement.isEmpty() && !m.Details.Equal(c.openAgreement.Details) {
+		return fmt.Errorf("input open agreement details do not match the saved open agreement details")
+	}
+
+	// If the expiry of the agreement is past the max expiry the channel will accept, error.
+	if m.Details.ExpiresAt.After(time.Now().Add(c.maxOpenExpiry)) {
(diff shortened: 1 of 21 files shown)


acharb:
@leighmcculloch PTAL, I moved the check to the end of the Confirm functions as you suggested [here](https://github.com/stellar/experimental-payment-channels/pull/183#discussion_r672542533)

leighmcculloch (approved):
👏🏻

sdk/state/open.go
```diff
@@ -161,6 +161,16 @@ func (c *Channel) ConfirmOpen(m OpenAgreement) (open OpenAgreement, authorized b
 		if err != nil {
 			return
 		}
+
+		// If an agreement ever surpasses 2 signatures per tx, error.
+		if len(m.DeclarationSignatures) > 2 || len(m.CloseSignatures) > 2 || len(m.FormationSignatures) > 2 {
+			authorized = false
```
acharb, line 167: arguably we could return `authorized` as true here since it is fully signed 🤔 . Could come in use for whatever code calls this?

leighmcculloch: Nah, I think you should stick with false. It might be authorized, but if you submitted it with extra signatures, the tx would fail.

PR #177, opened 2021-07-16
sdk/state: add some helpful funcs

### LINKED ISSUE #175: sdk/state: add Equal funcs to OpenAgreement and CloseAgreement
One thing I discovered working on the #122 is that it would be helpful to have `Equal` functions on the `OpenAgreement` and `CloseAgreement` types. I suspect we also will need them for #163, but we also need it for users of the SDK.

leighmcculloch:
### What
Add some functions that I've found helpful:
 - Equal functions on agreements.
 - Functions to get the escrow account objects from the channel.

### Why
It's helpful for the user to be able to compare agreements to know if an agreement they have is the same as a new agreement.

It's helpful to be able to access the full state of the escrow account as known by the channel, and not just the balance that we expose now.

I discovered this while working on #122.

Close #175 

### Known Limitations
I shortcutted writing the equals methods by using the cmp.Equal package. We should really write a proper equals package that doesn't use reflection under the surface, but this should be sufficient for now.

### DIFF
--- a/sdk/state/open_test.go
+++ b/sdk/state/open_test.go
@@ -1,13 +1,130 @@
 package state
 
 import (
+	"strconv"
 	"testing"
 	"time"
 
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/network"
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func TestOpenAgreement_Equal(t *testing.T) {
+	testCases := []struct {
+		oa1       OpenAgreement
+		oa2       OpenAgreement
+		wantEqual bool
+	}{
+		{OpenAgreement{}, OpenAgreement{}, true},
+		{
+			OpenAgreement{
+				Details: OpenAgreementDetails{
+					ObservationPeriodTime:      time.Minute,
+					ObservationPeriodLedgerGap: 2,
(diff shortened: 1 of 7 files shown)


sdk/state/open_test.go
```diff
@@ -1,14 +1,131 @@
...
 	"time"
 
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/network"
+	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func TestOpenAgreement_Equal(t *testing.T) {
```
acharb, line 15: 💭 `cmp` compares signatures as well? In which case can we add a test case of 2 different signatures to

acharb (approved):
lgtm, left 1 💭

sdk/state/open_test.go
leighmcculloch: Yup. The third last and last test case test that actually 😄.

acharb: ah I meant two existing sigs that are different, not existent and non-existent

I don't see that case being tested 🤔

leighmcculloch: Oh gotcha. That's what the last test case was supposed to be but looks like it has a bug. I'll fix it in another PR.

PR #178, opened 2021-07-16
sdk/state: export initiator from Channel

leighmcculloch:
### What
Export the initiator variable from the Channel.

### Why
The caller already knows about it since they choose who the initiator is during config, and it is helpful in picking who should take the first action in some processes like the open process. We might remove this later on to hide the notion of initiator, but for now having it is useful.

Something I discovered working on #122.


acharb (approved):
lgtm

PR #182, opened 2021-07-19
sdk/state: fix test

### LINKED ISSUE #175: sdk/state: add Equal funcs to OpenAgreement and CloseAgreement

leighmcculloch:
### What
Add a test case for open and close agreement equality functions that check if the signatures are different the function returns false.

### Why
@acharb pointed out that the tests were not actually testing that case because of a mistake in the last test cases. 

https://github.com/stellar/experimental-payment-channels/pull/177/files#r671552880

Close #175


acharb (approved):
lgtm, thanks!

PR #183, opened 2021-07-19
sdk/state: move validation code to own methods

acharb:
**WHAT**
move `Confirm<open,payment,close>` validation code to their own `validate<open,payment,close>` methods

**WHY**
now that the validation code is growing it makes sense to move to their own methods. This helps with both readability and with unit testing that validation code

this refactoring made sense while working on: https://github.com/stellar/experimental-payment-channels/issues/142

### DIFF
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -144,7 +144,7 @@ func (c *Channel) validateOpen(m OpenAgreement) error {
 func (c *Channel) ConfirmOpen(m OpenAgreement) (open OpenAgreement, authorized bool, err error) {
 	err = c.validateOpen(m)
 	if err != nil {
-		return m, authorized, fmt.Errorf("validating open agreement failed: %w", err)
+		return m, authorized, fmt.Errorf("validating open agreement: %w", err)
 	}
 
 	// at the end of this method, if no error, then save a new channel openAgreement. Use the
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -121,12 +121,6 @@ func (c *Channel) validateOpen(m OpenAgreement) error {
 		return fmt.Errorf("input open agreement expire too far into the future")
 	}
 
-	// If the open agreement has extra signatures, error.
-	if len(m.DeclarationSignatures) > 2 || len(m.CloseSignatures) > 2 || len(m.FormationSignatures) > 2 {
-		return fmt.Errorf("input open agreement has too many signatures, has declaration: %d,"+
-			" close: %d, formation: %d, max of 2 allowed for each",
-			len(m.DeclarationSignatures), len(m.CloseSignatures), len(m.FormationSignatures))
-	}
 	return nil
 }
 
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -110,6 +110,26 @@ func (c *Channel) ProposeOpen(p OpenParams) (OpenAgreement, error) {
 	return open, nil
 }
 
+func (c *Channel) validateOpen(m OpenAgreement) error {
+	// If the open agreement details don't match the open agreement in progress, error.
+	if !c.openAgreement.isEmpty() && !m.Details.Equal(c.openAgreement.Details) {
+		return fmt.Errorf("input open agreement details do not match the saved open agreement details")
+	}
+
+	// If the expiry of the agreement is past the max expiry the channel will accept, error.
+	if m.Details.ExpiresAt.After(time.Now().Add(c.maxOpenExpiry)) {
+		return fmt.Errorf("input open agreement expire too far into the future")
+	}
+
+	// If the open agreement has extra signatures, error.
+	if len(m.DeclarationSignatures) > 2 || len(m.CloseSignatures) > 2 || len(m.FormationSignatures) > 2 {
+		return fmt.Errorf("input open agreement has too many signatures, has declaration: %d,"+
+			" close: %d, formation: %d, max of 2 allowed for each",
+			len(m.DeclarationSignatures), len(m.CloseSignatures), len(m.FormationSignatures))
+	}
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


sdk/state/payment.go
```diff
@@ -80,19 +80,30 @@ func (c *Channel) ProposePayment(amount int64) (CloseAgreement, error) {
 
 var ErrUnderfunded = fmt.Errorf("account is underfunded to make payment")
 
-// ConfirmPayment confirms a close agreement. The original proposer should only have to call this once, and the
-// receiver should call twice. First to sign the agreement and store signatures, second to just store the new signatures
-// from the other party's confirmation.
-func (c *Channel) ConfirmPayment(ca CloseAgreement) (closeAgreement CloseAgreement, authorized bool, err error) {
-	// If the close agreement details don't match the close agreement in progress, error.
+func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
```
leighmcculloch, line 83: I usually prefer breaking out validations, but in the case of the `ConfirmPayment` function validation can't be done entirely prior to signature verification. The balance field is validated during verification. This means this function does not completely validate a payment.

sdk/state/open.go
```diff
@@ -110,6 +110,26 @@ func (c *Channel) ProposeOpen(p OpenParams) (OpenAgreement, error) {
...
+		return fmt.Errorf("input open agreement expire too far into the future")
+	}
+
+	// If the open agreement has extra signatures, error.
+	if len(m.DeclarationSignatures) > 2 || len(m.CloseSignatures) > 2 || len(m.FormationSignatures) > 2 {
+		return fmt.Errorf("input open agreement has too many signatures, has declaration: %d,"+
+			" close: %d, formation: %d, max of 2 allowed for each",
+			len(m.DeclarationSignatures), len(m.CloseSignatures), len(m.FormationSignatures))
+	}
+	return nil
```
leighmcculloch, line 130: I think I need to see the larger change it fits into to understand why we need the signature length check.

This validation is incomplete because this will only prevent a confirmation from having more than two at the start of the function, but if it has two remote signatures, then the local adds signatures, then it will have three. Maybe shift adding the signature count validation into the PR you're planning that'll verify for extra signatures?

I suspect signature validation will need to happen at the end, or we'll need to restructure verifying and signing signatures so it happens in a single pass, more like how we do it here: https://github.com/stellar/go/blob/90387a8/txnbuild/transaction.go#L1262-L1301.

leighmcculloch (commented):
Couple thoughts inline. Let me know your thoughts.

sdk/state/payment.go
leighmcculloch: I wonder if we can change the way the balance validations work. I think we can if we introduce some way of tracking who sender of a payment is. I'll give this some more thought. This problem is fine to ignore for now I think.

sdk/state/open.go
acharb: the larger change of checking extra signatures can be seen in this PR: https://github.com/stellar/experimental-payment-channels/pull/169 . Removing this dup code from this PR sorry about that.

But that's a great point about needing to check at the end I didn't think about that. Will make that change in that PR

sdk/state/open_test.go
```diff
@@ -222,17 +222,15 @@ func TestConfirmOpen_rejectsDifferentOpenAgreements(t *testing.T) {
 		// invalid ObservationPeriodTime
 		d := oa
 		d.ObservationPeriodTime = 0
-		_, authorized, err := channel.ConfirmOpen(OpenAgreement{Details: d})
-		require.False(t, authorized)
+		err := channel.validateOpen(OpenAgreement{Details: d})
```
leighmcculloch, line 225: I missed this on my first pass: I think it is more important we test the exported interface of our SDK by calling `ConfirmOpen`. I think the result should be equivalent since the open is the first thing that takes place in the ConfirmOpen. What's your motivation for changing it?

sdk/state/payment.go
acharb: ya I noticed that as well that the entire validation wouldn't be in this function. I thought moving the bulk of the validation to a different method was worth the trade off though, since `ConfirmPayment` was getting pretty bloated imo. I can add a comment explaining that, and then we can remove if we find a good way of having all the validations in this method

sdk/state/open.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/183/commits/423700c2602323a4309bd19d3b65e6815fa1eb7d

sdk/state/payment.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/183/commits/423700c2602323a4309bd19d3b65e6815fa1eb7d

sdk/state/open_test.go
acharb: motivation was doing a unit test of `validateOpen` instead of `ConfirmOpen` which is doing more than just validating. One of the reasons I wanted to move these validation checks into their own methods was so we could improve the unit tests. Ie able to test just validation and not everything the `Confirm<>` methods are doing

wdyt?

leighmcculloch: It's a trade off, and I don't think the pay off is worth the cost at this early point. It is easier to perform sweeping refactors and know that what we have still does the same thing it did before if we stick to testing, where possible, the exported surface area. I don't think it is worth optimizing that away yet. Because of that I'd focus on testing the exported interface rather than internals, and focus on behavior driven tests. I think the refactor is still good, but the tests should be able to stay the same.

acharb: fair enough making tests more specific may be preemptive atm, changed back 👍 
https://github.com/stellar/experimental-payment-channels/pull/183/commits/1ccb79048b9b74d49ff7d071858b5633c76adbb7

leighmcculloch (approved):
🎉 Looks great.

PR #184, opened 2021-07-19
sdk/state: move txs back to openTxs

leighmcculloch:
### What
Re-add building of the close txs to the open txs functions.

### Why
I removed the functionality in #181, but realized that was a mistake. For CAP-40 we will need to build the close txs to be able to build the open txs because the hashes of the close txs are embedded in the formation tx.


acharb (approved):
lgtm

PR #190, opened 2021-07-21
sdk/state: implement CAP-40 on open

leighmcculloch:
For #179


sdk/state/open.go
```diff
@@ -90,14 +103,40 @@ func (c *Channel) openTxs(d OpenAgreementDetails) (decl, close, formation *txnbu
...
+	declTx, closeTx, formationTx, err := c.openTxs(openAgreement.Details)
 	if err != nil {
-		return nil, fmt.Errorf("building declaration and close txs for latest close agreement: %w", err)
+		return nil, fmt.Errorf("building txs for for open agreement: %w", err)
 	}
 	formationTx, err = formationTx.AddSignatureDecorated(openAgreement.FormationSignatures...)
 	if err != nil {
 		return nil, fmt.Errorf("attaching signatures to formation tx for latest close agreement: %w", err)
 	}
+	for _, s := range openAgreement.DeclarationSignatures {
```
acharb, line 114: 💡 for these steps of finding the specific signature I think we should add comments. w/o context it's not completely clear why we have to do this imo

sdk/txbuild/formation.go
```diff
@@ -7,26 +7,50 @@ import (
...
+	StartSequence     int64
+	Asset             txnbuild.Asset
+	ExpiresAt         time.Time
+	DeclarationTxHash [32]byte
+	CloseTxHash       [32]byte
+	ConfirmingSigner  *keypair.FromAddress
 }
 
 func Formation(p FormationParams) (*txnbuild.Transaction, error) {
+	extraSignerKeys := [2]xdr.SignerKey{}
```
acharb, line 27: 💭 similarly a comment here

acharb (approved):
overall makes sense to me 👍 . It was a bit hard for me to review because we aren't adding tests in this PR, but glad those are coming in https://github.com/stellar/experimental-payment-channels/issues/193

added a couple spots I think could be good candidates for comments as well

sdk/txbuild/formation.go
leighmcculloch: Good point. c4ae700

sdk/state/open.go
leighmcculloch: Good point. This code was gnarly enough to author let alone read. 07d1f8c

PR #192, opened 2021-07-21
all: update quickstart to include CAP-21 extraSigners and CAP-40

leighmcculloch:
### What
Update the quickstart image to include CAP-21 extraSigners and CAP-40. Also added in additional retry mechanics.

### Why
#189 and #190 introduce use of CAP-21 extraSigners and CAP-40, but I want to introduce the upgrade of the quickstart image and the new XDR independent of using the new functionality.

The change includes additional retry mechanics because after updating Horizon and Core, which included bringing in 2 months of changes from both respective main/master branches, there is an odd condition occurring when trying to submit a regular payment transaction during test setup. The payment occurs immediately after creating the new account and for some reason Horizon sees the transaction get a bad sequence number result from Core, but discards the response. Immediately resubmitting the transaction is successful. It's going to take a little while to nut this problem out and to understand if this is a problem in Core or Horizon, and if this is a problem in code that exists on the master branch of either or in our forks. The retry is a hold over so that we can keep working an building on CAP-40 without being blocked on this issue.

For #179


sdk/state/integrationtests/main_test.go
```diff
@@ -2,13 +2,15 @@ package integrationtests
...
+	"net/http"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/stellar/go/clients/horizonclient"
 )
 
-const horizonURL = "http://localhost:8000"
+const horizonURL = "http://host.docker.internal:8000"
```
leighmcculloch, line 13:
```suggestion
const horizonURL = "http://localhost:8000"
```

sdk/state/integrationtests/helpers_test.go
```diff
@@ -215,13 +225,15 @@ func randomPositiveInt64(t *testing.T, max int64) int64 {
...
+func retry(t *testing.T, maxAttempts int, f func() error) (err error) {
+	t.Helper()
 	for i := 0; i < maxAttempts; i++ {
 		err = f()
 		if err == nil {
 			return
 		}
-		time.Sleep(time.Second)
+		t.Logf("failed attempt %d at performing a retry-able operation: %v", i, err)
+		time.Sleep(5 * time.Second)
```
acharb, line 236: ❓ is a full 5 seconds needed? In thoery seems long to me and could slow down the workflows a noticeable amount. But if 5 is what you found is needed empirically then sgtm 👍

acharb (approved):
lgtm 👍

sdk/state/integrationtests/helpers_test.go
leighmcculloch: Let's try 2.
```suggestion
		time.Sleep(2 * time.Second)
```

Issue #107, opened 2021-06-25
sdk/state: prevent new payments after a coordinated close


leighmcculloch:
We should prevent new payments if a coordinated close has been generated since a coordinate close once authorized can always be executed instantly invalidating all following payments.

PR #197, opened 2021-07-22
sdk/state: reject payment if a coordinated close already accepted

acharb:
**WHAT**
in ProposePayment and ConfirmPayment check to see if there's an accepted coordinated close , and if so don't allow new payments

**WHY**
with a fully signed coordinated close a participant could submit a C_i/D_i near instantly, which invalidates later payments

closes https://github.com/stellar/experimental-payment-channels/issues/107

### DIFF
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -115,12 +115,14 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 		c.latestUnauthorizedCloseAgreement.Details.ObservationPeriodLedgerGap == 0 {
 		return fmt.Errorf("cannot propose payment after proposing a coordinated close")
 	}
+
 	// If a coordinated close has been accepted already, error.
 	if !c.latestAuthorizedCloseAgreement.isEmpty() && c.latestAuthorizedCloseAgreement.Details.ObservationPeriodTime == 0 &&
 		c.latestAuthorizedCloseAgreement.Details.ObservationPeriodLedgerGap == 0 {
 		return fmt.Errorf("cannot confirm payment after an accepted coordinated close")
 	}
 
+	// If the new close agreement details are incorrect, error.
 	if ca.Details.IterationNumber != c.NextIterationNumber() {
 		return fmt.Errorf("invalid payment iteration number, got: %d want: %d", ca.Details.IterationNumber, c.NextIterationNumber())
 	}
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -115,7 +115,7 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 	if !c.latestUnauthorizedCloseAgreement.isEmpty() && !ca.Details.Equal(c.latestUnauthorizedCloseAgreement.Details) {
 		return fmt.Errorf("close agreement does not match the close agreement already in progress")
 	}
-	if ca.Details.ConfirmingSigner != nil && ca.Details.ConfirmingSigner.Address() != c.localSigner.Address() &&
+	if ca.Details.ConfirmingSigner.Address() != c.localSigner.Address() &&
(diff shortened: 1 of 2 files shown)


sdk/state/payment.go
```diff
@@ -111,6 +119,10 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 		ca.Details.ConfirmingSigner.Address() != c.remoteSigner.Address() {
 		return fmt.Errorf("close agreement confirmer does not match a local or remote signer, got: %s", ca.Details.ConfirmingSigner.Address())
 	}
+	if !c.latestAuthorizedCloseAgreement.isEmpty() && c.latestAuthorizedCloseAgreement.Details.ObservationPeriodTime == 0 &&
```
acharb, line 122: I was considering adding a helper method like `isAuthorized()` instead for this check, but thought using `!isEmpty()` is good enough

sdk/state/payment.go
```diff
@@ -54,6 +54,13 @@ func (c *Channel) ProposePayment(amount int64) (CloseAgreement, error) {
 	if amount <= 0 {
 		return CloseAgreement{}, errors.New("payment amount must be greater than 0")
 	}
+
+	// If a coordinated close has been accepted already, error.
```
leighmcculloch, line 58: ❗ We should also check if we have proposed a coordinated close by checking if the `latestUnauthorizedCloseAgreement` has zeroes too. If we've proposed a coordinated close we shouldn't ever propose or sign any new close agreements, even if we never get a response back from the other participant. The reason being that they could sign and submit that close at anytime, even if we've both agreed to new close agreements after it.

sdk/state/payment.go
leighmcculloch: ❗ Same comment as above, we need to check the unauthorized as well.

leighmcculloch (commented):
One more thing that needs adding, otherwise this looks great.

sdk/state/payment.go
acharb: 👍 added here: https://github.com/stellar/experimental-payment-channels/pull/197/commits/e342078733285002ecf2f183ae7586a23744d240

sdk/state/payment.go
```diff
@@ -110,16 +110,16 @@ var ErrUnderfunded = fmt.Errorf("account is underfunded to make payment")
...
 	// If a coordinated close has been proposed by this channel already, error.
 	if !c.latestUnauthorizedCloseAgreement.isEmpty() && c.latestUnauthorizedCloseAgreement.Details.ObservationPeriodTime == 0 &&
 		c.latestUnauthorizedCloseAgreement.Details.ObservationPeriodLedgerGap == 0 {
 		return fmt.Errorf("cannot propose payment after proposing a coordinated close")
 	}
+	// If a coordinated close has been accepted already, error.
+	if !c.latestAuthorizedCloseAgreement.isEmpty() && c.latestAuthorizedCloseAgreement.Details.ObservationPeriodTime == 0 &&
+		c.latestAuthorizedCloseAgreement.Details.ObservationPeriodLedgerGap == 0 {
+		return fmt.Errorf("cannot confirm payment after an accepted coordinated close")
+	}
```
leighmcculloch: Why did these need to be reordered? The commit is named "fix test", but I can't see how reordering these would result in changed behavior.

leighmcculloch (approved):
🎉 Looks great. I have one question, but looks great regardless.

sdk/state/payment.go
acharb: technically it was already impossible to propose a new payment if you proposed a coordinated close and it's pending
1. if you propose a payment with `O` period of not zero to match the `latestAuthorized` details, you'll fail the compare against `latestUnauthorized` check
2. if you propose a payment with `O` period of zero to match the new `latestUnauthorized` details, you'll fail the compare against `latestAuthorized` check

so to make the check explicit it has to go first 😄

PR #198, opened 2021-07-22
sdk/state: check ConfirmingSigner isn't nil

acharb:
**WHAT**
in close and payment check that the given close agreement doesn't have a nil `ConfirmingSigner` before checking it

**WHY**
otherwise the check throws a runtime error if the other party gives you a close agreement with an empty `ConfirmingSigner`

ran into this while working on https://github.com/stellar/experimental-payment-channels/pull/197


leighmcculloch (approved):
💯 Thanks for catching and fixing this.

Also, I've been meaning to make it easier for us to use Keypairs by adding an Equal method to them that we could use safely. I opened a PR for that for us to use in the future: https://github.com/stellar/go/pull/3776/files

PR #200, opened 2021-07-23
sdk/state: update cap-40 signer to use xor in hint

### LINKED ISSUE #196: cap, stc, core, go sdk, sdk/state: modify signed payload hint to be xor
After a conversation with @stanford-scs it became clear we can use the xor of the payload and the key as the hint. It makes things a little more complex, but only in the application layer and will likely be better for Core.

leighmcculloch:
### What
Update the payload signatures so they use the xor hint introduced by stellar/stellar-protocol#1011.

### Why
Core and horizon have been updated in stellar/docker-stellar-core-horizon@bace27e to use the new xor hint, and the change is breaking and needs updating here.

Close #196


sdk/state/close.go
```diff
@@ -74,7 +74,13 @@ func (c *Channel) CloseTxs() (declTx *txnbuild.Transaction, closeTx *txnbuild.Tr
...
 		}
 		if signed {
-			declTx, err = declTx.AddSignatureDecorated(s)
+			var closeTxHash [32]byte
+			closeTxHash, err = closeTx.Hash(c.networkPassphrase)
+			if err != nil {
+				return nil, nil, fmt.Errorf("hashing close tx for including payload sig in declaration tx: %w", err)
+			}
+			payloadSig := xdr.NewDecoratedSignatureForPayload(s.Signature, s.Hint, closeTxHash[:])
+			declTx, err = declTx.AddSignatureDecorated(payloadSig)
```
leighmcculloch, line 83: These functions adding signatures to things are getting pretty gnarly. I think it is worth us trying a refactor where each signature gets its own field. I think it will simplify some things a lot. I'll probably keep the refactor for another time though.

acharb (approved):
I'm a bit fuzzy on what `xdr.NewDecoratedSignatureForPayload` is doing exactly atm. But from this project's perspective lgtm

leighmcculloch:
> I'm a bit fuzzy on what `xdr.NewDecoratedSignatureForPayload` is doing exactly atm. But from this project's perspective lgtm

The function creates a decorated signature that can be used to satisfy the payload signer. A payload signer's siganture has a different `hint` than the same signature not for a payload signer. Normally the hint would simply be the hint of the key that signed. But a payload signer includes the hint of the payload in it.

So when the close signature is attached to the close transaction is has the regular hint.

But when the close signature is attached to the declaration transaction to satisfy the signed payload signer, it has a modified hint.

The for payload functions in the Go SDK I added are a bit hard to understand, but of all the different functions I tried writing they were the least obtuse. The exact APIs that'll end up in the Go SDK I think will be something the Horizon team will help provide some guidance on because it is pretty confusing.

Issue #202, opened 2021-07-23
sdk/state: ingest transactions and update state with missing signatures


leighmcculloch:
The channel needs some way of having its internal state updated by transactions that can be seen on the network. I think this functionality largely belongs in the state package, but maybe parts of it are a new package.

If the state package sees a transaction that is the declaration for an unauthorized close agreement that the participant never got the confirmation for it should use the transaction to populate signatures we didn't get making the local close agreement authorized now.

I don't think it is possible to implement some bad actor tests easily (#193) without this.

Related to #214 #215 #218

PR #204, opened 2021-07-23
sdk/state: ingest decl tx to collect missing signatures

### LINKED ISSUE #193: sdk/state: add integration tests that exercise bad actor scenarios relating to CAP-40
@acharb pointed out there are bad actor scenarios we aren't testing relating to CAP-40, and that it'd be beneficial if the integration tests proved that a bad actors actions are prevented by showing that a transaction fails, or are recoverable by doing things like contesting a close.

We probably need two scenarios tested:
1. Demonstration that it is not possible to submit a declaration transaction without the extra signers, so that it isn't possible to submit the declaration transaction without disclosing all the signatures required for the close tx.
2. How the SDK is used to recover signatures from, which will probably highlight missing capabilities.

acharb:
### What
Add the start of the ingestion functionality to the SDK where the SDK will accept as input transactions that have been seen on network and update its internal state based off that seen transaction.

Also add an integration test that uses the ingestion and shows how CAP-40's supporting transaction signature disclosure makes it possible for a participant to recover signatures for other transactions in the transaction set even if the other participant doesn't share those signatures with them directly.

### Why

We don't have an integration test for these scenarios that relies on it, and in the process of writing the integration tests the gaps in the SDK were identified.

Close #193 #202

Introduce TODOs comments for #214 #215

Co-authored-by: Leigh McCulloch <351529+leighmcculloch@users.noreply.github.com>


sdk/state/ingest_test.go
```diff
@@ -0,0 +1,80 @@
...
+	// Pretend that responder broadcasts the declaration tx from the authorized
+	// agreement to the network, making it visible to initiator.
+	declTx, _, err := responderChannel.CloseTxs()
+	require.NoError(t, err)
+	err = initiatorChannel.IngestTx(declTx)
+	require.NoError(t, err)
+
+	// The initiator channel and responder channel should have the same close
+	// agreements.
+	initiatorChannel.Balance()
```
marcelosalloum, line 76: Do you need this line of code?
```suggestion
```

sdk/state/integrationtests/state_test.go
```diff
@@ -810,3 +813,195 @@ func TestOpenUpdatesCoordinatedCloseCoordinateThenStartCloseByRemote(t *testing.
...
+	initiatorChannel, responderChannel := initChannels(t, initiator, responder)
+
+	s := initiator.EscrowSequenceNumber + 1
+	i := int64(1)
+	e := int64(0)
+	t.Log("Vars: s:", s, "i:", i, "e:", e)
+
+	// Open
+	t.Log("Open...")
+	{
```
marcelosalloum, line 842: TIL! I had never seen curly brackets being used to separate scopes.

Any specific reason for separating the scopes here? If I understood correctly you're basically using it to isolate test cases inside this test function.

marcelosalloum (approved):
LGTM!

sdk/state/integrationtests/state_test.go
leighmcculloch: To make it really clear where some parts start and end. It also isolates variables created in that scope so they can't be used later, which prevents me from accidentally using values for one piece in another.

sdk/state/ingest_test.go
leighmcculloch: No I don't, good catch. Removing in #222.

PR #205, opened 2021-07-24
sdk/state: simplify some equal funcs using new stellar/go funcs

leighmcculloch:
### What
Use the new Equal funcs that were added to the keypair package in the stellar/go repo for comparing keypairs.

### Why
To simplify difficult to read code.

### DIFF
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -28,7 +28,7 @@ type CloseAgreementDetails struct {
 func (d CloseAgreementDetails) Equal(d2 CloseAgreementDetails) bool {
 	// TODO: Replace cmp.Equal with a hand written equals.
 	type CAD CloseAgreementDetails
-	return cmp.Equal(CAD(d), CAD(d2), cmp.AllowUnexported(keypair.FromAddress{}))
+	return cmp.Equal(CAD(d), CAD(d2))
 }
 
 // CloseAgreement contains everything a participant needs to execute the close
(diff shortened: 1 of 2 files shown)


sdk/state/payment.go
```diff
@@ -28,7 +28,7 @@ type CloseAgreementDetails struct {
 func (d CloseAgreementDetails) Equal(d2 CloseAgreementDetails) bool {
 	// TODO: Replace cmp.Equal with a hand written equals.
 	type CAD CloseAgreementDetails
-	return cmp.Equal(CAD(d), CAD(d2), cmp.AllowUnexported(keypair.FromAddress{}))
```
marcelosalloum, line 31: Q: why are you removing `cmp.AllowUnexported(keypair.FromAddress{})` here?

marcelosalloum (approved):
LGTM! Great simplification 😄

sdk/state/payment.go
leighmcculloch: It's related to the use of the Equal function which was added to the keypair package in the go repo. When the Equal function wasn't present the cmp package uses reflection to check each exported field. The keypairs have an unexported field and this option tells the cmp package to compare the unexported fields of that type. Now that the Equal function exists the cmp package will use the Equal function instead of reflection.

marcelosalloum: Sounds good, thanks for explaining.

PR #207, opened 2021-07-26
sdk/state: use keypair Equal in more places

leighmcculloch:
### What
Use keypair.Equal in more places rather than do manual comparisons.

### Why
I updated how we compare keypairs in some places in #205 and missed these other two places.

Note that the keypair.Equal function safely handles the case when the left-hand-side or right-hand-side are nil, so there is no need for the nil checks anymore either.


marcelosalloum (approved):
LGTM!

PR #213, opened 2021-07-27
sdk/state: make integration test horizon url configurable

leighmcculloch:
### What
Make the Horizon URL that is used for integration tests configurable.

### Why
A lot of the time, including on CI, the URL being `localhost:8000` works fine. For those of us who use Docker and are hosting the quickstart image's port on the host port on Mac, we need it to be set to `host.docker.internal:8000`. Using an environment variable is a simple low effort solution to make it configurable.


marcelosalloum (approved):
LGTM!

PR #220, opened 2021-07-28
sdk/msg: add rudimentary messages

### LINKED ISSUE #216: sdk/msg: create a package with a set of messages for networked communication of a channel
To coordinate payment channels between parties and increase interoperability there will need to be a common message format that is used for communication.

This package probably won't be that final message format, but we need something rudimentary that makes prototyping and experimenting fast. For this reason this should be a smaller scoped work with less focus on getting a perfect message format, but rather defining the smallest possible implementation. For the sake of simplicity this should probably be JSON and as much as possible communicate the raw state that is exposed from the `sdk/state` package in the `OpenAgreement` and `CloseAgreement` types.

The `examples/console` application should inform the messages defined here.

Related to #217.

leighmcculloch:
### What
Add a rudimentary bare minimum set of messages that can be used to coordinate a `sdk/state` payment channel.

### Why
We need some basic messages so that in every example we aren't reinventing some new message format. At some point we can define an optimal message format, likely XDR based, but right now we only need something that works and JSON is easy.

Close #216


sdk/msg/msg.go
```diff
@@ -0,0 +1,53 @@
...
+	TypePaymentRequest  Type = 300
+	TypePaymentResponse Type = 301
+	TypeCloseRequest    Type = 400
+	TypeCloseResponse   Type = 401
+)
+
+type Message struct {
+	Type Type
+
+	Hello *Hello `json:",omitempty"`
```
marcelosalloum, line 26: Can we rename "Hello" to something more straightforward? Maybe "Content" would be easier to understand.

marcelosalloum (approved):
LGTM!

sdk/msg/msg.go
leighmcculloch: In protocol that have a handshake or introduction phase, a `Hello` message is a common term to use for the first message in a handshake process where participants introduce themselves. TLS has a ClientHello and ServerHello ([eg](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.1.2)). Stellar Core has a Hello message ([eg](https://github.com/stellar/stellar-core/blob/ff56e7f207b918aae53710abcc001511ead7ece9/src/xdr/Stellar-overlay.x#L32-L43)). Unlike the other messages there's no request and response for the hello.

Lmk if I've misunderstood your concern.

marcelosalloum: TIL!

This addresses my concern, thanks for explaining.

Issue #194, opened 2021-07-22
sdk/account: create package with utilities for verifying the state of accounts


leighmcculloch:
The payment channel in sdk/state is dependent on the state of the escrow accounts being safe for the participants. There are a number of things the open transactions in the channel cannot ensure or prevent and the accounts must be analyzed after the open is complete to verify that the channel is indeed safe to use.

Some of the things that need checking are:
- If there are extra signers on the accounts that would allow anyone to control the account unilaterally.
- Are the trustlines authorized.

This is not an exhaustive list of things that must be checked on an account.

PR #221, opened 2021-07-29
sdk/agent: add rudimentary logic for coordinating networked channel

### LINKED ISSUE #217: sdk/agent: create a package that provides rudimentary functionality to coordinate a networked channel
To make a payment channel embeddable with a minimal set of code we probably need to provide functionality to coordinate over a network connection using a common set of messages (#216).

This new package probably won't be that final agent, but we need something rudimentary that makes prototyping and experimenting fast. Hopefully we learn from it and can iterate on it to a place that's usable in a production application in future milestones. Worse case we learn from it, scrap it, and write a new agent.

The agents role is to coordinate three things:
- An active TCP connection with another participant on the other side. Possibly this could just be an `io.ReaderWriterCloser` and the agent doesn't need to know it is a networked connection.
- Horizon:
   - As an output to submit txs to the network.
   - As an input for state updates that come from it.
   - As an input for account state verification at channel open, using the not-yet born `sdk/account` package (#194).
- The channel state, controlled by the `sdk/state` package.

Using @nikhilsaraf's [document](https://docs.google.com/document/d/10EuAvYm7j3IKpSHDEsCJsmG7q7VsALio_b2P8h5skPo) and [research](https://docs.google.com/document/d/1-u5F1rcri3p4O0bmJ7bI9XI2IDmqK3Bpk66H4VpsvOE/edit#heading=h.o3rsevsg7x1e) as guides, and influenced by how the SDK has evolved, what we've learnt so far, and limiting scope to what we need for a single payment channel, I think we probably need an API similar to the below. Whoever tackles this issue should take full reign to build whatever API is more appropriate though.

```
type Agent struct {
	conn io.ReaderWriterCloser
	// ...
}

type Event struct {
	// ...
}

type EventHandler func(e Event)

func (a *Agent) SetConn(io.ReaderWriterCloser)
func (a *Agent) Open(...) error
func (a *Agent) Pay(amount string) error
func (a *Agent) Close() error

func (a *Agent) RegisterErrorHandler(func(error))
func (a *Agent) RegisterOpenHandler(func())
func (a *Agent) RegisterPaymentHandler(amount string)
func (a *Agent) RegisterCloseHandler(finalAmount string)
```

For the scope of this issue I think we should defer having the agent persist state to disk, but it should expose something that an application could be serialized to a local file using [encoding/gob](https://pkg.go.dev/encoding/gob) or [encoding/json](https://pkg.go.dev/encoding/json).

For the scope of this issue I also think we should hardcode some negotiable things like the observation period and max open expiry period.

The `examples/console` application contains a rudimentary `Agent` which may be usable as a starting point.

leighmcculloch:
### What
Port the existing rudimentary agent from the example to the sdk/agent package.

### Why
So that we can iterate on that agent being usable in multiple examples, or used by others.

Close #217


examples/console/main.go
```diff
@@ -83,20 +93,27 @@ func run() error {
 	escrowAccountKey := keypair.MustRandom()
 	fmt.Fprintln(os.Stdout, "escrow account:", escrowAccountKey.Address())
 
-	submitter := &Submitter{
-		HorizonClient:     horizonClient,
+	horizon := &horizon.Horizon{
+		HorizonClient: horizonClient,
+	}
+	submitter := &submit.Submitter{
+		SubmitTxer:        horizon,
```
accordeiro, line 100: Should this be `SubmitTxEr`? Not sure what the convention is, but this was a bit confusing to me.

accordeiro (approved):
@leighmcculloch and I discussed offline – this is a spike for now to move the agent code to the actual SDK, and there are known issues that will be addressed in follow up PRs.

examples/console/main.go
leighmcculloch: No SubmitTxer because it is an interface for the SubmitTx function. It's a horrible name, I couldn't think of anything else in the moment but I can change it to something better. Originally I had `Submitter` but then now there are two different types with the same name. Interestingly Horizon's own submission code has this exact problem, it has two Submitter types that wrap each other and do different things, so I at least don't feel alone in this predicament.

leighmcculloch:
This is a work in progress and the agent package should be considered not something really for use yet outside of its existing example. I'm going to merge this and continue to iterate. There are follow up issues for specific things and I'll create more issues in addition to them as well. #224 #225 #194 #226

PR #222, opened 2021-07-29
sdk/state: remove unused code in test

leighmcculloch:
### What
Remove line of code that's result is unused and has no effect.

### Why
@marcelosalloum reported it in #204 but the PR was auto-merged before it could be fixed.


marcelosalloum (approved):
LGTM!

Issue #103, opened 2021-06-25
sdk/state: don't allow unauthorized states to be discarded


leighmcculloch:
We need to make sure to preserve any unauthorized close agreement that we haven't received signatures from the other participant. This is necessary to ensure the participants can close a channel in any event that a participant falls behind confirming and but submits a declaration without sharing signatures for the close with the other participant. This problem is highlighted by the changes we've made to use CAP-40 (#176), but existed before it as well.

This probably means making the `lastUnauthorizedCloseAgreement` a list of agreements, where whenever agreement in the list gets updated it moves itself to `lastAuthorizedCloseAgreement` and discards any agreements that proceed it.

A simpler solution to this would be to say that only one unauthorized agreement can exist at a time, which is essentially what we have now, although it is unenforced, and to enforce that.

I think we should do the simpler solution until we find a need for supporting multiple in flight at once. I think the protocol should be able to support multiple, which it does, but the SDK doesn't need to initially.

acharb:
if I understand this correctly the edge case is:
```
A proposes {ca1}
B confirms {ca1}, but doesn't send sigs back to A
A proposes {ca2}, B submits ca1, and A can't finish ca1 because second proposal cleared its partially finished {ca1}
```
the simpler solution of enforcing that A cannot propose {ca2} until {ca1} is finished makes sense to me (at the expense of TPS I think), but I'm wondering shouldn't A be able to get what it needs from the {ca1} that B broadcasted to the network?

PR #228, opened 2021-08-04
sdk/state: enforce only one close agreement open at a time

acharb:
**WHAT**
enforce that only one unauthorized close agreement can exist at a time

**WHY**
currently this edge case exists:
```
A proposes {ca1}
B confirms {ca1}, but doesn't send sigs back to A
A proposes {ca2}, B submits ca1, and A can't finish ca1 because proposing cleared the latest unauthorized
```
by enforcing that A cant propose a {ca2}, which clears its {ca1}, it prevents this edge case and protects A.

closes https://github.com/stellar/experimental-payment-channels/issues/103

note:
- if A wants to close the channel after {ca1}, A would need to do an uncoordinated close, as this change prevents them from proposing a coordinated close which would also clear its unfinished {ca1}.

### DIFF
--- a/sdk/state/payment_test.go
+++ b/sdk/state/payment_test.go
@@ -846,3 +846,62 @@ func TestChannel_ProposeAndConfirmPayment_rejectIfAfterCoordinatedClose(t *testi
 	_, err = senderChannel.ConfirmPayment(p)
 	require.EqualError(t, err, "validating payment: cannot confirm payment after an accepted coordinated close")
 }
+
+func TestChannel_enforceOnlyOneCloseAgreementAllowed(t *testing.T) {
+	localSigner := keypair.MustRandom()
+	remoteSigner := keypair.MustRandom()
+	localEscrowAccount := &EscrowAccount{
+		Address:        keypair.MustRandom().FromAddress(),
+		SequenceNumber: int64(101),
+		Balance:        int64(100),
+	}
+	remoteEscrowAccount := &EscrowAccount{
+		Address:        keypair.MustRandom().FromAddress(),
+		SequenceNumber: int64(202),
+		Balance:        int64(100),
+	}
+
+	senderChannel := NewChannel(Config{
+		NetworkPassphrase:   network.TestNetworkPassphrase,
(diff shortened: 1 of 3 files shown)


sdk/state/payment_test.go
```diff
@@ -854,3 +854,62 @@ func TestChannel_ProposeAndConfirmPayment_rejectIfChannelNotOpen(t *testing.T) {
...
+	_, err = senderChannel.ProposePayment(10)
+	require.NoError(t, err)
+
+	// sender should not be able to propose a second payment until the first is finished
+	_, err = senderChannel.ProposePayment(20)
+	require.EqualError(t, err, "cannot start a new payment while an unfinished one exists")
+
+	// sender should not be able to propose coordinated close while unfinished payment exists
+	_, err = senderChannel.ProposeClose()
+	require.EqualError(t, err, "cannot propose coordinated close while an unfinished payment exists")
```
marcelosalloum, line 914: In the PR description, you mentioned `A` should be able to do an uncoordinated close. Can you do it in this test just to show uncoordinated closes won't error?

marcelosalloum (commented):
LGTM! I left a question below and can approve after you answer/address that.

sdk/state/payment_test.go
acharb: ah good question, it's actually already covered in the integration tests [(here)](https://github.com/stellar/experimental-payment-channels/blob/df1e4d83b13c7dea65836f31109db8740feeb202/sdk/state/integrationtests/state_test.go#L967-L993)

The uncoordinated close case isn't tested as a unit test because the sdk is currently supporting coordinated closes only by design (the happy case), so the integration tests are covering the uncoordinated closes cases (where something non-happy happens).

I added this check in the unit test though: https://github.com/stellar/experimental-payment-channels/pull/228/commits/2564ea5e34c0c7513f489957c478746c75b5848d
to make sure the original close agreement isn't lost (which means they can do an uncoordinated close with it if needed)

marcelosalloum (approved):
LGTM!

Issue #186, opened 2021-07-20
sdk/state: prevent proposing or confirming payments or closes if not open


leighmcculloch:
The SDK lets a caller propose or confirm when the channel isn't open. It should error in these cases.

PR #230, opened 2021-08-05
sdk/state: error if proposing or confirming a payment or close if channel is not open yet

acharb:
**WHAT**
error if the channel is trying to propose or confirm a payment or close if the channel is not open yet

**WHY**
a channel shouldn't be able attempt to propose/confirm a payment or close if the channel is not open yet. Although the methods would fail, checking explicitly prevents unexpected behavior.

closes https://github.com/stellar/experimental-payment-channels/issues/186

### DIFF
--- a/sdk/state/integrationtests/helpers_test.go
+++ b/sdk/state/integrationtests/helpers_test.go
@@ -16,7 +16,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// functions to be used in the state_test integration tests
+// functions to be used in the sdk/state/integrationtests integration tests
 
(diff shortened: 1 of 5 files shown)


sdk/state/integrationtests/helpers_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// functions to be used in the state_test integration tests
```
acharb, line 19: we renamed this package so just cleaning this up

marcelosalloum (approved):
LGTM! Nice addition 🎉

Issue #223, opened 2021-07-30
sdk/state: prevent proposing or confirming opens if open or closed


leighmcculloch:
Similar to #186 the sdk/state package should not allow an open to be proposed or confirmed if the channel is already open, or was open and is now closed.

PR #231, opened 2021-08-06
sdk/state: reject new open proposal/confirm if channel is already open

acharb:
**WHAT**
throw an error if a channel has already been opened (and is still open or closed), and trying to propose or confirm a new open.

Also adds a helper method `isFull` for checking if an openAgreement is fully signed, indicating it's submittable and the channel is considered open

**WHY**
explicitly defining that you can't open a channel that has already been opened prevents unexpected channel behavior

closes https://github.com/stellar/experimental-payment-channels/issues/223


marcelosalloum (approved):
LGTM!

Issue #215, opened 2021-07-27
sdk/state: ingest transactions and update state with seq num changes


leighmcculloch:
The channel needs some way of having its internal state updated by transactions that can be seen on the network.

If the state package sees a transaction that alters the sequence number of the initiator escrow account it should be a signal that the channel is closing. This should automatically trigger things like #186 so that the participant can no longer transact.

If the sequence number seen is not the sequence number of the latest authorized agreement the participant should have some way of knowing they need to initiate their own close process.

Related to #202 #214 #218

leighmcculloch:
See #204 for the code to build on.

Issue #218, opened 2021-07-28
sdk/state: ingest transactions and update to be open once open submitted


leighmcculloch:
The channel needs some way of having its internal state updated by transactions that can be seen on the network.

If the state package sees a transaction that opens the channel, i.e. the formation transaction, the state package should consider the channel opened at that point. Once this is implemented this should make it so the channel isn't open until this transaction has been seen on the network.

The logic that ingests the transaction should also assess the transaction's meta xdr to check that the state of the account is good for the channel. The meta xdr contains full details about the state of the initiator's escrow account after the transaction was executed. If the account isn't in a good state, we should consider the channel in an error state.

Related to #202 #214 #215

leighmcculloch:
See #204 for the code to build on.

Issue #199, opened 2021-07-23
sdk/state: store a copy of built txs


leighmcculloch:
The SDK needs to keep ahold of built transactions rather than recreating them when it needs them.

Currently the SDK regenerates transactions whenever it needs them. For the most part this is fine. However, if the Stellar SDK we are using makes a change to how it builds a transaction such that the resulting XDR is different, the transaction hash will be different and the signatures will be invalidated.

It's convenient that the agreements we store in the channel are lightweight and only contain the information that needs to be passed back and forth with the other participant. It would be a-shame to have to include the full transactions in those agreements because the transactions are large in comparison. We should find a way to store the transactions in the channel with the agreement but not inside it.

One option is we could have some internal type that wraps up the existing agreement types and also includes the built transactions. It would then also be easy to also store transaction hashes as well, which would simplify some other code. For example:

```go
type Channel struct {
	// ...

	openPackage openPackage

	latestAuthorizedClosePackage   closePackage
	latestUnauthorizedClosePackage closePackage
```

```go
type openPackage struct {
	OpenAgreement   OpenAgreement
	DeclarationTx   *txnbuild.Transaction
	DeclarationHash [32]byte
	CloseTx         *txnbuild.Transaction
	CloseHash       [32]byte
	FormationTx     *txnbuild.Transaction
	FormationHash   [32]byte
}
```

```go
type closePackage struct {
	CloseAgreement  CloseAgreement
	DeclarationTx   *txnbuild.Transaction
	DeclarationHash [32]byte
	CloseTx         *txnbuild.Transaction
	CloseHash       [32]byte
}
```

PR #232, opened 2021-08-11
sdk/state: ingest transaction and set channel closing state

acharb:
**WHAT**
when ingesting a transactions save the escrow account sequence number on the channel. From that value infer what the channel `CloseState` is.

**WHY**
the channel needs a way of ingesting a broadcasted transaction from the other party, and then adjusting its channel state accordingly

closes https://github.com/stellar/experimental-payment-channels/issues/215

### DIFF
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -29,7 +29,7 @@ func (c *Channel) IngestTx(tx *txnbuild.Transaction) error {
 
 func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Transaction) {
 	// If the transaction's source account is not the initiator's escrow
-	// account, then the transaction is not a part of a close agreement.
+	// account, return.
 	if tx.SourceAccount().AccountID != c.initiatorEscrowAccount().Address.Address() {
 		return
 	}
@@ -46,6 +46,12 @@ func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Tr
 // If the transaction should be able to provide this data and cannot, the
 // function errors.
 func (c *Channel) ingestTxToUpdateUnauthorizedCloseAgreement(tx *txnbuild.Transaction) error {
+	// If the transaction's source account is not the initiator's escrow
+	// account, then the transaction is not a part of a close agreement.
+	if tx.SourceAccount().AccountID != c.initiatorEscrowAccount().Address.Address() {
+		return nil
+	}
+
 	ca := c.latestUnauthorizedCloseAgreement
 
 	// If there is no unauthorized close agreement, then there's no need to try
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -21,12 +21,21 @@ func (c *Channel) IngestTx(tx *txnbuild.Transaction) error {
 
 	// If we found an unauthorized close agreement has begun closing, update our unauthorized to
 	// become authorized.
-	if c.CloseState() == CloseEarlyClosing {
+	closeState, err := c.CloseState()
+	if err != nil {
+		return fmt.Errorf("getting close state: %w", err)
+	}
+
+	if closeState == CloseEarlyClosing {
 		err := c.updateUnauthorizedAgreement(tx)
 		if err != nil {
 			return err
 		}
-		if c.CloseState() != CloseClosing {
+		closeState, err := c.CloseState()
+		if err != nil {
+			return fmt.Errorf("getting close state: %w", err)
+		}
+		if closeState != CloseClosing {
 			return fmt.Errorf("converting unauthorized agreement to authorized should put channel in closing state")
 		}
 	}
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -42,6 +42,9 @@ func (c *Channel) ingestTxToUpdateCloseState(tx *txnbuild.Transaction) error {
 		if err != nil {
 			return err
 		}
+		if c.CloseState() != CloseClosing {
+			return fmt.Errorf("converting unauthorized agreement to authorized should put channel in closing state")
+		}
 	}
 
 	return nil
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -15,18 +15,50 @@ import (
 // determine if successful or not, and understand changes in the ledger as a
 // result.
 func (c *Channel) IngestTx(tx *txnbuild.Transaction) error {
-	// Use the tx to update the latest unauthorized close agreement if possible.
-	err := c.ingestTxToUpdateUnauthorizedCloseAgreement(tx)
+	txHash, err := tx.Hash(c.networkPassphrase)
 	if err != nil {
-		return err
+		return fmt.Errorf("hashing tx: %w", err)
 	}
 
-	// TODO: If the tx hash matches an authorized or unauthorized declaration,
-	// mark the channel as closing.
-	// TODO: If the tx hash matches an authorized or unauthorized close, mark the
-	// channel as closed.
-	// TODO: If the tx is for an older declaration, mark the channel as closing with
-	// requiring bump.
+	// If the transaction is the declTx of our latest unauthorized, update our latest unauthorized
+	// and mark the channel as closing.
+	if !c.latestUnauthorizedCloseAgreement.isEmpty() {
+		unauthorizedDeclTx, _, err := c.closeTxs(c.openAgreement.Details, c.latestUnauthorizedCloseAgreement.Details)
+		if err != nil {
+			return fmt.Errorf("building txs for latest unauthorized close agreement: %w", err)
+		}
+		unauthorizedDeclTxHash, err := unauthorizedDeclTx.Hash(c.networkPassphrase)
+		if err != nil {
+			return fmt.Errorf("hashing latest unauthorized declaration tx: %w", err)
+		}
+		if txHash == unauthorizedDeclTxHash {
+			// Use the tx to update the latest unauthorized close agreement if possible.
+			err = c.ingestTxToUpdateUnauthorizedCloseAgreement(tx)
+			if err != nil {
+				return err
+			}
+			c.setCloseState(1)
+			return nil
+		}
+	}
+
+	// If the transaction is the declTx of our latest authorized, mark the channel as closing.
+	authorizedDeclTx, _, err := c.closeTxs(c.openAgreement.Details, c.latestAuthorizedCloseAgreement.Details)
+	if err != nil {
+		return fmt.Errorf("building txs for latest unauthorized close agreement: %w", err)
+	}
+	authorizedDeclTxHash, err := authorizedDeclTx.Hash(c.networkPassphrase)
+	if err != nil {
+		return fmt.Errorf("hashing latest unauthorized declaration tx: %w", err)
+	}
+	if txHash == authorizedDeclTxHash {
+		c.setCloseState(1)
+		return nil
+	}
+
+	// If the transaction is an older declTx, mark the channel as needs closing.
+	c.setCloseState(2)
+
 	// TODO: Use the transaction result to affect on success/failure.
 
 	return nil
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -19,25 +19,9 @@ func (c *Channel) IngestTx(tx *txnbuild.Transaction) error {
 
 	c.ingestTxToUpdateInitiatorEscrowAccountSequence(tx)
 
-	// If we found an unauthorized close agreement has begun closing, update our unauthorized to
-	// become authorized.
-	closeState, err := c.CloseState()
+	err := c.ingestTxToUpdateUnauthorizedCloseAgreement(tx)
 	if err != nil {
-		return fmt.Errorf("getting close state: %w", err)
-	}
-
-	if closeState == CloseStateEarlyClosing {
-		err := c.updateUnauthorizedAgreement(tx)
-		if err != nil {
-			return err
-		}
-		closeState, err := c.CloseState()
-		if err != nil {
-			return fmt.Errorf("getting close state: %w", err)
-		}
-		if closeState != CloseStateClosing {
-			return fmt.Errorf("converting unauthorized agreement to authorized should put channel in closing state")
-		}
+		return err
 	}
 
 	return nil
@@ -53,7 +37,7 @@ func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Tr
 	c.setInitiatorEscrowAccountSequence(tx.SourceAccount().Sequence)
 }
 
-// updateUnauthorizedAgreement uses the signatures in the transaction to
+// ingestTxToUpdateUnauthorizedCloseAgreement uses the signatures in the transaction to
 // authorize an unauthorized close agreement if the channel has one.
 // This process helps to give a participant who proposed an agreement the
 // ability to close the channel if they did not receive the confirmers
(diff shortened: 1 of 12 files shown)


sdk/state/state.go
```diff
@@ -35,6 +35,24 @@ type Channel struct {
 
 	latestAuthorizedCloseAgreement   CloseAgreement
 	latestUnauthorizedCloseAgreement CloseAgreement
+
+	closeState CloseState
+}
+
+type CloseState int
```
acharb, line 42: a CloseState field is a simplification for now for this PR. It goes against the pattern of the other code in this sdk of getting the channel state from the latest agreements. I couldn't think of a good way of signaling a close state for now, so simplifying for now and going to do that in this ticket: https://github.com/stellar/experimental-payment-channels/issues/233

sdk/state/ingest.go
```diff
@@ -15,71 +15,129 @@ import (
...
 	}
 	if txHash != declTxHash {
-		return nil
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// updateUnauthorizedAgreement uses the signatures in the transaction to
```
acharb, line 115: same method comment above, just renamed the method

sdk/state/state.go
acharb: I refactored it in this commit: https://github.com/stellar/experimental-payment-channels/pull/232/commits/85ceac50beb42dab6dd4ec55c9439789b03b7a58

now the CloseState is inferred from comparing the network sequence number to what we expect it to be. This more goes with the pattern we're currently using where the channel state is inferred from Agreements and data from the network

sdk/state/state.go
```diff
@@ -37,6 +40,46 @@ type Channel struct {
...
 
+type CloseState int
+
+const (
+	CloseError CloseState = iota - 1
+	CloseNone
+	CloseEarlyClosing // a proposed declTx before fully confirmed is submitted
+	CloseClosing      // latest declTx is submitted
+	CloseNeedsClosing // an earlier declTx is submitted
+	CloseClosed
```
leighmcculloch, line 51: Can you use the full name of the type as the prefix?
```suggestion
	CloseStateError CloseState = iota - 1
	CloseStateNone
	CloseStateEarlyClosing // a proposed declTx before fully confirmed is submitted
	CloseStateClosing      // latest declTx is submitted
	CloseStateNeedsClosing // an earlier declTx is submitted
	CloseStateClosed
```

I think we should also reduce this to the states we plan to use now:
```suggestion
	CloseStateNone CloseState = iota - 1
	CloseStateEarlyClosing // a proposed declTx before fully confirmed is submitted
	CloseStateClosing      // latest declTx is submitted
	CloseStateNeedsClosing // an earlier declTx is submitted
	CloseStateClosed
```

Do you think we could get away with a single closing state? If we can, I think we should for now and say that participants always try to close the channel themselves when they see it closing from the other party. We can also say that we always see if we need to copy signatures of declarations submitted too, that way we don't need a state to differentiate that case. It keeps it simpler, probably covers some edge cases automatically by doing so too.
```suggestion
	CloseStateNone CloseState = iota - 1
	CloseStateClosing
	CloseStateClosed
```

Thoughts?

sdk/state/state.go
```diff
@@ -37,6 +40,46 @@ type Channel struct {
...
+)
+
+// CloseState infers the close state from the network sequence number and the expected EI seq num
+func (c *Channel) CloseState() CloseState {
+	latestDeclSequence := txbuild.StartSequenceOfIteration(c.startingSequence, c.latestAuthorizedCloseAgreement.Details.IterationNumber)
+	latestCloseSequence := latestDeclSequence + 1
+
+	latestUnauthorizedDeclSequence := txbuild.StartSequenceOfIteration(c.startingSequence, c.latestUnauthorizedCloseAgreement.Details.IterationNumber)
```
leighmcculloch, line 59: This few lines leak some logic (the seq number offsets of the decl and close txs) that otherwise only exists in the txbuild package, so I'm pondering if there's a way we can keep that logic in the txbuild package.

I think one way we can do that is to get the current latest agreement's txs and get the seq number from them. This would be less efficient, but I don't think we need to be worried about efficiency here yet. And efficiency will be less of an issue once we start caching txs, which we'll need to at some point (#199).

sdk/state/ingest.go
```diff
@@ -15,71 +15,67 @@ import (
...
-	// TODO: Use the transaction result to affect on success/failure.
+	return nil
+}
+
+func (c *Channel) ingestTxToUpdateCloseState(tx *txnbuild.Transaction) error {
+	// If the transaction's source account is not the initiator's escrow
+	// account, then the transaction is not a part of a close agreement.
+	if tx.SourceAccount().AccountID != c.initiatorEscrowAccount().Address.Address() {
+		return fmt.Errorf(`transaction source account is not the initiator escrow account,
+			found: %s, should be: %s`, tx.SourceAccount().AccountID, c.initiatorEscrowAccount().Address.Address())
```
leighmcculloch, line 33: I don't think we want an error in this case. In the current logic the `IngestTx` function will be a no-op with no error if you pass in a tx that has no affect on the channel state. This PR changes that to error when a tx doesn't affect the state.

The ingest function will eventually need to handle tx's from multiple accounts, for handling things like withdrawals, where any account can be the source.

sdk/state/ingest.go
```diff
@@ -15,71 +15,67 @@ import (
 // determine if successful or not, and understand changes in the ledger as a
 // result.
 func (c *Channel) IngestTx(tx *txnbuild.Transaction) error {
-	// Use the tx to update the latest unauthorized close agreement if possible.
-	err := c.ingestTxToUpdateUnauthorizedCloseAgreement(tx)
+	// TODO: Use the transaction result to affect on success/failure.
+
+	err := c.ingestTxToUpdateCloseState(tx)
```
leighmcculloch, line 20: Could we keep `c.ingestTxToUpdateUnauthorizedCloseAgreement` and `c.ingestTxToUpdateCloseState` as separate functions? They appear to update different things and we probably don't gain a lot by coupling them together. We can merge them later or abstract shared state if it looks like it would be beneficial.

leighmcculloch (commented):
Few comments inline.

sdk/state/state.go
```diff
@@ -20,7 +21,9 @@ type Channel struct {
 	networkPassphrase string
 	maxOpenExpiry     time.Duration
 
-	startingSequence int64
+	startingSequence      int64
+	networkEscrowSequence int64 // the network sequence number of the initiator escrow account
```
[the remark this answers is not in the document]
acharb: good point
https://github.com/stellar/experimental-payment-channels/pull/232/commits/50a2685863eaa6ee703459541e6841aa0602fcd7

sdk/state/state.go
acharb: replacing `CloseEarlyClosing` with `CloseClosing` and just always trying to grab the declTx sigs I can see. But if the other party submits an earlier declTx, I think that's a special case we'd want to flag differently? eg. in `CloseClosing` the user doesn't need to take action if they don't want, since this is a safe agreement and the other user can close it. 

With `CloseNeedsClosing` the user _needs_ to take action.

wdyt? If we want to simplify for now I can add a TODO for the `CloseNeedsClosing` case

sdk/state/ingest.go
acharb: separated here
https://github.com/stellar/experimental-payment-channels/pull/232/commits/bc6c6bc6c1ed02e3b54f7fc2f4a29a3c72d22f5b

sdk/state/ingest.go
acharb: 👍 
https://github.com/stellar/experimental-payment-channels/pull/232/commits/bc6c6bc6c1ed02e3b54f7fc2f4a29a3c72d22f5b

sdk/state/state.go
acharb: ya I didn't especially like exporting that either, changed here. A bit less pretty though because it introduces an error possibility in `CloseState()` now
https://github.com/stellar/experimental-payment-channels/pull/232/commits/35568c959e2dc72859dce762fe8c762a0c5fa0f2

sdk/state/state.go
acharb: also `CloseError` is being used currently, if the EI seq num is an unexpected value for some reason, or there's an error in getting `CloseState()`. We could keep that as `CloseStateNone` but I think an error state will be used pretty soon eg. in [this ticket ](https://github.com/stellar/experimental-payment-channels/issues/218)

leighmcculloch: I'm not sure our statuses can capture every situation whe

I still want to remove `CloseError` and to reassign that scenario to another state. It will be harder to represent CloseError in a general state that is about opening and closing. It sounds like `CloseError` indicates that the channel is broken. We should work really hard to make it so no situation can arise where the channel is broken. I'd argue that there are a finite number of states and we shouldn't have anything outside them:

- Not open - Seq < s means not open yet
- Open - Seq = s, formation tx ingested and successful, open
- Closing - Seq > s, signers still engaged, closing
- Closed - Seq > s, signers reverted, closed

I'm simplifying a bit for where there has been no withdrawal, which introduces `e`.

I feel like for the open case we want to be absolutely sure that the formation tx succeeded, but in the closing/closed states it's more about the seq num and the state of signers on the users escrow account (not both escrow accounts).

For the close states though we can use sequence to indicate if there is a tx submitted, then should evaluate the state of the account, which we can do with the tx meta and result data. Using only the sequence might be misleading. For example if I close earlier 

Then if we need it add something for figuring out what state the channel is closing in, but that might not even be a state. Or make it easier to participate in the close whichever case. I feel like if we can do the latter it will result in less code to maintain.

acharb: removed the `CloseStateError` here
https://github.com/stellar/experimental-payment-channels/pull/232/commits/ba856ea6f2a6625d804e92eef3cce67a8e640d29

if alright with you I'd like to push this PR through, and then refactor the CloseState with this ticket https://github.com/stellar/experimental-payment-channels/issues/218 : which will be incorporating the Open states, and taking into account the tx meta and result data as your comment talks about.

wdyt?

sdk/state/ingest.go
```diff
@@ -15,71 +15,70 @@ import (
...
-	// If there is no unauthorized close agreement, then there's no need to try
-	// and update it.
-	if ca.isEmpty() {
-		return nil
-	}
-
-	// Load the declaration and close transactions for the unauthorized close
-	// agreement.
-	declTx, closeTx, err := c.closeTxs(c.openAgreement.Details, ca.Details)
+	declTx, closeTx, err := c.closeTxs(c.openAgreement.Details, c.latestUnauthorizedCloseAgreement.Details)
```
leighmcculloch, line 71: This change appears unnecessary, the close agreement is already loaded above:
```suggestion
	declTx, closeTx, err := c.closeTxs(c.openAgreement.Details, ca.Details)
```

leighmcculloch (commented):
Few more comments inline.

sdk/state/ingest.go
```diff
@@ -15,71 +15,70 @@ import (
...
+			return err
+		}
+		closeState, err := c.CloseState()
+		if err != nil {
+			return fmt.Errorf("getting close state: %w", err)
+		}
+		if closeState != CloseStateClosing {
+			return fmt.Errorf("converting unauthorized agreement to authorized should put channel in closing state")
+		}
+	}
```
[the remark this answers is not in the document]
acharb: gotchya, my thinking was only call `updateUnauthorizedAgreement` when needed, so using the `CloseStateEarlyClosing` state. Calling it every time makes sense to me I should have changed it to that. Changed back to here: https://github.com/stellar/experimental-payment-channels/pull/232/commits/ec6870bb18fe64951de8df8800bb495699a072f4

which has the original code of checking transaction hashes, returning nil if not relevant etc. Also removes the `CloseStateEarlyClosing` since we don't need it anymore

sdk/state/ingest.go
```diff
@@ -15,71 +15,70 @@ import (
...
-	// TODO: Use the transaction result to affect on success/failure.
+	if closeState == CloseStateEarlyClosing {
+		err := c.updateUnauthorizedAgreement(tx)
+		if err != nil {
+			return err
+		}
+		closeState, err := c.CloseState()
+		if err != nil {
+			return fmt.Errorf("getting close state: %w", err)
+		}
```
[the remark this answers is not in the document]
acharb: removed

sdk/state/ingest.go
```diff
@@ -15,71 +15,70 @@ import (
...
-	}
+	// Look for the signatures on the tx that are required to fully authorize
+	// the unauthorized close agreement, then confirm the close agreement.
 	declTxHash, err := declTx.Hash(c.networkPassphrase)
 	if err != nil {
 		return fmt.Errorf("hashing latest unauthorized declaration tx: %w", err)
 	}
-	if txHash != declTxHash {
-		return nil
-	}
```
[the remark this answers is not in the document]
acharb: explained in this comment: https://github.com/stellar/experimental-payment-channels/pull/232#discussion_r690750587 but it was under the assumption we're only calling the `updateUnauthorizedAgreement` method when we know the ingested tx is the declTx. But I changed it back

sdk/state/ingest.go
```diff
@@ -15,71 +15,70 @@ import (
...
+		return fmt.Errorf("unauthorized close agreement is empty")
 	}
 
 	ca := c.latestUnauthorizedCloseAgreement
 
-	// If there is no unauthorized close agreement, then there's no need to try
-	// and update it.
-	if ca.isEmpty() {
-		return nil
-	}
```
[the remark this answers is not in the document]
acharb: explained here: https://github.com/stellar/experimental-payment-channels/pull/232#discussion_r690750587
but it was under the assumption this method was only called when we expected the ingested tx to be the latestUnauthorized declTx. Changed back here though: https://github.com/stellar/experimental-payment-channels/pull/232/commits/ec6870bb18fe64951de8df8800bb495699a072f4

sdk/state/ingest.go
acharb: changed back here: https://github.com/stellar/experimental-payment-channels/pull/232/commits/ec6870bb18fe64951de8df8800bb495699a072f4

sdk/state/ingest.go
```diff
@@ -15,39 +15,37 @@ import (
...
 // If the transaction should be able to provide this data and cannot, the
 // function errors.
 func (c *Channel) ingestTxToUpdateUnauthorizedCloseAgreement(tx *txnbuild.Transaction) error {
-	// If the transaction's source account is not the initiator's escrow
-	// account, then the transaction will not be a transaction for a close
-	// agreement and won't be able to update any unauthorized close agreement.
-	if tx.SourceAccount().AccountID != c.initiatorEscrowAccount().Address.Address() {
-		return nil
-	}
-
```
leighmcculloch, line 50: I think you need to add this back. This function should ignore any tx that isn't for the EI account. (All the changes to this function look superficial and unrelated to setting the channel closing state.)

sdk/state/ingest_test.go
```diff
@@ -64,16 +64,142 @@ func TestChannel_IngestTx_updateUnauthorizedCloseAgreement(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, CloseStateClosing, closeState)
 
 	// The initiator channel and responder channel should have the same close
 	// agreements.
 	assert.Equal(t, int64(8), initiatorChannel.Balance())
 	assert.Equal(t, int64(8), responderChannel.Balance())
 	assert.Equal(t, initiatorChannel.LatestCloseAgreement(), responderChannel.LatestCloseAgreement())
+
+	// TODO - initiatorChannel should be in a close state now, so shouldn't be able to propose/confirm new payments
```
leighmcculloch, line 83: I don't think this TODO is correct, ingesting the declTx shouldn't result in the channel being in a close state, it'll still be in a closing state. It won't enter a closed state until the close tx is submitted and ingested.

sdk/state/state.go
```diff
@@ -37,6 +38,53 @@ type Channel struct {
...
+
+	return -1, fmt.Errorf("initiator escrow account sequence has unexpected value")
+}
+
+func (c *Channel) setInitiatorEscrowAccountSequence(seqNum int64) {
+	if c.initiator {
+		c.localEscrowAccount.SequenceNumber = seqNum
+	} else {
+		c.remoteEscrowAccount.SequenceNumber = seqNum
+	}
```
leighmcculloch, line 85: This function shouldn't need to know how to select the escrow account as we have a function that does that already. I think this should work:
```suggestion
	c.initiatorEscrowAccount().SequenceNumber = seqNum
```

sdk/state/state.go
```diff
@@ -37,6 +38,53 @@ type Channel struct {
 	latestUnauthorizedCloseAgreement CloseAgreement
 }
 
+type CloseState int
+
+const (
+	CloseStateNone         CloseState = iota
+	CloseStateClosing                 // latest declTx is submitted
+	CloseStateNeedsClosing            // an earlier declTx is submitted
```
leighmcculloch, line 46: I think this state's name is deceiving, and a little odd that it describes something that needs to happen rather than describes the channel itself. The state is closing, but it is closing with out-of-date state.
```suggestion
	CloseStateClosingWithOutdatedState
```

sdk/state/ingest_test.go
```diff
@@ -64,16 +64,142 @@ func TestChannel_IngestTx_updateUnauthorizedCloseAgreement(t *testing.T) {
...
+
+	// Pretend that responder broadcasts the old declTx, and
+	// initiator ingests it.
+	err = initiatorChannel.IngestTx(oldDeclTx)
+	require.NoError(t, err)
+	closeState, err := initiatorChannel.CloseState()
+	require.NoError(t, err)
+	require.Equal(t, CloseStateNeedsClosing, closeState)
+
+	// TODO - initiator should not be able to propose/confirm new close agreements
```
leighmcculloch, line 204: FYI I don't think you need to test this in this test, it's sufficient to test that the state is set appropriately, but if you want to I think it's fine too.

sdk/state/ingest_test.go
leighmcculloch: Actually you're right, it should still prevent propose and confirm, but it'll be in a closing state. I misread `be in a close state now` as saying `closed state`.

sdk/state/ingest.go
acharb: got lost in the shuffle, added back here: https://github.com/stellar/experimental-payment-channels/pull/232/commits/79e69df149389d443cd0332ac45329c29e1857bb

interestingly we actually don't need it since the tx hashes are being compared. But I agree to end early if it's not the EI account

sdk/state/state.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/232/commits/3cf217ac37068c9eb1897f5a6961e02429cfa6b6

sdk/state/ingest_test.go
acharb: ya  I guess that is covered in the other unit test coverage https://github.com/stellar/experimental-payment-channels/pull/232/commits/93bdc78262142a14ffd5d11c9c1d65a98fe73b70

it was more of a reminder to make sure those validations check against this new state, but that's noted in this [ticket](https://github.com/stellar/experimental-payment-channels/issues/233#issue-967612416)

leighmcculloch (approved):
👏🏻 I look forward to seeing how the open state and close state come together.

Issue #214, opened 2021-07-27
sdk/state: ingest transactions and update state with balance changes


leighmcculloch:
The channel needs some way of having its internal state updated by transactions that can be seen on the network.

If the state package sees a transaction that alters the balance of either escrow account it should update the internal balance for the account stored in the state package.

Related to #202 #215 #218

leighmcculloch:
See #204 for the code to build on.

PR #234, opened 2021-08-12
sdk/state: ingest new transaction and update escrow balances

acharb:
(can't be merged until / pointing to: https://github.com/stellar/experimental-payment-channels/pull/232)

**WHAT**
ingest a transaction and update the escrow account balances by looking at the transaction result meta. For both a withdrawal and deposit.

**WHY**
if an escrow account balance is changed by a withdrawal or deposit on the network, the channel should know and update its balances to be in sync

closes https://github.com/stellar/experimental-payment-channels/issues/214


sdk/state/ingest.go
```diff
@@ -108,3 +115,37 @@ func (c *Channel) ingestTxToUpdateUnauthorizedCloseAgreement(tx *txnbuild.Transa
 
 	return nil
 }
+
+func (c *Channel) ingestTxToUpdateBalances(tx *txnbuild.Transaction) error {
+	localEscrowAddress := c.localEscrowAccount.Address.Address()
+	remoteEscrowAddress := c.remoteEscrowAccount.Address.Address()
+
+	for _, op := range tx.Operations() {
```
leighmcculloch, line 123: I suspect we will struggle to cover all cases with this approach of inferring the new balance from the operations in the transaction. This function basically needs to become a stellar-core instance to analyze the tx and execute it.

Instead of inspecting the transaction and trying to figure out from the transaction what the balance of the account has become would it be possible to parse the result/meta XDR that's available for every tx to simply read out the new balance? The meta XDR should contain a before and after snapshot of any accounts and their trust lines that are affected.

I could be wrong about the meta XDR being a feasible approach, but it's what I'd explore and try out.

sdk/state/ingest.go
```diff
@@ -108,3 +115,37 @@ func (c *Channel) ingestTxToUpdateUnauthorizedCloseAgreement(tx *txnbuild.Transa
...
+		if paymentOp, ok := op.(*txnbuild.Payment); ok {
```
leighmcculloch, line 124: There are other operations that an credit an escrow account. Claim claimable balance, account merge, path payment strict receive, path payment strict send, maybe others. This is an example of why reading the current balance out of the meta XDR may be more resilient to all the variations.

sdk/state/ingest.go
```diff
@@ -108,3 +115,37 @@ func (c *Channel) ingestTxToUpdateUnauthorizedCloseAgreement(tx *txnbuild.Transa
...
+			// Check for a withdrawal. Check operation source first then tx source.
+			if paymentOp.SourceAccount == localEscrowAddress {
+				c.localEscrowAccount.Balance -= amount
+			} else if paymentOp.SourceAccount == remoteEscrowAddress {
+				c.remoteEscrowAccount.Balance -= amount
+			} else if tx.SourceAccount().AccountID == localEscrowAddress {
+				c.localEscrowAccount.Balance -= amount
+			} else if tx.SourceAccount().AccountID == remoteEscrowAddress {
+				c.remoteEscrowAccount.Balance -= amount
+			}
```
leighmcculloch, line 139: I think these conditions need to check that the tx source is the account and the op source is nil or equal too, or only the op source is equal. I think by checking each only individually there's an edge case being missed here where the escrow account is the tx source but the op source is some other account. I can't imagine what tx would do that, but we should probably avoid making that assumption in this code.

leighmcculloch (commented):
Left some comments inline.

sdk/state/ingest.go
```diff
@@ -14,6 +15,7 @@ import (
 // TODO: Accept the xdr.TransactionResult and xdr.TransactionMeta so code can
 // determine if successful or not, and understand changes in the ledger as a
 // result.
+// TODO: Make sure the ingested tx is fee bumped.
```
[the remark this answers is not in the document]
acharb: this was a reminder not necessarily for this PR but in general making sure the submitted transaction to the network is fee bumped, otherwise the balances will be off if the escrow account is going to pay or has paid a transaction fee.

Maybe the TODO should be moved somewhere else as it may not directly apply to `IngestTx` 🤔

sdk/state/ingest.go
acharb: I'll take a look at the meta XDR way. I thought about this and I was thinking the only way an escrow account balance might change is from a `Withdrawal` or `Deposit` defined in the SEP, which is a payment operation.

but if the meta XDR is a safer all encompassing way then makes sense to me

leighmcculloch: That would be true for withdrawal, but deposit can happen using any operation. The method of deposit isn't defined in the SEP and can take any form since the channel can't limit how the deposit occurs.

sdk/state/ingest.go
acharb: I think I have that test case covered [here](https://github.com/stellar/experimental-payment-channels/pull/234/files#diff-342d9dd35e25c88e1eb70e4d18a848b11d59461fc433e1e802686b6ba3cb73f7R256-R264). I tried covering all the edges cases in that test lmk if I'm missing one though.

My thinking here is first check if the op source is an escrow account. If op source is nil or some other account, then go to the next `else if` looking at the tx source

sdk/state/ingest.go
acharb: ah I took `"a standard payment operation"` literally as only a payment op, you're right it can be other types

acharb: playing around in lab it looks like the ResultMetaXdr _could_ work. Here are the results for a couple deposit types, where account `GD4...3II` increases in balance:
[Payment Op](https://laboratory.stellar.org/#xdr-viewer?input=AAAAAgAAAAIAAAADAA%2F9%2FQAAAAAAAAAAWPnYf%2B6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAXJLOWLwANrPwAAAAFAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABAA%2F9%2FQAAAAAAAAAAWPnYf%2B6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAXJLOWLwANrPwAAAAGAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABAAAABAAAAAMAD%2F18AAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABddU3scAA%2F8sgAAAAEAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAEAD%2F39AAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABdjSVwcAA%2F8sgAAAAEAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAMAD%2F39AAAAAAAAAABY%2Bdh%2F7qRA3e3ji%2BB6xB1aHgk44%2BPtp6vznYjtEmG3MAAAABcks5YvAA2s%2FAAAAAYAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAEAD%2F39AAAAAAAAAABY%2Bdh%2F7qRA3e3ji%2BB6xB1aHgk44%2BPtp6vznYjtEmG3MAAAABcevbUvAA2s%2FAAAAAYAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAA%3D&type=TransactionMeta&network=test)
[Path Payment Send](https://laboratory.stellar.org/#xdr-viewer?input=AAAAAgAAAAIAAAADAA%2F8wAAAAAAAAAAAWPnYf%2B6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAXSHbm1AANrPwAAAACAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAABAA%2F8wAAAAAAAAAAAWPnYf%2B6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAXSHbm1AANrPwAAAADAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAABAAAABAAAAAMAD%2FzAAAAAAAAAAABY%2Bdh%2F7qRA3e3ji%2BB6xB1aHgk44%2BPtp6vznYjtEmG3MAAAABdIdubUAA2s%2FAAAAAMAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAEAD%2FzAAAAAAAAAAABY%2Bdh%2F7qRA3e3ji%2BB6xB1aHgk44%2BPtp6vznYjtEmG3MAAAABc%2FhhVUAA2s%2FAAAAAMAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAMAD%2FyyAAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABdIdugAAA%2F8sgAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAEAD%2FzAAAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABdRZ7mAAA%2F8sgAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAA%3D&type=TransactionMeta&network=test)
[Claim Claimable Balance](https://laboratory.stellar.org/#xdr-viewer?input=AAAAAgAAAAIAAAADAA%2F9fAAAAAAAAAAA%2BPWlKVce34STyusA3ZTyYJANR5AIvOalC9bYovGEmc0AAAAXUWe5HAAP%2FLIAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAABAA%2F9fAAAAAAAAAAA%2BPWlKVce34STyusA3ZTyYJANR5AIvOalC9bYovGEmc0AAAAXUWe5HAAP%2FLIAAAABAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAABAAAABgAAAAMAD%2F1OAAAAAAAAAABY%2Bdh%2F7qRA3e3ji%2BB6xB1aHgk44%2BPtp6vznYjtEmG3MAAAABcks5aTAA2s%2FAAAAAUAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAEAD%2F18AAAAAAAAAABY%2Bdh%2F7qRA3e3ji%2BB6xB1aHgk44%2BPtp6vznYjtEmG3MAAAABcks5aTAA2s%2FAAAAAUAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAMAD%2F0SAAAABAAAAACOCqli4ggwTMj2RMhLLA9g5b6IK6dL2avxl7syBWvLfwAAAAEAAAAAAAAAAPj1pSlXHt%2BEk8rrAN2U8mCQDUeQCLzmpQvW2KLxhJnNAAAAAAAAAAAAAAAAC%2BvCAAAAAAAAAAABAAAAAQAAAABY%2Bdh%2F7qRA3e3ji%2BB6xB1aHgk44%2BPtp6vznYjtEmG3MAAAAAAAAAACAAAABAAAAACOCqli4ggwTMj2RMhLLA9g5b6IK6dL2avxl7syBWvLfwAAAAMAD%2F18AAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABdRZ7kcAA%2F8sgAAAAEAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAEAD%2F18AAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABddU3scAA%2F8sgAAAAEAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAA%3D&type=TransactionMeta&network=test)

the `ledger_entry_updated` shows the account's new balance. I'm not 100% on how this field is generated/how safe it is to rely on. I can ask the Horizon team.

Another thought I had was maybe the [ingest library](https://github.com/stellar/go/tree/master/ingest) could be a good fit here?

leighmcculloch: Horizon includes the meta XDR for every transaction in it's JSON response. I don't think the state package needs a dependency on the ingest package or on the horizon response, rather we can just pass in the xdr object. Where the xdr comes from can be up to the agent, and in the agent I'll likely use the horizon JSON response.

leighmcculloch: For example, the `result_meta_xdr` field: https://horizon.stellar.org/transactions/4185451ea7460167d9aca63bcf19710b5c683bb65cd9d4ee508ef02eeee74e89

acharb: ok cool, sgtm for using the `result_meta_xdr`!

sdk/state/ingest.go
acharb: resolving since this is changed now

sdk/state/ingest.go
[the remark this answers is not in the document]
leighmcculloch: Now that you're switching to the result meta xdr approach this shouldn't matter since a non-fee bumped tx that reduces the balance of the account should result in the new balance being less? I think anyway.

The txbuild package only builds txs with zero fees, so other code is required to fee bump the txs when submitting. I think validating that is probably out of scope, or not needed.

sdk/state/ingest.go
```diff
@@ -108,3 +112,33 @@ func (c *Channel) ingestTxToUpdateUnauthorizedCloseAgreement(tx *txnbuild.Transa
...
+
+// ingestTxMetaToUpdateBalances uses the transaction result meta data
+// from a transaction response to update local and remote escrow account
+// balances.
+func (c *Channel) ingestTxMetaToUpdateBalances(resultMetaXDR string) {
+	// If not a valid resultMetaXDR string, return.
+	var txMeta xdr.TransactionMeta
+	err := xdr.SafeUnmarshalBase64(resultMetaXDR, &txMeta)
+	if err != nil {
+		return
```
leighmcculloch, line 123: We should return the error if we can't decode. That seems like a problem to me, rather than something we'd ignore.

sdk/state/ingest.go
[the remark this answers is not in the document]
acharb: https://github.com/stellar/experimental-payment-channels/pull/234/commits/9d8a554876535a0aa13a2d99fc7f33692c096ba3

sdk/state/ingest.go
```diff
@@ -108,3 +112,33 @@ func (c *Channel) ingestTxToUpdateUnauthorizedCloseAgreement(tx *txnbuild.Transa
...
+			if !ok {
+				continue
+			}
+
+			switch updated.Data.Account.AccountId.Address() {
+			case c.localEscrowAccount.Address.Address():
+				c.UpdateLocalEscrowAccountBalance(int64(updated.Data.Account.Balance))
+			case c.remoteEscrowAccount.Address.Address():
+				c.UpdateRemoteEscrowAccountBalance(int64(updated.Data.Account.Balance))
+			}
```
leighmcculloch, line 140: Does this work for issued assets as well as native balance? We might need to look in a different place for issued assets. You could do that as a separate PR though if you wanted to, might not be trivial.

acharb:
@leighmcculloch changed to use the result meta xdr, let me know what you think. 

Specifically wondering about: 
1. this interface: `IngestTx(tx *txnbuild.Transaction, resultMetaXDR string) `

2. I'm not 100% on `ingestTxMetaToUpdateBalances` not throwing an error if the xdr is bad. Follows the pattern of the other ingest functions, but if the accounts were updated on the network but the xdr string is messed up for whatever reason we'd probably want to signal that with an error to the `agent` or whoever calls this

leighmcculloch:
> Specifically wondering about:
> 1. * this interface: `IngestTx(tx *txnbuild.Transaction, resultMetaXDR string) `

I think we can defer thinking about this too hard right now. We may prefer it to be an `xdr.TransactionMeta` but I don't really know yet, and the xdr string is fine and doesn't need to block merging this.

sdk/state/ingest.go
leighmcculloch: Respond to https://github.com/stellar/experimental-payment-channels/pull/234#issuecomment-902237782:

> 2\. I'm not 100% on `ingestTxMetaToUpdateBalances` not throwing an error if the xdr is bad. Follows the pattern of the other ingest functions, but if the accounts were updated on the network but the xdr string is messed up for whatever reason we'd probably want to signal that with an error to the `agent` or whoever calls this

I think in this case the error should be returned. We only ignore conditions that aren't errors, where it's information that is irrelevant. In this case we don't know if the information is irrelevant because we can't decode it.

sdk/state/ingest.go
acharb: ya  if this looks good I'll extend it to non-native assets, tracked here:
https://github.com/stellar/experimental-payment-channels/issues/237

sdk/state/ingest.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/234/commits/39ba4560aa27873a6fc6c14037dd5870d696039c

sdk/state/ingest_test.go
```diff
@@ -55,6 +55,9 @@ func TestChannel_IngestTx_latestUnauthorizedDeclTx(t *testing.T) {
 	responderChannel.UpdateLocalEscrowAccountBalance(100)
 	responderChannel.UpdateRemoteEscrowAccountBalance(100)
 
+	// To prevent xdr parsing error.
+	dummyXDR := "AAAAAgAAAAIAAAADABArWwAAAAAAAAAAWPnYf+6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAWg8TZOwANrPwAAAAKAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABABArWwAAAAAAAAAAWPnYf+6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAWg8TZOwANrPwAAAALAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABAAAABAAAAAMAD/39AAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABdjSVwcAA/8sgAAAAEAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAEAECtbAAAAAAAAAAD49aUpVx7fhJPK6wDdlPJgkA1HkAi85qUL1tii8YSZzQAAABee5CYcAA/8sgAAAAEAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAMAECtbAAAAAAAAAABY+dh/7qRA3e3ji+B6xB1aHgk44+Ptp6vznYjtEmG3MAAAABaDxNk7AA2s/AAAAAsAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAEAECtbAAAAAAAAAABY+dh/7qRA3e3ji+B6xB1aHgk44+Ptp6vznYjtEmG3MAAAABZIKg87AA2s/AAAAAsAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAA="
```
leighmcculloch, line 59: Another name instead of `dummy` can be `placeholder`, which doesn't carry some of the date offensive tones that can trigger some folks.

sdk/state/ingest_test.go
```diff
@@ -191,9 +200,88 @@ func TestChannel_IngestTx_oldDeclTx(t *testing.T) {
...
+	initiatorChannel.UpdateRemoteEscrowAccountBalance(10_000_0000000)
+
+	// Deposit, payment of 20 xlm to initiator escrow.
+	paymentResultMeta := "AAAAAgAAAAIAAAADABAqFAAAAAAAAAAAWPnYf+6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAXHr20ywANrPwAAAAGAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABABAqFAAAAAAAAAAAWPnYf+6kQN3t44vgesQdWh4JOOPj7aer852I7RJhtzAAAAAXHr20ywANrPwAAAAHAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAABAAAABAAAAAMAECn+AAAAAAAAAADp1ZmB/CD39aALHiR9KGnkQKe/9b7ukKKd6Uzf23JoWAAAABdIdugAABAp/gAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAEAECoUAAAAAAAAAADp1ZmB/CD39aALHiR9KGnkQKe/9b7ukKKd6Uzf23JoWAAAABdUYqoAABAp/gAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAMAECoUAAAAAAAAAABY+dh/7qRA3e3ji+B6xB1aHgk44+Ptp6vznYjtEmG3MAAAABcevbTLAA2s/AAAAAcAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAEAECoUAAAAAAAAAABY+dh/7qRA3e3ji+B6xB1aHgk44+Ptp6vznYjtEmG3MAAAABcS0fLLAA2s/AAAAAcAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAA="
```
leighmcculloch, line 240: How did you generate these?

leighmcculloch (approved):
One suggestion, one question. The question doesn't need to block merging. Otherwise looks great. I'm surprised with how simple the usage of the xdr ended up being. TIL!

sdk/state/ingest_test.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/234/commits/b7229b4caa4252b5d1124ea5f22a3b2b9d3907d9  😄

sdk/state/ingest_test.go
acharb: from lab. And then if the balance was slightly off because of tx fees I converted to object, changed the balance, and then converted back to xdr to mimic as if it was fee bumped

leighmcculloch: It might be worth adding to their comments the version of the protocol they were generated with, in this case v17. So in the future if we need to update them we know what might be different.

acharb: good idea https://github.com/stellar/experimental-payment-channels/pull/234/commits/0be276227600a502ce1ccccfd420fe1962584ca1

PR #236, opened 2021-08-17
all: update for go 1.17

leighmcculloch:


sdk/go.mod
```diff
@@ -11,3 +11,18 @@ require (
 	github.com/stretchr/testify v1.7.0
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+require (
```
acharb, line 15: ❓ are these supposed to be added here?

examples/console/go.mod
```diff
@@ -11,3 +11,22 @@ require (
 	github.com/stellar/go v0.0.0-00010101000000-000000000000
 	gopkg.in/yaml.v2 v2.3.0 // indirect
 )
+
+require (
```
acharb, line 15: same question here. These didn't show up for me locally when I upgraded to 1.17

leighmcculloch: They won't show up until the `go` directive at the top of the file is changed to `1.17`. `go mod tidy` respects the go directive to some degree, so even if you're using go 1.17, if the file says 1.16, it uses the logic from go 1.16 for how to generate the file, but nothing else.

sdk/go.mod
leighmcculloch: Go 1.17 changed the layout of the file a bit. I don't remember the details.

acharb (approved):
lgtm!

Issue #237, opened 2021-08-19
sdk/state: extend ingesting tx to update balances to non-native assets


acharb:
builds on https://github.com/stellar/experimental-payment-channels/pull/234 which is only updating for native assets now. So extend to non-native as well next

PR #238, opened 2021-08-20
sdk/state: look for and update non-native balances for the escrow accounts

acharb:
**WHAT**
look for non-native asset balance changes when parsing through the resultMetaXDR. If a change is found update the channel escrow account balance.

**WHY**
channel supports both native and non-native assets so need to account for both.

closes https://github.com/stellar/experimental-payment-channels/issues/237


sdk/state/ingest.go
```diff
@@ -126,20 +126,44 @@ func (c *Channel) ingestTxMetaToUpdateBalances(resultMetaXDR string) error {
 		return fmt.Errorf("parsing the result meta xdr: %w", err)
 	}
 
-	// Find leder changes for the local and remote escrow accounts, if any,
-	// to update their balance.
+	channelAsset := c.openAgreement.Details.Asset
+
+	// Find ledger changes for the escrow account channel asset balances,
+	// if any, and then update.
```
leighmcculloch, line 132:
```suggestion
	// Find ledger changes for the escrow accounts' balances,
	// if any, and then update.
```

leighmcculloch (approved):
Looks superb 👏🏻.

PR #240, opened 2021-08-21
sep: change title

leighmcculloch:
### What
Change the title of the SEP to refer to the mechanics of the payment channel rather than transactions.

### Why
The SEP contains a specification of the transactions and the basic mechanics required to form and maintain a payment channel safely, so it defines more than just the transactions. For this reason I wanted to drop the `transactions` from the title. However, the SEP doesn't define much about how coordination should occur and so it is not a sufficient SEP in its current form to be considered a specification of the entire payment channel. The scope of the SEP is really the bare minimum mechanics which can sit under a variety of applications.


acharb (approved):
lgtm!

PR #244, opened 2021-08-23
sdk/agent: move logic for determining initiator to a func

leighmcculloch:
### What
Move logic for determining initiator to a function.

### Why
It is isolated logic and it is helpful to give it a name.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -250,6 +250,10 @@ var handlerMap = map[msg.Type]func(*Agent, msg.Message, *msg.Encoder) error{
 	msg.TypeCloseResponse:   (*Agent).handleCloseResponse,
 }
 
+func isInitiator(self, other *keypair.FromAddress) bool {
+	return self.Address() > other.Address()
+}
+
 func (a *Agent) handleHello(m msg.Message, send *msg.Encoder) error {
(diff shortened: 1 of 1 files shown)


sdk/agent/agent.go
```diff
@@ -250,6 +250,10 @@ var handlerMap = map[msg.Type]func(*Agent, msg.Message, *msg.Encoder) error{
 	msg.TypeCloseResponse:   (*Agent).handleCloseResponse,
 }
 
+func isInitiator(self, other *keypair.FromAddress) bool {
+	return self.Address() > other.Address()
```
acharb, line 254: ❓ is this just saying whichever address is higher in the lexical order is the initiator? Is that just a simplification for now?

acharb (commented):
just had 1 ❓

sdk/agent/agent.go
leighmcculloch: > is this just saying whichever address is higher in the lexical order is the initiator?

Yup.

> Is that just a simplification for now?

Yes, but it might work long term. It doesn't make any difference to the participants who the initiator is. We need to pick one to be and this seems simple.

acharb (approved):
lgtm!

PR #245, opened 2021-08-23
sdk/agent: add todo to remove logic

leighmcculloch:
### What
Add todo to remove some logic that won't be needed once ingestion occurs.

### Why
I stumbled on the code when working on something else and realized we will need to rewrite it to add ingestion in the agent and don't want to forget.


acharb (approved):
lgtm

PR #246, opened 2021-08-24
sdk/agent: move channel creation into open process

leighmcculloch:
### What
Move the channel creation in the agent into the open process instead of happening as part of the hello process.

### Why
The agent creates the channel the moment the agent has the hello information from the other participant, but this is somewhat unnecessary, introduces stale data, and complicates the channel setup.

It is unnecessary because the agent doesn't need the channel until the open process begins.

Stale data may be introduced because at the moment the sequence number of the escrow accounts is needed to be known when initializing the channel, however it could have changed by the time the open starts and when the open starts is when it is required.

The channel setup is complicated when the channel is created early because channel creation records who the initiator is. This has required us to come up with a formula to choose the initiator and who the initiator is must be detached from who proposes the open. We can simply make the initiator the person proposing the open if the initialization of the channel occurs during the open. You could argue that we should set the initiator during the first call to Open in the Channel type, but that has a side-effect that the initiator isn't a constant value for the lifetime of the Channel value and the type is simpler if we can say many of the config fields are constants.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -292,11 +292,11 @@ func (a *Agent) handleHello(m msg.Message, send *msg.Encoder) error {
 
 	h := m.Hello
 
-	fmt.Fprintf(a.LogWriter, "other's escrow account: %v\n", h.EscrowAccount.Address())
 	a.otherEscrowAccount = &h.EscrowAccount
-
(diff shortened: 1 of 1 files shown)


acharb (approved):
lgtm! 💯

PR #247, opened 2021-08-24
(wip) ingest formation tx and add OpenState

acharb:


sdk/state/ingest.go
```diff
@@ -30,6 +30,105 @@ func (c *Channel) IngestTx(tx *txnbuild.Transaction, resultMetaXDR string) error
...
+
+			if address == c.initiatorSigner().Address() && signer.Weight == xdr.Uint32(1) {
+				initiatorSignerFound = true
+			} else if address == c.responderSigner().Address() && signer.Weight == xdr.Uint32(1) {
+				responderSignerFound = true
+			}
+		}
+		if !initiatorSignerFound || !responderSignerFound || len(ea.Signers) != 2 {
+			return fmt.Errorf("incorrect signer weights found")
+		}
```
leighmcculloch, line 113: I know this is a WIP PR and therefore not ready for review, but I took a quick peak and Wanted to point out that it is important that there are no other signers on the account. If a SetOptions adds a new signer I don't know if all signers that were already on the account are included on that operations meta changes. Hopefully they are, but I wanted to mention that.

acharb: i'll add a test case for that 👍

sdk/state/ingest.go
```diff
@@ -169,3 +169,116 @@ func (c *Channel) ingestTxMetaToUpdateBalances(resultMetaXDR string) error {
...
+		if !initiatorSignerFound || !responderSignerFound {
+			return fmt.Errorf("incorrect signer weights found")
+		}
+		if len(ea.Signers) != requiredNumOfSigners {
+			return fmt.Errorf("incorrect number of signers, found: %d want: %d", len(ea.Signers), requiredNumOfSigners)
+		}
+	}
+
+	// Validate the trustlines are correct.
+	// TODO - should an initiator escrow be allowed to have extraneous trustlines (if not add test)?
```
acharb, line 263: @leighmcculloch wdyt about a native asset channel EI having existing trustlines? I can't think of an explicit reason to now allow it, but might not be thinking of something

leighmcculloch: There's two routes we can go:
1. Verify that an account is in a very specific state, only has the trustline required, has specific flags, etc.
2. Or, verify that only the things we need are set the way we need.

I think we should do the latter because the first approach will always be a moving target. For example, when accounts get liquidity pools, do we check for those too? I don't think we want that sort of burden if we can avoid it.

I think we should do our best to ignore anything that has no bearing on the security of the channel. This means:
- Verifying that there are only the signers we specify, this is one thing we must check full state on.
- Verifying that the trust line for the channel's asset is has an appropriate balance that other than auth revocation is definitely available to supply the closing transaction's payment. That includes taking into account trust line liabilities, so I've opened a new issue #251 to cover adding liabilities.
- Verifying that the trust line for the channel's asset is authorized. \*
- Ignoring everything else.

\* We could ignore the authorization status of the trust line if we hope to support [SEP-8](https://stellar.org/protocol/sep-8) in the manner described in https://github.com/stellar/experimental-payment-channels/discussions/34#discussioncomment-765861. We aren't implementing support for SEP-8 right now so I think it makes sense to check authorization in our current implementation even if the authorization can be revoked anytime, and we can remove it if we go down a path of implementing SEP-8.

PR #249, opened 2021-08-24
sep: fix formating and heading

leighmcculloch:
### What
Update the wrapping of a paragraph and change the heading leveling of one section.

### Why
The paragraph wasn't word wrapped consistently with the rest of the document.

The section heading had incorrect leveling. It is supposed to be a sub-section of the security section, but it is top-level.


acharb (approved):
lgtm!

PR #250, opened 2021-08-24
sep: add discussion about risks with queueing payments

leighmcculloch:
### What
Add discussion about the risks of queueing multiple payments by sending multiple payments to the other participant before they confirm each individually.

### Why
CAP-40 gives us a bonus benefit that participants can make multiple payments without waiting for the confirmation from teh receiver. Some future version of the protocol defined in the SEP could use that bonus and make it possible for participants to speed up and slow down independent of each other and be allowed to fall behind for a period or indefinitely. There is one side-effect and tradeoff to that though which is important to weigh with any implementation that uses it. I wanted to document it so that we didn't forget about it if we went all in on that approach in the future.

### DIFF
--- a/specifications/sep-payment-channel-mechanics.md
+++ b/specifications/sep-payment-channel-mechanics.md
@@ -691,7 +691,8 @@ remaining transactions.
 
 This protocol implicitly supports participants sending multiple payments to the
 other participant without the other participant confirming each payment
-immediately. This introduces some risk to the sending participant because the
+immediately. This introduces some risk to the sending participant if they
+queue multiple payments to the receiver without any replies, because the
 receiving participant can prevent the channel from being closed for
 approximately the observation period multiplied by the number of uncomfirmed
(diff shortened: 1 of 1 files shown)


specifications/sep-payment-channel-mechanics.md
```diff
@@ -687,6 +687,19 @@ could submit a subset of the transactions required by the process and the
 participant will not have any other capability to authorize and submit the
 remaining transactions.
 
+### Queueing Multiple Payments
+
+This protocol implicitly supports participants sending multiple payments to the
+other participant without the other participant confirming each payment
+immediately. This introduces some risk to the sending participant because the
```
acharb, line 694: 💭  explicitly saying  in the description the risk only happens if the sender queues payments
```suggestion
immediately. This introduces some risk to the sending participant if they
queue multiple payments to the receiver without any replies, because the
```

acharb (approved):
lgtm, just 1 💭

PR #257, opened 2021-08-26
sdk/agent: implement event handling with channels

### LINKED ISSUE #226: sdk/agent: add event handlers for open, payment, close
The agent doesn't expose anyway for an application to track changes in state or payments received. As was specified in #217 we need to add event handlers for anything that can occur. At the very least we need the below functions, but we likely need others.

```
func (a *Agent) RegisterErrorHandler(func(error))
func (a *Agent) RegisterOpenHandler(func())
func (a *Agent) RegisterPaymentHandler(amount string)
func (a *Agent) RegisterCloseHandler(finalAmount string)
```

leighmcculloch:
### What
Add event handlers for the things that can occur in the background during open, payment, and close.

### Why
The agent receives messages from the other participant containing confirmations/responses to requests made by this recipient, and requests from the other participant. Some of these operations that take place in the background need to be made known to the application using the agent.

The agent will also monitor the state of the network for changes that require action, and once that logic is added we'll need to add events using the same pattern.

A channel is used for delivering these background events because it makes it easier for developers who don't want to handle concurrent routines in their code. They can simply use a buffered channel and poll when they want to. Or they can put it in a go routine and trigger any functions they have for each.

Supersedes #242 
Close #226

### Known Limitations
This change is somewhat limited because some of the change depends on ingestion having been integrated into the agent, but I'm deferring doing that integration until the integration and state changes have been finished in the state package. Because of that there are some TODOs, but I think it is worth merging this and revisiting it during the ingestion work.


examples/console/main.go
```diff
@@ -103,6 +103,27 @@ func run() error {
 		FeeAccount:        accountKey,
 		FeeAccountSigners: []*keypair.Full{signerKey},
 	}
+	events := make(chan agent.Event)
```
acharb, line 106: 💡   thoughts on adding a "poll" command to the example similar to as you described in the description:
> They can simply use a buffered channel and poll when they want to.

I think logging doesn't show the  power of this events channel. But being able to poll on command the events could simulate what a user might do

wouldn't be in this PR though, just an idea that popped into my head

sdk/agent/events.go
```diff
@@ -0,0 +1,52 @@
...
+// PaymentReceivedAndConfirmedEvent occurs when a payment is received and has
+// been confirmed as valid.
+type PaymentReceivedAndConfirmedEvent struct {
+	CloseAgreement state.CloseAgreement
+}
+
+func (e PaymentReceivedAndConfirmedEvent) event() {}
+
+// PaymentSentAndConfirmedEvent occurs when a payment is sent and the other
+// participant has confirmed the payment.
```
acharb, line 36: 💭  i"m not necessarily tied to this suggestion. But when I first read the comment and variable name I thought the event might be triggered when the other party confirmed not when we confirmed the confirm. Feel free to ignore if it makes it more confusing
```suggestion
// PaymentSentAndConfirmedEvent occurs when a payment is sent, the other
// participant confirmed, and we confirmed.
```

sdk/agent/events.go
```diff
@@ -0,0 +1,52 @@
...
+// ErrorEvent occurs when an error has occurred, and contains the error
+// occurred.
+type ErrorEvent struct {
+	Err error
+}
+
+func (e ErrorEvent) event() {}
+
+// ConnectedEvent occurs when the agent is connected to another participant.
+type ConnectedEvent struct{}
```
acharb, line 18: ❓  should we add a `DisconnectedEvent`? I don't _think_ this is covered by the ErrorEvent currently, but if it is maybe we should make it its own event type

acharb (commented):
lgtm 👍  had a question and some thoughts, non blocking but wanted to see your thoughts before approving

sdk/agent/events.go
leighmcculloch: Probably, but nothing triggers it at the moment. There are a few edge conditions that aren't covered in the agent yet that need plugging. I created an issue for this at #264.

sdk/agent/events.go
leighmcculloch: > when the other party confirmed not when we confirmed the confirm

I'm not sure the distinction is super important from the caller of the function as that detail of how the confirmations occur exactly is abstracted by the agent. In fact, that would suggest we could rename these events as `PaymentReceived` and `PaymentSent`.

Wdyt?

leighmcculloch: 4aae901

examples/console/main.go
acharb: going to approve since this would be in a separate PR if we decided to do

acharb (approved):
lgtm!

examples/console/main.go
leighmcculloch: We could have some example tests for sure that demonstrate different methods of use.

PR #260, opened 2021-08-27
sdk/state: ingest from raw tx xdr and support fee bump txs

### LINKED ISSUE #253: sdk/state: ingest tx accept fee bump or non-fee bump txs
The `IngestTx` currently accepts only a non-fee bumped tx. This isn't really a problem since for a fee bumped tx it can be unwrapped and the inner tx can be passed in. However, I think it would be better if the ingest tx function did that itself since the result meta will be for the fee bump, so it would be clearer and more consistent if the `IngestTx` func accepted a `*txnbuild.GenericTransaction` and unwrapped it if necessary.

leighmcculloch:
### What
Ingest raw tx XDR instead of objects into the state ingest functions, and add support for fee bump txs as input.

### Why
When I started hooking the state machine's `IngestTx` function into the agent I discovered I was going to have to do this somewhere, and it seems like the most natural thing is to make the ingestion functions handle fee bumps. The easiest way to accept either a fee bump or non-fee bump tx is to accept the XDR and let the function figure it out. XDR is easier than using an object because the `txnbuild.GenericTransaction` type that wraps fee bumps and normal transactions is not constructible from outside the txnbuild package which means we'd have to convert to XDR first then create a `txnbuild.GenericTransaction` with that XDR, then pass it in. We may as well do that inside the `IngestTx`. This isn't a huge shift in philosophy because the `IngestTx` function already accepts other information about the transaction as XDR, so if anything it makes it more consistent.

Close #253


acharb (approved):
lgtm!

PR #262, opened 2021-08-27
sdk/agent: address data races

### LINKED ISSUE #256: sdk/agent: address concurrent use of channel
The state.Channel does not support concurrent use of itself. This is intentional as it has no internal need to do so since its logic all runs on the routine of the caller.

The agent.Agent uses the channel in at least two routines at the same time, and so there exists race conditions that do not currently exist. The agent was ported from the console example and implementing a perfect use of the channel wasn't necessary in that example, but as part of the SDK we must address this.

There are a few approaches we can take and we need to consider what makes the most sense in the context of all the values that the agent might need to access concurrently, as it may not be limited to its `channel state.Channel`.

For example, the agent has a `conn io.ReadWriter` which in most uses will probably be a `net.Conn`. `net.Conn` is safe to use concurrently and so on the surface we don't need a strategy for using it concurrently. However, `net.Conn` implements its `Write` calls by locking a mutex internally such that writes are never interleaved (see https://github.com/golang/go/issues/27203#issuecomment-534386117) and that must inform how we handle the concurrent access of other elements within the agent. The reason this is important is because we could introduce deadlock if two routines attempt to lock the agent and the net.Conn in reverse order.

This problem also needs to consider, or affect, how we implement notifications/event handling.

leighmcculloch:
### What
Add a mutex to the `agent.Agent` that serializes any reads and writes to state that is not immutable, namely the internal `state.Channel` and the other variables that record the other participants escrow details.

### Why
There is a data race in the agent because the agent runs a go routine in the background that receives messages from the other participant. It is possible for a message to arrive inbound while a payment is being proposed or other operation is being started.

I chose a mutex over using channels to interact with the `state.Channel` object because it seems like the more suitable tool. We could use a channel to pass commands to a background routine that operated on the channel, or we could even pass functions to a background routine to serialize them. Both of these alternative approaches using channels would be more complicated because data would also need to be passed back and so the sending routine needs to block on a returning value. Because of this I think a mutex is simpler and easier to understand the implications.

Not all functions in the agent need to be serialized. Some functions do not utilize any of the mutable fields of the agent.

Functions that access only the `conn io.ReadWriter` may not necessarily need locking because that field will typically be a `net.Conn` and `net.Conn`s are safe to use concurrently because their reads are concurrent and their writes use an internal lock. However, I have chosen to treat any function using the `conn io.ReadWriter` as requiring a lock since there is no guarantee the value is a `net.Conn` since the type of the field is `io.ReadWriter`. I don't think this will have any material impact.

Close #256


sdk/agent/agent_test.go
```diff
@@ -234,3 +234,139 @@ func TestAgent_openPaymentClose(t *testing.T) {
...
+	<-remoteOpened
+
+	// Make a payment.
+	err = localAgent.Payment("50.0")
+	require.NoError(t, err)
+	err = remoteAgent.Payment("50.0")
+	require.NoError(t, err)
+
+	<-localPaymentConfirmedOrError
+	<-remotePaymentConfirmedOrError
```
acharb, line 371: 💭  should we include the Close steps in this test as well? Since there's a possible data race there as well?

sdk/agent/agent.go
```diff
@@ -52,21 +53,23 @@ type Agent struct {
 
 	Events chan<- Event
 
+	// mu is a lock for the mutable fields of this type. It should be locked
+	// when reading or writing any of the mutable fields. The mutable fields are
+	// listed below. If pushing to a chan, such as Events, it is unnecessary to
+	// lock.
+	mu sync.Mutex
```
acharb, line 60: ❓ more for my own understanding but was putting the mutex lock on the channel struct considered? Since the channel is closer to the data that has the data race. Or is it bc we needed to lock the `agent.conn` as well

reason I'm wondering is because if it was on the channel then we could put the responsibility of locking in the sdk eg. on the `UpdateLocalEscrowAccountBalance` and `UpdateRemoteEscrowAccountBalance` methods

acharb (commented):
overall looks great 👍  TIL of the `-race` flag. Had 1 💭  and an overall ❓ .

sdk/agent/agent.go
leighmcculloch: We could, there were a couple reasons I went with in the agent though.

I'm not 100% sure if it would actually be a problem, but we need to be careful to avoid deadlock whenever there are multiple synchronizing primitives in use. The `conn`, if it's a `net.Conn`, has an internal lock, and the state.Channel will have a lock, and if both are locked in alternating orders we could end up with deadlock. It's a little safer to nest the mutex around all the entities we need to synchronize on.

Another reason is that the state.Channel has no use for a mutex itself. It doesn't run any go routines. So it seems to me like the requirements of the agent would be leaking into the channel. This would mean any use of the `state.Channel` without the agent would be bound to use the synchronizing pattern that we added for the agent, so it would couple the two together a little.

Let me know if you see tradeoffs I'm not seeing.

acharb: nope I was more just curious than anything. Both those reasons make sense, thx for explaining!

sdk/agent/agent_test.go
leighmcculloch: Good idea. f86f328. Note that there's no closing event being triggered yet even though the state exists for it. That'll become available in #265.

leighmcculloch: I just removed this change. I remember now why I hadn't included it. I'm moving the channel into an odd state because both participants have proposed conflicting payments. This happens so that the frontend and backend of the channel are working at the same time, but then it makes the test pretty complex from then on because the test has to handle which one got there first. I don't think it's worth the extra complexity to try to exercise the close functions. Testing concurrency is imperfect anyway.

acharb (approved):
lgtm! 💯

PR #266, opened 2021-08-30
sdk/state: remove duplicate updating sequence number

leighmcculloch:
### What
Remove duplicate setting the cache of initiator escrow account's sequence number during ingestion.

### Why
There was a TODO comment saying that we needed to combine the sequence number being updated in two locations. It looks like the other location is sufficient, so I thought I'd delete the TODO and the line of code. The change required updating some tests that called directly into the `ingestFormationTx` since the formation tx ingestion now relies on logic outside that function that is called in the `IngestTx` function. The change then also required rewriting the "not the formationTx, should return with no state change" test to make it a valid tx that still wouldn't trigger the state change since the tx that was being used was a declaration tx and would transition the channel directly to closing.


acharb (approved):
lgtm!

Issue #261, opened 2021-08-27
sdk/state: change open/confirm/payment validations to check against ChannelState() when possible


acharb:
right now validations for open/confirm/payment are checking against the state of the latest agreements. Now that we have an overall ChannelState() enum, we should instead (or possibly also) check against that value

acharb:
adding next steps for when I or someone jumps back on this.

We'll need to add more granular states for the open steps. @leighmcculloch 's suggestion:
https://github.com/stellar/experimental-payment-channels/pull/270#discussion_r698941132 .

Probably 2 PRs to complete:
1. add the missing states, and assert them in the tests
2. swap out the validations to use these new channel states

PR #270, opened 2021-08-30
sdk/state: refactor validations to check against the channel state

acharb:
**WHAT**
changes the validations for checking if the channel is open to use the Channel State instead of checking the openAgreement is empty or not

**WHY**
the channel state better reflects if the channel is actually open based off of ingested transactions. And more accurately gives the state of the channel.

closes https://github.com/stellar/experimental-payment-channels/issues/261


sdk/state/payment.go
```diff
@@ -158,14 +162,18 @@ var ErrUnderfunded = fmt.Errorf("account is underfunded to make payment")
...
+	}
+	if cs < StateOpen {
 		return fmt.Errorf("cannot confirm a payment before channel is opened")
 	}
 
 	// If a coordinated close has been proposed by this channel already, error.
 	if !c.latestUnauthorizedCloseAgreement.isEmpty() && c.latestUnauthorizedCloseAgreement.Details.ObservationPeriodTime == 0 &&
 		c.latestUnauthorizedCloseAgreement.Details.ObservationPeriodLedgerGap == 0 {
-		return fmt.Errorf("cannot propose payment after proposing a coordinated close")
+		return fmt.Errorf("cannot confirm payment after proposing a coordinated close")
```
acharb, line 176: this should have been `confirm` before

sdk/txbuildtest/txbuildtest.go
```diff
@@ -57,3 +58,78 @@ func BuildResultMetaXDR(ledgerEntryResults []xdr.LedgerEntryData) (string, error
...
+type FormationResultMetaParams struct {
+	InitiatorSigner string
+	ResponderSigner string
+	InitiatorEscrow string
+	ResponderEscrow string
+	StartSequence   int64
+	Asset           txnbuild.Asset
+}
+
+func BuildFormationResultMetaXDR(params FormationResultMetaParams) (string, error) {
```
acharb, line 71: added this helper method. When generating the resultMetaXDR for the integration tests this chunk of code for the formationTx Result XDR was written enough and took up annoyingly enough space that it was worth abstracting away

sdk/state/open.go
```diff
@@ -224,7 +229,11 @@ func (c *Channel) ProposeOpen(p OpenParams) (OpenAgreement, error) {
 
 func (c *Channel) validateOpen(m OpenAgreement) error {
 	// if the channel is already open, error.
-	if c.openAgreement.isFull() {
+	cs, err := c.State()
+	if err != nil {
+		return fmt.Errorf("getting channel state: %w", err)
+	}
+	if cs >= StateOpen {
```
acharb, line 236: I think this is better than `if cs != StateNone`, in the case the channel errored on the first attempt to open, so they should be allowed to try again

sdk/state/payment.go
```diff
@@ -96,7 +96,11 @@ func (c *Channel) ProposePayment(amount int64) (CloseAgreement, error) {
 	}
 
 	// If the channel is not open yet, error.
-	if c.latestAuthorizedCloseAgreement.isEmpty() {
+	cs, err := c.State()
+	if err != nil {
+		return CloseAgreement{}, fmt.Errorf("getting channel state: %w", err)
+	}
+	if cs < StateOpen {
```
acharb, line 103: wondering if `if cs != StateOpen` is better here though, @leighmcculloch wdyt?

sdk/state/payment_test.go
```diff
@@ -216,7 +218,7 @@ func TestChannel_ConfirmPayment_acceptsSameObservationPeriod(t *testing.T) {
 		SequenceNumber: int64(202),
 	}
```
acharb, line 220: deceiving amount of changes in this file. Replacing putting the channel into an Open State by changing internal data by doing so using external methods instead. Which should prevent having to make these changes more often in the future

sdk/state/payment.go
```diff
@@ -96,7 +96,11 @@ func (c *Channel) ProposePayment(amount int64) (CloseAgreement, error) {
 	}
 
 	// If the channel is not open yet, error.
-	if c.latestAuthorizedCloseAgreement.isEmpty() {
+	cs, err := c.State()
+	if err != nil {
+		return CloseAgreement{}, fmt.Errorf("getting channel state: %w", err)
+	}
+	if cs != StateOpen {
```
leighmcculloch, line 103: 👍🏻 I think this change to using state is good.

sdk/state/payment.go
```diff
@@ -158,14 +162,18 @@ var ErrUnderfunded = fmt.Errorf("account is underfunded to make payment")
 // on the state of the close agreement signatures.
 func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 	// If the channel is not open yet, error.
-	if c.latestAuthorizedCloseAgreement.isEmpty() {
+	cs, err := c.State()
+	if err != nil {
+		return fmt.Errorf("getting channel state: %w", err)
+	}
+	if cs < StateOpen {
```
leighmcculloch, line 169: 👍🏻 I think using the state here is good, however I think the condition needs to be `cs != StateOpen`, because any non-open state should not be supported here.

sdk/state/payment_test.go
```diff
@@ -467,6 +583,35 @@ func TestChannel_ConfirmPayment_initiatorRejectsPaymentThatIsUnderfunded(t *test
...
+		require.NoError(t, err)
+
+		successResultXDR, err := txbuildtest.BuildResultXDR(true)
+		require.NoError(t, err)
+		resultMetaXDR, err := txbuildtest.BuildFormationResultMetaXDR(txbuildtest.FormationResultMetaParams{
+			InitiatorSigner: localSigner.Address(),
+			ResponderSigner: remoteSigner.Address(),
+			InitiatorEscrow: localEscrowAccount.Address.Address(),
+			ResponderEscrow: remoteEscrowAccount.Address.Address(),
+			StartSequence:   localEscrowAccount.SequenceNumber + 1,
```
leighmcculloch, line 606: 💡
```suggestion
			StartSequence:   ftx.SequenceNumber(),
```

sdk/state/close_test.go
```diff
@@ -107,16 +109,41 @@ func TestChannel_ProposeClose(t *testing.T) {
...
+			StartSequence:   localEscrowAccount.SequenceNumber + 1,
+			Asset:           txnbuild.NativeAsset{},
+		})
+		require.NoError(t, err)
+
+		err = localChannel.IngestTx(ftxXDR, successResultXDR, resultMetaXDR)
+		require.NoError(t, err)
+		err = remoteChannel.IngestTx(ftxXDR, successResultXDR, resultMetaXDR)
+		require.NoError(t, err)
+	}
```
leighmcculloch, line 146: :bulb: The improvements to the tests look great. The fact that so many places in the tests needed updating for the state to be accurate, I think it would be worth adding into every test that was changed assertions on the result of calling `channel.State()`, that way these tests will ensure that the state value returned from the function is consistent over time. Wdyt?

leighmcculloch (commented):
I left some ideas (💡), concerns (🤔), notes about things that need addressing (❗). Because I raised some concerns about using the state check in some places, I explicitly marked the places where the state looks good (👍🏻).

The improvements to the tests look great. 🎉

sdk/state/close.go
```diff
@@ -93,7 +93,11 @@ func (c *Channel) ProposeClose() (CloseAgreement, error) {
 	}
 
 	// If the channel is not open yet, error.
-	if c.latestAuthorizedCloseAgreement.isEmpty() {
+	cs, err := c.State()
+	if err != nil {
+		return CloseAgreement{}, fmt.Errorf("getting channel state: %w", err)
+	}
+	if cs < StateOpen {
 		return CloseAgreement{}, fmt.Errorf("cannot propose a coordinated close before channel is opened")
```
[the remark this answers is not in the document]
acharb: 🤔   in my head I was thinking that, by checking the channel is in the correct state for proposing/confirming a close, then we're checking the prerequisites for doing the action are completed. So in this case checking the state means we're also checking the channel went through the open steps, `latestAuthorizedAgreement` looks good, and sequence number is in the right place. I liked that better than explicitly checking the data for these propose/confirm methods.

I guess it's the difference of checking the data explicitly vs an abstracted method here, which is more disconnected 🤔 . wdyt, I might not be thinking of something so if you think the former is better I can change back here

acharb:
note: moved the test improvements to this PR: https://github.com/stellar/experimental-payment-channels/pull/277

to use the channel.State for validations it looks like we'll need to add more state values, to ensure we're validating at a granular enough level. Closing this PR for now as that will take some additional thought and possible refactoring

Issue #269, opened 2021-08-30
sdk/txbuildtest: create test helper funcs for creating result and meta xdr


leighmcculloch:
We're starting to write more and more tests that require writing valid result xdr and result meta xdr.

For the few tests we've written so far using XDR that we've manually generated and inserted into tests has worked pretty well. For example:

https://github.com/stellar/experimental-payment-channels/blob/48ed2c39c73ec751ef38a15109d653a40bd2eedf/sdk/state/ingest_test.go#L255

Generating these XDRs manually has helped us gain some confidence on our code's ability to parse real result meta, but it isn't going to scale if we need to generate it for every test we write.

Any tests we write in the `sdk/agent` need the result and meta XDR because the agent only progresses state by circulating submitting transactions back into the state machines ingestion functionality. Therefore any test we write needs to generate these files unless we write tests that fiddle with the internal state of the channel.

So that we can continue to write outside-in tests as much as possible I think we should write some test helper functions that take a built transactions and return a result XDR or result meta XDR for different conditions. The XDR they generate doesn't need to be perfect, but just sufficient. We will write integration tests (#268) separately to the majority of our tests that will validate real XDR in the main use cases.

I'm thinking something like:

```go
package txbuildtest

// BuildResult returns a result XDR base64 encoded that is successful or not based on the input parameter.
func BuildResult(success bool) string

// BuildResultMeta returns a result meta XDR base64 encoded that contains arbitrary operations that result in the given account and trustline state.
func BuildResultMeta(txChangesAfter xdr.LedgerEntryChanges) string
```

Maybe we could make this even easier, such as `txbuildtest.BuildResultMeta(txXDR)`, but I'm not sure that is possible.

PR #271, opened 2021-08-31
sdk/state: add helper methods for building xdr strings in tests

acharb:
**WHAT**
adds helper methods to be used in tests for creating desired transaction result xdr and transaction meta xdr.

**WHY**
the current xdr strings used in the tests work but are difficult to adjust when data needs to be changed

closes https://github.com/stellar/experimental-payment-channels/issues/269

note: I didn't change all the current xdr strings in the tests to use these new helper methods. I changed one to show how the helper methods could be used, and the rest can be changed on a need basis.


sdk/state/ingest_test.go
```diff
@@ -820,8 +821,44 @@ func TestChannel_IngestTx_updateState_invalid_initiatorEscrowHasExtraSigner(t *t
...
+			Updated: &xdr.LedgerEntry{
+				Data: xdr.LedgerEntryData{
+					Type: xdr.LedgerEntryTypeAccount,
+					Account: &xdr.AccountEntry{
+						AccountId: xdr.MustAddress(responderEscrow.Address()),
+					},
+				},
+			},
+		},
+	})
```
leighmcculloch, line 860: 💡 I think we might be able to get away with only supporting `...interface{}`, and the function can assume it's the after state we're wanting to set and pass the value into `xdr.NewLedgerEntryData`. It would reduce the test code down to:
```suggestion
	resultMetaXDR, err := txbuildtest.BuildResultMetaXDR(
		&xdr.AccountEntry{
			AccountId: xdr.MustAddress(initiatorEscrow.Address()),
			SeqNum:    102,
			Signers: []xdr.Signer{
				{ Key: xdr.MustSigner("GAKDNXUGEIRGESAXOPUHU4GOWLVYGQFJVHQOGFXKBXDGZ7AKMPPSDDPV"), Weight: 1 },
				},
				Thresholds: xdr.Thresholds{0, 2, 2, 2},
			},
		},
		&xdr.AccountEntry{
			AccountId: xdr.MustAddress(responderEscrow.Address()),
		},
	})
```

sdk/state/ingest_test.go
```diff
@@ -820,8 +821,44 @@ func TestChannel_IngestTx_updateState_invalid_initiatorEscrowHasExtraSigner(t *t
...
+					Type: xdr.LedgerEntryTypeAccount,
+					Account: &xdr.AccountEntry{
+						AccountId: xdr.MustAddress(initiatorEscrow.Address()),
+						SeqNum:    102,
+						Signers: []xdr.Signer{
+							{
+								Key: xdr.SignerKey{
+									Type:    xdr.SignerKeyTypeSignerKeyTypeEd25519,
+									Ed25519: xdr.MustAddress("GAKDNXUGEIRGESAXOPUHU4GOWLVYGQFJVHQOGFXKBXDGZ7AKMPPSDDPV").Ed25519,
+								},
```
leighmcculloch, line 840: 💡 There are some helpers for some XDR types in the xdr package, so you can shorten this a little.
```suggestion
								Key: xdr.MustSigner("GAKDNXUGEIRGESAXOPUHU4GOWLVYGQFJVHQOGFXKBXDGZ7AKMPPSDDPV"),
```

leighmcculloch (approved):
Looks great. I left a couple suggestions (💡) that I think might reduce the boilerplate that needs repeating in tests, but defer to you.

sdk/txbuildtest/txbuildtest.go
```diff
@@ -0,0 +1,49 @@
...
+	tr := xdr.TransactionResult{
+		FeeCharged: 123,
+		Result: xdr.TransactionResultResult{
+			Code:    code,
+			Results: &[]xdr.OperationResult{},
+		},
+	}
+	trXDR, err := xdr.MarshalBase64(tr)
+	if err != nil {
+		return "", fmt.Errorf("converting transaction result to base64 xdr: %w", err)
```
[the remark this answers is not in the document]
acharb: https://github.com/stellar/experimental-payment-channels/pull/271/commits/a9e88491513cf0f195a9b6a9d63600cd3d112694

sdk/state/ingest_test.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/271/commits/55d2381e67f9855f83edc2a4bb7d529262f52553

sdk/state/ingest_test.go
acharb: I couldn't think of a good way of fitting `xdr. NewLedgerEntryData` into it, since we may need to pass in multiple ledgerEntryData object of different types (like in `TestChannel_IngestTx_updateState_invalid_initiatorEscrowHasExtraSigner`)

but I changed it to assume it's an update change so we only have to pass in `[]xdr.LedgerEntryData` now, which also really simplifies what we have to pass in.

going to merge since approved but if you see something that can be improved lmk

leighmcculloch: If you change the input type to `entries ...interface{}`, and then inside the function: 
```go
for _, e := range entries {
	switch le := e.(type) {
	case xdr.AccountEntry:
		xdr.NewLedgerEntryData("Account", le)
	case xdr.TrustLineEntry:
		xdr.NewLedgerEntryData("TrustLine", le)
	default:
		panic("unrecognized entry type")
	}
}
```

leighmcculloch: Assuming it's an update sounds good to me.

acharb: Since we're requiring it to be a `[]xdr.LedgerEntryData{}` anyways any reason to make the input an interface? The input would be the same. Imo I think it's clearer making the inputs to be a required type 🤔

leighmcculloch: I was only suggesting so that we don't have to build the ledger entry data manually in tests. In tests I don't think it matters much if we have interface{}'s. Either way, this is great as is.

PR #272, opened 2021-08-31
sdk/agent: add NewAgent and Config

leighmcculloch:
### What
Add `NewAgent` and `Config` to the `sdk/agent` package and use those to setup the `agent.Agent`.

### Why
That's the same pattern we used in the `sdk/state` package for the `state.Channel`. It seemed to work pretty well there. It makes it really clear what are things that get set at start. It's helpful if we use the same pattern.


acharb (approved):
lgtm!

PR #273, opened 2021-08-31
sdk/agent, sdk/state: add persistence

### LINKED ISSUE #227: sdk/agent, sdk/state: define a story for persistence
We need to decide how the SDK will support persistence.

My preference is we find an elegant way to let the application decide how to persist the state.

I think this would look like:
- Adding an event handler to the agent that returns a copy of the state whenever is has changed, or for the sake of safety, whenever the agent calls a function on the channel.
- Moving the changing state of a channel into a type that has all its fields exported so the state of the channel can be easily copied out and is accessible to the caller.

leighmcculloch:
### What
Add methods to `state.Channel` for snapshotting and restoring a snapshot. Add a callback to `agent.Agent` for snapshotting the state of the agent and its channel, and a function for restoring that state.

### Why
It's necessary for the channel state to be persisted somewhere reliably so that an application using them can be stopped and restarted.

This approach of calling back with a snapshot of the state defers any need for the agent to choose a persistence method or define what persistence might look like. It could be more performant to have the agent write directly to a database, but there isn't a huge amount of data to store since the agent doesn't need to store data for every transactions processed. For where this project is write now storing the state in a blob seems like the shortest path to experimentation.

Close #227


sdk/agent/agent_test.go
```diff
@@ -39,6 +40,46 @@ func (f streamerFunc) StreamTx(cursor string, accounts ...*keypair.FromAddress)
...
+type snapshoterFunc func(a *Agent, s Snapshot)
+
+func (f snapshoterFunc) Snapshot(a *Agent, s Snapshot) {
+	f(a, s)
+}
+
+func assertAgentSnapshotsAndRestores(t *testing.T, agent *Agent, config Config, snapshot Snapshot) {
+	t.Helper()
+
+	snapshotJSON, err := json.Marshal(snapshot)
```
acharb, line 52: ❓ is this just making sure the Snapshot can be converted to JSON and back?

sdk/agent/agent.go
```diff
@@ -90,6 +100,32 @@ func NewAgent(c Config) *Agent {
...
 
+// Snapshot is a snapshot of the agent and its dependencies excluding any fields
+// provided in the Config when instantiating an agent. A Snapshot can be
+// restored into an Agent using NewAgentWithSnapshot.
+type Snapshot struct {
+	OtherEscrowAccount       *keypair.FromAddress
+	OtherEscrowAccountSigner *keypair.FromAddress
+	StreamerCursor           string
+	State                    *struct {
+		Initiator bool
```
acharb, line 111: ❓ is there a reason for putting `Inititiator` on the agent Snapshot? Instead of the Channel Snapshot since it's defined on state.Channel?

sdk/state/state.go
```diff
@@ -36,6 +36,40 @@ func NewChannel(c Config) *Channel {
 	return channel
 }
 
+type Snapshot struct {
```
acharb, line 39: 💭  thoughts on moving the Snapshot structs and logic to a snapshot.go file? It would make the state.go file less crowded and easier to tie the snapshot_test.go tests to the code it's testing

sdk/state/snapshot_test.go
```diff
@@ -0,0 +1,248 @@
...
+	require.Equal(t, channel, restoredChannel)
+}
+
+func TestNewChannelWithSnapshot(t *testing.T) {
+	localSigner := keypair.MustRandom()
+	remoteSigner := keypair.MustRandom()
+	localEscrowAccount := keypair.MustRandom().FromAddress()
+	remoteEscrowAccount := keypair.MustRandom().FromAddress()
+
+	localConfig := Config{
```
acharb, line 38: ❓  the current snapshot relies on the user to remember these fields in the `Config` struct, which then can be used to rehydrate a channel. But what if the user loses or accidentally changes these values? ie, why not include these in the snapshot?

sdk/state/snapshot_test.go
```diff
@@ -0,0 +1,248 @@
...
+		require.NoError(t, err)
+
+		cs, err := localChannel.State()
+		require.NoError(t, err)
+		assert.Equal(t, StateClosed, cs)
+
+		cs, err = remoteChannel.State()
+		require.NoError(t, err)
+		assert.Equal(t, StateClosed, cs)
+	}
```
acharb, line 247: 💭 can we add a final check? Since ingesting the closeTx could change agent/channel state
```suggestion
	}
	assertChannelSnapshotsAndRestores(t, localConfig, localChannel)
	assertChannelSnapshotsAndRestores(t, remoteConfig, remoteChannel)
```

acharb (commented):
💯  looks great left mostly ❓ and some 💭

sdk/state/snapshot_test.go
leighmcculloch: >what if the user loses or accidentally changes these values

Applications building on this should have tests that pick that up pretty quickly. It's the sort of thing if it is broken, it's always broken rather than intermittently broken.

>why not include these in the snapshot?

My first attempt at adding snapshotting had the function signature as `NewChannelFromSnapshot(s Snapshot)` without the config, however the `Config`s for both the `Channel` and the `Agent` both contain a mix of state information, dependencies (e.g. `Submitter`, `Streamer`, etc), and sensitive information (e.g. signing keys). If we include some of the `Config` fields in the `Snapshot` struct when rehydrating we have to handle the case where they're inconsistent, and what should happen then.

I think the simplest path at the moment is to say that the `Snapshot` contains anything about the internal state of the channel that isn't already captured in the `Config` that's required to rehydrate the channel.

I have some ideas for how to refactor `Config` so that it would be easier to move more into the `Snapshot` to be more like what you suggest, but it's a bigger piece of work and it'll be easier to see how effective the refactor is if the `Snapshot` is already in place before that happens.

sdk/agent/agent.go
leighmcculloch: The reason is that the `Initiator` is not internal state of `state.Channel`, it's external config that the caller provides when setting it up. The `state.Snapshot` only stores internal state, not initialization config.

sdk/agent/agent.go
```diff
@@ -90,6 +100,32 @@ func NewAgent(c Config) *Agent {
...
+		Snapshot  state.Snapshot
+	}
+}
+
+// NewAgentWithSnapshot creates an agent using a previously generated snapshot
+// so that the new agent has the same state as the previous agent.
+func NewAgentWithSnapshot(c Config, s Snapshot) *Agent {
```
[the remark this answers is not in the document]
leighmcculloch: Good point. :+1:

sdk/agent/agent.go
```diff
@@ -101,6 +137,7 @@ type Agent struct {
 	balanceCollector        BalanceCollector
 	submitter               Submitter
 	streamer                Streamer
+	snapshoter              Snapshoter
```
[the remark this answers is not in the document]
leighmcculloch: Doh, thanks!

sdk/state/state.go
leighmcculloch: I think we should keep the `New` functions and their related types close to each other. That's a good point with the disconnect with `snapshot_test.go` though. I renamed it to `state_test.go`.

sdk/agent/agent_test.go
```diff
@@ -66,21 +107,25 @@ func TestAgent_openPaymentClose(t *testing.T) {
...
-		escrowAccountKey:    localEscrow.FromAddress(),
-		escrowAccountSigner: localSigner,
-		logWriter:           io.Discard,
-		events:              localEvents,
+		EscrowAccountKey:    localEscrow.FromAddress(),
+		EscrowAccountSigner: localSigner,
+		LogWriter:           io.Discard,
+		Events:              localEvents,
 	}
+	localConfig.Snapshoter = snapshoterFunc(func(a *Agent, s Snapshot) {
```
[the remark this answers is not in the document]
leighmcculloch: Unfortunately we can't because the `localConfig` is referenced inside the `snapshotterFunc`, so it has to be set outside the definition of `localConfig`. This shouldn't have any functional/material impact though.

sdk/state/state.go
```diff
@@ -36,6 +36,40 @@ func NewChannel(c Config) *Channel {
...
+
+	OpenAgreement            OpenAgreement
+	OpenExecutedAndValidated bool
+	OpenExecutedWithError    bool
+
+	LatestAuthorizedCloseAgreement   CloseAgreement
+	LatestUnauthorizedCloseAgreement CloseAgreement
+}
+
+func NewChannelWithSnapshot(c Config, s Snapshot) *Channel {
```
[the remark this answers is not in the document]
leighmcculloch: :100:

leighmcculloch (commented):
Thanks for the feedback! I addressed the comments in 272886ce4660e71922d62c9ba2b28d5ab69ff382.

sdk/state/snapshot_test.go
leighmcculloch: 💯 Thanks for catching that.

leighmcculloch (commented):
Missed one comment, addressed now in 4ece060.

sdk/state/snapshot_test.go
acharb: ok sgtm 👍 . Could you add this line: `Snapshot contains anything about the internal state of the channel that isn't already captured in the Config` as a comment above the Snapshot struct as a reminder?

acharb (approved):
lgtm!, just had one last [suggestion](https://github.com/stellar/experimental-payment-channels/pull/273#discussion_r701986406) for adding a comment

sdk/state/snapshot_test.go
leighmcculloch: 💯 Good call.  The docs are a bit thin and missing from some types/funcs I'm adding here.

leighmcculloch (commented):
👍🏻 Addressed in fd9ade9.

Issue #49, opened 2021-05-25
sdk/state: implement withdraw


leighmcculloch:
Once the channel is open a participant may want to withdraw funds that belong to them without closing the entire channel. It is possible to top-up/deposit into the channel, it could be possible to withdraw without closing also.

leighmcculloch:
> I'd like to defer implementing withdraw and revisit the design at some point in the future, rather than implementing it now.
> 
> @acharb @nikhilsaraf @accordeiro @tomerweller Thoughts?

I'm going to close this issue as no one has chimed in to champion keeping this feature. Please reopen if you think withdrawal is something we need.

leighmcculloch:
Reopened but will keep out of a milestone.

PR #274, opened 2021-08-31
sdk/state: remove old todo

leighmcculloch:
### What
Remove old todo.

### Why
The todo comment has been there for a while but isn't something we're likely to do soon. It's something we'll need when we do #49 but we can address how that is implemented when and if we come to it.


acharb (approved):
lgtm!

PR #276, opened 2021-08-31
sdk/state: reorder funcs

leighmcculloch:
### What
Move the `Config` and `NewChannel` type and function above the `Channel` type.

### Why
The type and function are the entry point to using the `Channel`, but they're buried below some other types and functions. Moving them to the top makes them easier to discover and matches how the same pattern in the `sdk/agent` package.


acharb (approved):
lgtm!

PR #277, opened 2021-09-01
sdk/state: improving unit and integration tests by using the ingest logic

acharb:
**WHAT**
update tests to use the ingest logic for putting channels into an open state, and validate the channel.State is in `StateOpen`

**WHY**
before channels were put into an open state by accessing internal data. Putting them into an OpenState by using exported methods will help with having to change the tests after future logic changes, and validating the channel state will ensure the channels are in the right state going forward


leighmcculloch (approved):
🎉

PR #278, opened 2021-09-01
sdk/state: store starting sequence in open agreement

### LINKED ISSUE #239: sdk/state: include starting sequence in open agreement
The starting sequence is not include in any agreement and therefore relies on assumptions and introduces some ambiguity. Both participants determine the starting sequence of the escrow account by looking up the current sequence of the initiator account. We can make the open process more deterministic by simply including the starting sequence in the open agreement.

@acharb brought this up today.

leighmcculloch:
### What
Store the starting sequence in the open agreement.

### Why
The starting sequence is an assumed value that implicitly forms part of the agreement, however it would be better if it was an explicit part of the agreement and stored there rather than store inside the channel as a global value.

As a bonus this lets an escrow account holder to state that the open will occur at some future sequence number. On the surface this might be a cause for concern for the responder but it doesn't really change the security of anything since the initiator could always shift their sequence number after starting the open handshake. This just provides some flexibility and improves the debug-ability in cases that the two participants for some reason think the starting sequence is different.

I didn't plan on doing this now except that was I work on #227 (#273) it is inconvenient that the escrow accounts have their sequence numbers stored and what is stored there is important, so that led me to moving this up.

Close #239


sdk/state/ingest.go
```diff
@@ -273,9 +273,11 @@ func (c *Channel) ingestFormationTx(tx *txnbuild.Transaction, resultXDR string,
 	}
 
 	// Validate the initiator escrow account sequence number is correct.
-	if int64(initiatorEscrowAccountEntry.SeqNum) != c.startingSequence {
+	fmt.Println("meta", initiatorEscrowAccountEntry.SeqNum)
```
acharb: leftover print statements?

sdk/state/ingest.go
```diff
@@ -273,9 +273,11 @@ func (c *Channel) ingestFormationTx(tx *txnbuild.Transaction, resultXDR string,
...
+	fmt.Println("formation", formationTx.SequenceNumber())
+	if int64(initiatorEscrowAccountEntry.SeqNum) != formationTx.SequenceNumber() {
```
acharb, line 276: 💭  it _should_ always be the same value but thoughts on pulling the sequence number from the channel.openAgreement? To get it from the source

sdk/state/open.go
```diff
@@ -251,8 +254,6 @@ func (c *Channel) ConfirmOpen(m OpenAgreement) (open OpenAgreement, err error) {
 		return OpenAgreement{}, fmt.Errorf("validating open agreement: %w", err)
 	}
 
-	c.startingSequence = c.initiatorEscrowAccount().SequenceNumber + 1
-
```
acharb, line 255: ❓  this was implicitly checking that the EI.SequenceNumber is correct (otherwise the signature validation fails). How does the responder know now if the proposed sequence number is correct?

or does the responder just assume it's correct, and if it's a lesser value then the formation fails and channel doesn't open?

acharb (commented):
this change is awesome 💯  had 1 💭  and 1 ❓

sdk/state/open.go
leighmcculloch: I initially had the same concern. However, after more thought, there's no way the responder can know for sure whether a sequence is correct or not, even if the responder was to attempt to determine the sequence themselves it might not be correct. The only thing the responder can check is if the sequence is in the past. If it's in the past – i.e. the sequence is less than the current escrow account sequence – then there's a good chance the initiator is just trying to waste the responders time. I don't think it is really worth checking for that though as there's no material impact.

sdk/state/ingest.go
leighmcculloch: We could pull it from either place. Worth noting that `formationTx` is not the `tx` being passed in to be ingested, it's the tx that `OpenTx` creates so it is based on the internal state of the channel and not an external source, so we're still comparing the external data against purely internal data.

If we pulled it from the open agreement we'd be pulling the `openAgreement.Details.StartingSequence`. As it happens the `formationTx` sequence is set based off that value, so I think comparing them against one or the other is much the same. Something I like about pulling it from the `formationTx` is that on Stellar the rule is that if a tx is executed the account becomes the account's sequence number, so this check is really guaranteeing that has occurred.

acharb: ya  I was thinking since the formationTx gets the sequenceNum from the open agreement, so the original source. But formationTx sgtm 👍 

but jw what do you mean by this?
> if a tx is executed the account becomes the account's sequence number

acharb (approved):
lgtm!

sdk/state/ingest.go
leighmcculloch: >what do you mean by this?

That's how the Stellar protocol works. When a tx is executed The source accounts sequence number becomes the sequence number of the tx. It just happens account sequence numbers also have to be sequential, but that requirement will be lifted in cap21.

PR #279, opened 2021-09-01
sdk/state: remove escrow seq and balance from new interface

### LINKED ISSUE #239: sdk/state: include starting sequence in open agreement

leighmcculloch:
### What
Remove the escrow sequence and balance from the interface of instantiating a new channel.

### Why
These values rarely need to be part of initializing a channel and are info that'll be set via ingestion. The changes in #278 removed some of the dependencies on this, which make it possible to remove, but I wanted to keep it in a separate PR to keep the reviewable chunk digest-able.

Close #239


acharb (approved):
lgtm!

Issue #235, opened 2021-08-17
sdk/state: error if proposing or confirming a payment or close if channel open is authorized but not executed


leighmcculloch:
In #230 we added the capability to identify when the channel hasn't been authorized to be open, and blocking proposing or confirming any payments until that has occurred. When a channel is authorized to be open the channel hasn't actually opened until the open tx has been submitted to the network and the state of the accounts inspected after that to verify the channel is opened and safe to make payments on.

acharb:
@leighmcculloch this is addressing the same as https://github.com/stellar/experimental-payment-channels/issues/261 correct?

leighmcculloch:
This issue is new logic, error in conditions it currently doesn't error in. I thought #261 was a refactor, use the new state instead of interrogating other variables. So they seem like two different things. Maybe worth doing #261 first though.

PR #280, opened 2021-09-01
sdk/state: add validation that the channel has ingested/verified a formationTx before proposing/confirming payments or closes

acharb:
**WHAT**
adds a validation to check that the `channel.openExecutedAndValidated` is true before proposing/confirming a payment or close

**WHY**
if the flag is true then the channel ingested the formation transaction and the channel is officially open. If it's false then the channel should not be able to make or confirm new payments or closes.

closes https://github.com/stellar/experimental-payment-channels/issues/235


sdk/state/close.go
```diff
@@ -93,7 +93,7 @@ func (c *Channel) ProposeClose() (CloseAgreement, error) {
 	}
 
 	// If the channel is not open yet, error.
-	if c.latestAuthorizedCloseAgreement.isEmpty() {
+	if c.latestAuthorizedCloseAgreement.isEmpty() || !c.openExecutedAndValidated {
```
acharb, line 96: grouped these checks together, I think if either fail we'd want to throw the same error

leighmcculloch (approved):
👏🏻

PR #281, opened 2021-09-02
sdk/state: replace keypair Parse* with MustParse* in tests

leighmcculloch:
### What
Replace `keypair` `Parse*` with `MustParse*` in tests.

### Why
In most of our tests we keep them a little more brief by using the `MustParse*` functions to avoid error checking. A few places we are more verbose using the `Parse*` variations. Changing for consistency.


acharb (approved):
lgtm

PR #285, opened 2021-09-02
examples/console: add persistence

### LINKED ISSUE #227: sdk/agent, sdk/state: define a story for persistence

leighmcculloch:
### What
Add persistence to a simple JSON file to the `examples/console` example.

### Why
To exercise the functionality adding snapshotting (persistence) that was added in #273.

Close #227


examples/console/main.go
```diff
@@ -72,88 +95,131 @@ func run() error {
...
+			SequenceNumberCollector:    horizon,
+			BalanceCollector:           horizon,
+			Submitter:                  submitter,
+			Streamer:                   horizon,
+			EscrowAccountKey:           escrowAccountKey,
+			EscrowAccountSigner:        signerKey,
+			LogWriter:                  os.Stderr,
+			Events:                     events,
+		}
+		if filename != "" {
```
acharb, line 213: 💭  feels weird to me that there could be a case where file exists but filename is empty. Which would leave the config with an empty Snapshotter field.

Thoughts on erroring if filename isn't given on the cli? Alternatively adding a default filename name at the top eg. "demo-snapshot.json"

acharb (commented):
💯  just had 1 💭

examples/console/main.go
leighmcculloch: Thanks for catching this, the `filename != ""` condition is unnecessary. If there's a `file` there is always a `filename`.

leighmcculloch (commented):
Addressed feedback in 5d72ec4.

acharb (approved):
lgtm!

Issue #275, opened 2021-08-31
sdk/agent: handle errs and unexpected early exit during horizon streaming 


leighmcculloch:
In #265 we connected horizon streaming with the state ingestion to bring transactions into the application, however it does not yet handle errors, or rather it handles them a little aggressively by panicing.

https://github.com/stellar/experimental-payment-channels/blob/99c2143a18ae5f389e85ffc488ed220b7d5348c5/sdk/horizon/horizon.go#L125-L128

I think there are some things we need to decide:
- If streaming sees and error, should it stop and give up or retry.
- If it stops propogating an error makes sense, something has happened and some action needs taking. But for the retry case this propogation is less clear. Do we propogate so that other code, like the logger, can log about it, but it wouldn't take action because retrying will happen. Should it retry forever?

I suggest we improve it at least with:
- Infinite retrying.
- Propagate the error to an errors channel directly to the application, not via the agent.

Other possible improvements:
- Use a backoff strategy for retrying.

PR #288, opened 2021-09-02
sdk/horizon: split up horizon type

leighmcculloch:
### What
Move the functionality in the `horizon.Horizon` type into `horizon.BalanceCollector`, `horizon.SequenceNumberCollector`, `horizon.Submitter`, and `horizon.Streamer`.

### Why
For #275 I'm adding some fields to the type that holds the Horizon `Streamer` implementation and it is a bit odd to have these fields on the `horizon.Horizon` type when those fields only relate to a subset of that types functionality. It's a bit of an anti-pattern that the `horizon.Horizon` type holds a bunch of unrelated functionality. Splitting it up helps the change I'm going to be making clearer.

All the logic is a copy and paste in this change, no logic has been changed.

For #275


acharb (approved):
lgtm1

PR #289, opened 2021-09-02
sdk/horizon: add callback for handling errors

### LINKED ISSUE #275: sdk/agent: handle errs and unexpected early exit during horizon streaming

leighmcculloch:
### What
Add callback for handling errors to the `horizon.Streamer`.

### Why
If an error occurs while streaming we don't want the streamer to panic. Small outages in a Horizon deployment would cause the implementation to stop ingesting. We should retry ingestion whenever things go wrong but give the application a way to know what is going on so that they can cancel/stop ingestion if they choose.

It's noteworthy that the `StreamTx` doesn't bubble the error back up into the `Agent`. There would be little benefit of doing so. The `Streamer` is an external dependency provided to the agent and it can provide errors directly to the application. This ensures that the error handling scheme is entirely flexible and not dictated by the agent.

Close #275


acharb (approved):
lgtm!

acharb (approved):

Issue #248, opened 2021-08-24
sdk/state: expose payment amount in close agreements


leighmcculloch:
The sdk/state package has an inconsistent perspective on what it means to make a payment.

When proposing a payment the participant specifies the amount to pay the other participant. The SDK internalizes and abstracts a lot of complexity relating to how the settlement balance of the channel changes with payments to make it simpler for the caller to make a payment.

However, when a participant receives their payment they don't really receive a payment but only an update to the final settlement balance. The participant is on their own to understand what the new settlement balance means.

We need to make the payment experience consistent and there are probably a few ways we can do this. Let's discuss possible ideas in this issue.

cc @acharb

PR #291, opened 2021-09-02
sdk/state: add FinalPaymentAmount field to CloseAgreement

acharb:
**WHAT**
adds a field to CloseAgreement that tells what the final payment amount is, ie what the last payment amount is created by this CloseAgreement.

**WHY**
for the receiver of a payment there's currently not a good way of telling how much a payment is, since the close agreement only tells you the final balance. And comparing with the previous Balance introduces race conditions (described [here](https://github.com/stellar/experimental-payment-channels/issues/248)). So store a new field that gets validated at the time of confirming.

closes https://github.com/stellar/experimental-payment-channels/issues/248

### DIFF
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -194,14 +194,19 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 
 	// If the close agreement payment amount is incorrect, error.
 	pa := ca.Details.PaymentAmount
-	if ca.Details.ConfirmingSigner.Equal(c.localSigner.FromAddress()) && c.initiator {
+	if !c.initiatorProposedPayment(ca) {
 		pa = ca.Details.PaymentAmount * -1
 	}
 	if ca.Details.Balance-c.Balance() != pa {
 		return fmt.Errorf("close agreement payment amount is unexpected: current balance: %d proposed balance: %d payment amount: %d initiator proposed: %t",
-			c.Balance(), ca.Details.Balance, ca.Details.PaymentAmount, ca.Details.ConfirmingSigner.Equal(c.localSigner.FromAddress()) && c.initiator)
+			c.Balance(), ca.Details.Balance, ca.Details.PaymentAmount, c.initiatorProposedPayment(ca))
 	}
 	return nil
+}
+
+func (c *Channel) initiatorProposedPayment(ca CloseAgreement) bool {
+	return ca.Details.ConfirmingSigner.Equal(c.localSigner.FromAddress()) && !c.initiator ||
+		ca.Details.ConfirmingSigner.Equal(c.remoteSigner.FromAddress()) && c.initiator
 }
 
 // ConfirmPayment confirms an agreement. The destination of a payment calls this
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -197,19 +197,15 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 
 	// If the close agreement payment amount is incorrect, error.
 	pa := ca.Details.PaymentAmount
-	if !c.initiatorProposedPayment(ca) {
+	proposerIsResponder := ca.Details.ProposingSigner.Equal(c.responderSigner())
+	if proposerIsResponder {
 		pa = ca.Details.PaymentAmount * -1
 	}
 	if c.Balance()+pa != ca.Details.Balance {
 		return fmt.Errorf("close agreement payment amount is unexpected: current balance: %d proposed balance: %d payment amount: %d initiator proposed: %t",
-			c.Balance(), ca.Details.Balance, ca.Details.PaymentAmount, c.initiatorProposedPayment(ca))
+			c.Balance(), ca.Details.Balance, ca.Details.PaymentAmount, !proposerIsResponder)
 	}
 	return nil
-}
-
-func (c *Channel) initiatorProposedPayment(ca CloseAgreement) bool {
-	return ca.Details.ProposingSigner.Equal(c.localSigner.FromAddress()) && c.initiator ||
-		ca.Details.ProposingSigner.Equal(c.remoteSigner.FromAddress()) && !c.initiator
 }
 
 // ConfirmPayment confirms an agreement. The destination of a payment calls this
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -193,7 +193,7 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 
 	// If the close agreement payment amount is incorrect, error.
 	if ca.Details.Balance-c.Balance() != ca.Details.FinalPaymentAmount {
-		return fmt.Errorf("close agreement payment amount is incorrect")
+		return fmt.Errorf("close agreement payment amount is unexpected: current balance: %d proposed balance: %d payment amount: %d")
 	}
 	return nil
 }
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -23,7 +23,7 @@ type CloseAgreementDetails struct {
 	ObservationPeriodLedgerGap int64
 	IterationNumber            int64
 	Balance                    int64
-	FinalPaymentAmount         int64
+	PaymentAmount              int64
 	ProposingSigner            *keypair.FromAddress
 	ConfirmingSigner           *keypair.FromAddress
 }
@@ -118,11 +118,12 @@ func (c *Channel) ProposePayment(amount int64) (CloseAgreement, error) {
 		return CloseAgreement{}, fmt.Errorf("cannot start a new payment while an unfinished one exists")
 	}
 
-	finalPaymentAmount := amount
-	if !c.initiator {
-		finalPaymentAmount = amount * -1
+	newBalance := int64(0)
+	if c.initiator {
+		newBalance = c.Balance() + amount
+	} else {
+		newBalance = c.Balance() - amount
 	}
-	newBalance := c.Balance() + finalPaymentAmount
 
 	if c.amountToRemote(newBalance) > c.localEscrowAccount.Balance {
 		return CloseAgreement{}, fmt.Errorf("amount over commits: %w", ErrUnderfunded)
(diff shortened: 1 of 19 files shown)


sdk/state/payment.go
```diff
@@ -23,6 +23,7 @@ type CloseAgreementDetails struct {
 	ObservationPeriodLedgerGap int64
 	IterationNumber            int64
 	Balance                    int64
+	FinalPaymentAmount         int64
```
acharb, line 26: @leighmcculloch fyi `LastPaymentAmount` that you suggested [here](https://github.com/stellar/experimental-payment-channels/issues/248#issuecomment-906815483) had me initially confused with the meaning of last as in the previous close agreement. So I changed to use Final instead

leighmcculloch: I agree, naming for this is going to be hard.

`FinalPaymentAmount` also has an ambiguiity problem, because final payment amount sounds a lot like the net settlement amount, which is the final amount that will be exchanged between the participants when the channel closes.

Untitled-2021-08-20-2126-2

leighmcculloch: I agree, naming for this is going to be hard.

`FinalPaymentAmount` also has an ambiguiity problem, because final payment amount sounds a lot like the net settlement amount, which is the final amount that will be exchanged between the participants when the channel closes.

Untitled-2021-08-20-2126-2

The names we've considered so far, and some other names:

- `LastPaymentAmount` – Problem is this sounds a lot like the previous close agreements payment rather than this close agreements payment.
- `FinalPaymentAmount` – Problem is this sounds a lot like the net-settlement payment when the channel closes.
- `PaymentAmount` – Problem it sounds like it could be the close agreement, or the individual payment amount.
- `ProposerPaymentAmount` 
- `BalanceDiff` – This could be an initiator relative amount, similar to how you implemented the relativeness of the amount in this PR. We could use this to calculate a proposer payment amount.
- `LastBalance`/`PreviousBalance` – What if we just encode the last balance. So the current balance and last balance are encoded. Theoretically we can calculate whatever else we want from that. The name for this field is ultra clear too.

sdk/state/payment.go
```diff
@@ -117,12 +118,12 @@ func (c *Channel) ProposePayment(amount int64) (CloseAgreement, error) {
...
 	}
 
-	newBalance := int64(0)
-	if c.initiator {
-		newBalance = c.Balance() + amount
-	} else {
-		newBalance = c.Balance() - amount
+	finalPaymentAmount := amount
+	if !c.initiator {
+		finalPaymentAmount = amount * -1
```
leighmcculloch, line 123: It looks like the payment amount is relative to the initiator, like the balance, but I think for ergonomics we should aim for it to be relative to the sender/receiver. We don't specify who the sender/receiver is explicitly, but implicitly it's always the proposer in our implementation. Relative to the proposer would make the value always positive since we only allow the proposal of payments to the confirmer.

Wdyt?

Note, if we go with `LastBalance`/`PreviousBalance` as the field to encode in the close agreement mentioned in my comment above, then I think this initiative relative approach makes sense.

sdk/state/payment.go
```diff
@@ -188,6 +190,11 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 	if !ca.Details.ConfirmingSigner.Equal(c.localSigner.FromAddress()) && !ca.Details.ConfirmingSigner.Equal(c.remoteSigner) {
 		return fmt.Errorf("close agreement confirmer does not match a local or remote signer, got: %s", ca.Details.ConfirmingSigner.Address())
 	}
+
+	// If the close agreement payment amount is incorrect, error.
+	if ca.Details.Balance-c.Balance() != ca.Details.FinalPaymentAmount {
+		return fmt.Errorf("close agreement payment amount is incorrect")
```
leighmcculloch, line 196: Whether it is incorrect or not is subjective and sort of alludes to some malice that may not exist. I think we could say the amount is unexpected. We should probably also include the amounts in the error for debugging.
```suggestion
		return fmt.Errorf("close agreement payment amount is unexpected: current balance: %d proposed balance: %d payment amount: %d")
```

leighmcculloch (commented):
Some thoughts inline.

sdk/state/payment.go
acharb: imo I like having one payment amount field and one balance field. Instead of two balance fields, which feels redundant since we'd most likely just be using it to get the payment amount anyways. `LastBalance` also has the same problem of using `Last` 😆 .

I initially liked `ProposePaymentAmount` but then if I looked at a close agreement 5 back I wouldn't remember if I proposed or not.

I think my top vote would be `PaymentAmount`, and allow it to be negative. It's consistent with how Balance is defined currently.

 I don't feel _super_ strongly about this though, so if you like 2 balance fields I'm game

leighmcculloch: > `PaymentAmount`, and allow it to be negative

I think this will be confusing. When a proposer who is the responder proposes a payment they pass into the `ProposePayment` function a `amount` that is always positive, but when they receive `PaymentSentEvent` the `CloseAgreement` within will have `PaymentAmount` as negative. Now that I think of it, I think if we're going to have a field for the payment amount that mirrors the `amount` parameter of the `ProposePayment` function, we should make this field behave identically. Then there's consistency between the `ProposePayment(amount int64)` and the `CloseAgreement{PaymentAmount int64}`.

Wdyt?

> It's consistent with how Balance is defined currently.

That's true. I think the negative `Balance` is convenient for the internals, but it's a shame that it leaks outside via the `CloseAgreement` and we don't have to carry that to other fields if they don't need it.

acharb: > I think if we're going to have a field for the payment amount that mirrors the amount parameter of the ProposePayment function, we should make this field behave identically

true. So how about `PaymentAmount` and leave it as always positive? And then the `ProposingSigner` field can tell who was the one who proposed instead of the positive vs negative

leighmcculloch: Yup I think that works.

acharb: https://github.com/stellar/experimental-payment-channels/pull/291/commits/c51561b26664f3693a2d9474eb85208e0e50bd78

acharb: fyi added a helper method here for checking if the closeAgreement proposer is the initiator. That also led me to add a validation for checking that the ProposingSigner is either the local or remote signer, so that it's safe to use this helper method

https://github.com/stellar/experimental-payment-channels/pull/291/commits/ecb15c57dfa6e6a11e9585bd865930c52550682d

sdk/state/payment.go
```diff
@@ -188,9 +191,27 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 	if !ca.Details.ConfirmingSigner.Equal(c.localSigner.FromAddress()) && !ca.Details.ConfirmingSigner.Equal(c.remoteSigner) {
 		return fmt.Errorf("close agreement confirmer does not match a local or remote signer, got: %s", ca.Details.ConfirmingSigner.Address())
 	}
+	if !ca.Details.ProposingSigner.Equal(c.localSigner.FromAddress()) && !ca.Details.ProposingSigner.Equal(c.remoteSigner) {
+		return fmt.Errorf("close agreement proposer does not match a local or remote signer, got: %s", ca.Details.ConfirmingSigner.Address())
+	}
```
leighmcculloch, line 196: 🎉 Nice catch on this missing validation!

sdk/state/payment.go
```diff
@@ -188,9 +191,27 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
...
+
+	// If the close agreement payment amount is incorrect, error.
+	pa := ca.Details.PaymentAmount
+	if !c.initiatorProposedPayment(ca) {
```
leighmcculloch, line 200: 💡 We have the `responderSigner` function you can compare the proposer with, which I think is simpler than the `initiatorProposedPayment` function included in this PR.
```suggestion
	proposerIsResponder := ca.Details.ProposingSigner.Equal(c.responderSigner())
	if proposerIsResponder {
```

sdk/state/payment.go
```diff
@@ -188,9 +191,27 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
...
+		pa = ca.Details.PaymentAmount * -1
+	}
+	if ca.Details.Balance-c.Balance() != pa {
```
leighmcculloch, line 203: 💡 I don't feel strongly about this, but when I think in my mind about what we're testing it's that the current balance is modified by the payment amount to reach the new balance, so I had to do a double take on the above to figure out if it was right, but the following is what my mind expected:
```suggestion
	if c.Balance() + pa == ca.Details.Balance {
```

sdk/state/payment.go
```diff
@@ -188,9 +191,27 @@ func (c *Channel) validatePayment(ca CloseAgreement) (err error) {
 	if !ca.Details.ConfirmingSigner.Equal(c.localSigner.FromAddress()) && !ca.Details.ConfirmingSigner.Equal(c.remoteSigner) {
 		return fmt.Errorf("close agreement confirmer does not match a local or remote signer, got: %s", ca.Details.ConfirmingSigner.Address())
 	}
+	if !ca.Details.ProposingSigner.Equal(c.localSigner.FromAddress()) && !ca.Details.ProposingSigner.Equal(c.remoteSigner) {
+		return fmt.Errorf("close agreement proposer does not match a local or remote signer, got: %s", ca.Details.ConfirmingSigner.Address())
```
leighmcculloch, line 195: ❗ 
```suggestion
		return fmt.Errorf("close agreement proposer does not match a local or remote signer, got: %s", ca.Details.ProposingSigner.Address())
```

leighmcculloch (commented):
One minor bug❗, and a couple suggestions💡, otherwise 💯.

sdk/state/payment.go
acharb: I had a similar thought
https://github.com/stellar/experimental-payment-channels/pull/291/commits/e5b16611ccd94ef0ea127d397990f33a245700f3

sdk/state/payment.go
acharb: ooh didn't see that cool
https://github.com/stellar/experimental-payment-channels/pull/291/commits/a67f38e91ae613abcd3e289323d490296d70b3c2

leighmcculloch (approved):

PR #293, opened 2021-09-04
sdk/state: store tx hashes in agreements

leighmcculloch:
### What
Store tx hashes in the open and close agreements, and validate that they match built transactions when rebuilding txs, or confirming agreements.

### Why
To add some resilience to tx building so that it is immediately clear if an update causes txs to build different. To improve debugging if two participants interacting appear to have incorrect signatures due to txs being built differently.

I started adding this while working on storing txs (#199). Much of the code relies on verifying signatures to detect an issue with transaction building, which is fine and functional, but it is difficult to tell when transaction building is the cause of verification failure rather than incorrect signatures.

This is a bit off the planned path for #199, but is related to it.


leighmcculloch:
After making the change and reviewing the impact it occurred to me that this doesn't add enough value for the cost to code maintainability. This change wouldn't improve security and only improve debuggability, but the burden is large on an implementation. Not only on the code, but also on testing.

PR #294, opened 2021-09-04
sdk/state: store and reuse txs

### LINKED ISSUE #199: sdk/state: store a copy of built txs

leighmcculloch:
### What
Store transactions that the state machine builds, and reuse them, rather than rebuilding as they are needed.

### Why
Our initial plan was to rebuild transactions as needed, however if the Stellar Go SDK that we use to build txs ever changes the transactions being built even in the slightest the signatures stored won't match the rebuilt transactions. Storing transactions at the point in time we generate the signatures is the only way to ensure we can submit that same transaction in the far future.

This change also stores the transaction hashes alongside them, since it is small effort to do so. Storing both the transactions and their hashes has the bonus that it makes debugging easier since the snapshotted state of the channel includes its transactions.

Close #199


sdk/state/close.go
```diff
@@ -18,8 +18,16 @@ import (
 // 6. If A or B declines or is not responsive at any step, A or B may submit the
 //    original close tx after the observation period.
 
-func (c *Channel) closeTxs(oad OpenAgreementDetails, d CloseAgreementDetails) (txDecl *txnbuild.Transaction, txClose *txnbuild.Transaction, err error) {
-	txClose, err = txbuild.Close(txbuild.CloseParams{
+func (c *Channel) closeTxs(oad OpenAgreementDetails, d CloseAgreementDetails) (txs CloseAgreementTransactions, err error) {
```
acharb, line 21: 💡 can we add a method comment here describing how this method is essentially loading the "cached" transaction if the input open agreement and close agreement are the same as the channel ones? I think that's something not obvious but worth pointing out, and is how we call up the saved `channel.<>AgreementTransactions`

sdk/state/hash.go
```diff
@@ -0,0 +1,35 @@
+package state
+
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+type TransactionHash [32]byte
+
+func (h TransactionHash) String() string {
```
acharb, line 10: ❓  for my own understanding where/how are these methods called? I'm assuming somewhere implicitly

sdk/state/open.go
```diff
@@ -114,7 +125,10 @@ type OpenParams struct {
 	StartingSequence           int64
 }
 
-func (c *Channel) openTxs(d OpenAgreementDetails) (decl, close, formation *txnbuild.Transaction, err error) {
+func (c *Channel) openTxs(d OpenAgreementDetails) (txs OpenAgreementTransactions, err error) {
```
acharb, line 128: 💡 same thought here about adding a method comment similar to `closeTxs()`

sdk/state/state.go
```diff
@@ -45,12 +45,15 @@ type Snapshot struct {
...
-	OpenExecutedWithError    bool
-
-	LatestAuthorizedCloseAgreement   CloseAgreement
-	LatestUnauthorizedCloseAgreement CloseAgreement
+	OpenAgreement             OpenAgreement
+	OpenAgreementTransactions OpenAgreementTransactions
+	OpenExecutedAndValidated  bool
+	OpenExecutedWithError     bool
+
+	LatestAuthorizedCloseAgreement               CloseAgreement
```
acharb, line 53: ❓ in the [ticket](https://github.com/stellar/experimental-payment-channels/issues/199) you used the idea of a packaged field that grouped the agreements with their transactions. I initially liked that to decrease the chance of the transactions getting out of sync with the agreement
eg. updating `channel.latestAuthorizedCloseAgreementPackage` at one time vs updating `channel. latestAuthorizedCloseAgreement` and then `channel.latestAuthorizedCloseAgreementTransactions` afterward

I'm not pushing for the single field as it doesn't _really_ protect them getting out of sync, but just wondering what your thought process was for not packaging them.

acharb (commented):
looks awesome 🙌 . Left a couple questions/thoughts

sdk/state/close_test.go
```diff
@@ -27,15 +27,15 @@ func TestChannel_CloseTx(t *testing.T) {
 		LocalEscrowAccount:  localEscrowAccount,
 		RemoteEscrowAccount: remoteEscrowAccount,
 	})
-	channel.openAgreement = OpenAgreement{
+	oa := OpenAgreement{
```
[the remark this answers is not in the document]
leighmcculloch: I needed to build the txs with `closeTxs` using the agreement before setting the agreement in the channel because the channel looks at the agreement to decide if it should reuse the stored tx.

sdk/state/close_test.go
```diff
@@ -70,6 +76,43 @@ func TestChannel_CloseTx(t *testing.T) {
 		{Hint: localSigner.Hint(), Signature: []byte{1}},
 		{Hint: remoteSigner.Hint(), Signature: []byte{3}},
 	}, closeTx.Signatures())
+
+	// Check stored txs are used by replacing the stored tx with an identifiable
+	// tx and checking that's what is used for the authorized closing transactions.
+	imposterTx, err := txnbuild.NewTransaction(txnbuild.TransactionParams{
```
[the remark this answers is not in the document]
leighmcculloch: Good point. 865888f

sdk/state/open.go
```diff
@@ -41,38 +41,49 @@ func (oas OpenAgreementSignatures) isFull() bool {
...
-	err = verifySigned(formation, networkPassphrase, signer, s.Formation)
+	err = verifySigned(txs.Formation, networkPassphrase, signer, s.Formation)
 	if err != nil {
 		return fmt.Errorf("verifying formation signed: %w", err)
 	}
 	return nil
 }
 
+// OpenAgreementTransactions contain all the transaction hashes and
+// transactions for the transactions that make up the open agreement.
```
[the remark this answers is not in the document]
leighmcculloch: Thanks! b129ba6

sdk/state/close.go
leighmcculloch: Great point. 8709d01.

sdk/state/open.go
leighmcculloch: 💯 130b932

sdk/state/hash.go
leighmcculloch: The `String` function is an implementation of the fmt.Stringer interface, so if this hash is formatted into a string, fmt.Print* functions can print a hex version rather than a byte array. It's just for improved debugging. https://pkg.go.dev/fmt#Stringer

sdk/state/state.go
leighmcculloch: 💯 I really like that idea, but for a couple reasons I deferred doing that. CloseAgreement is like this unit of data that needs to be communicated with the other participant. Everything in it needs communicating. If we bundle transactions in there, then there is peripheral information in there. So, if we are going to do thed single object I'd probably want to create a new object that wraps the CloseAgreement and the CloseAgreementTransactions. But doing so would require touching every line of code that touches the existing fields. That's a big refactor which I think we should do independently of adding a new feature (this PR). I'm up for following up on that to at least see what it would look like.

acharb: sgtm for following up with. I'm not a huge advocate for it, only thing I was thinking is it would be one way to help ensure the agreements and transactions stay in sync since separately stored now. There also might be other/better ways that become clearer later, or if it's even a problem at all

acharb (approved):
lgtm!

sdk/state/state.go
leighmcculloch: 💯 I think it is compelling for that reason too.

PR #295, opened 2021-09-04
sdk/agent: fix reconnect

leighmcculloch:
### What
Change the hello message handling so that multiple hellos are allowed as long as they contain the same data.

### Why
When restoring a channel using a snapshot the hello messages will be exchanged again for the new setup. The agent considers that an error. That should be allowed.


acharb (approved):
lgtm!

PR #297, opened 2021-09-07
sdk/agent: exit early if eof

leighmcculloch:
### What
Exit the agent's receive loop early if EOF reached on agent's connection.

### Why
If an EOF error is seen there's no point attempting to read the connection again as that is an indication the connection is closed. In the example app this is clear because when one side disconnects the agent continuously logs the EOF error.

### Known Limitations
I'm not adding a test for this change because the test would need to block on the receive loop on a timeout waiting to see if it stops. It'd be brittle and not deliver a ton of value.


acharb (approved):
lgtm!

PR #304, opened 2021-09-09
sdk/state: rename CloseAgreement(.+) to Close$1

leighmcculloch:
### What
Replace the `CloseAgreement` and `OpenAgreement` prefixes on types with simply `Close` and `Open`.

Rename `CloseAgreementSignatures` to `CloseSignatures`.
Rename `CloseAgreementTransactions` to `CloseTransactions`.

Rename `OpenAgreementSignatures` to `OpenSignatures`.
Rename `OpenAgreementTransactions` to `OpenTransactions`.

### Why
As a follow up of the storing transactions work (#294) and as part of the work to make it easy for examples and experiments to store transactions as they occur in the channel I plan to move the `CloseAgreementTransactions` and `CloseAgreement` into some new type. There are quite a few types nested the term 'Agreement' has become a little meaningless as it seems like the word shows up in every type. I think it would be helpful to remove it from everywhere except the types that is in fact the full agreement, which is the `CloseAgreement` type. In this way the `CloseAgreement` is made up of many parts, but the only thing that is the agreement, is the `CloseAgreement` itself.

For #302

### DIFF
--- a/sdk/state/close.go
+++ b/sdk/state/close.go
@@ -14,7 +14,7 @@ import (
 // transactions and still has them stored internally then it will return those
 // previously built transactions, otherwise the transactions will be built from
 // scratch.
-func (c *Channel) closeTxs(oad OpenDetails, d CloseDetails) (txs CloseTransactions, err error) {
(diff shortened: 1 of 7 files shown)


acharb (approved):
lgtm!

PR #305, opened 2021-09-09
sdk/state: encapsulate stored txs inside close agreements

leighmcculloch:
### What
Encapsulate stored transactions `CloseTransactions` inside `CloseAgreement`, and move most of what is already inside `CloseAgreement` inside `CloseEnvelope` which lives inside `CloseAgreement`.

#### Before
Untitled-2021-08-20-2126-4

#### After
Untitled-2021-08-20-2126-3

#### Example of Channel State

An example of an agreement in the `state.Channel` `Snapshot` rendered as JSON with this change:

```json
"LatestAuthorizedCloseAgreement": {
  "Envelope": {
    "Details": {
      "ObservationPeriodTime": 0,
      "ObservationPeriodLedgerGap": 0,
      "IterationNumber": 4,
      "Balance": 490000000,
      "PaymentAmount": 30000000,
      "ProposingSigner": "GBFUR3LXINTVW74QKRL7T5WSPABUJORLFOOH22PXGQCCWCAV3SHCHCQN",
      "ConfirmingSigner": "GB5NCO6I2ROUAF2V7CRIKXQPQ5B4RIOUACTXEARXJE3PNJTXYYTONEX3"
    },
    "ProposerSignatures": {
      "Close": "XGTZ/GzUEZf0vofMXun/R0fqBwrENGGJGev/dW3ZGHYE9q1r2eMAuGAUNdkFtM1QiNd9Hvu58imk3+i3x1f+AA==",
      "Declaration": "E45br2sVGPAkbNnyXJRzQVd3KrJh3713ScUyn0PX3Wu1dJ85DRRzopiuQiMKzVkB/4KQtLNQBEEv26vd0lo+BA=="
    },
    "ConfirmerSignatures": {
      "Close": "Eqa1tn2sM2HjFNhPxnqrp37aip67DfHsBEbpJBowj/CvWEsz6Aj+7XRIhbeHn2J6HQcD9oegN/QEdx+4UAtHCQ==",
      "Declaration": "Jysy/n8r2JGF8FFXIPrVG0SVgAe7+r0eqd5bzQfW+UEEkz0R+GlQtFu/bskHVEMRaeVomGwJT3tRdU1nHZ1nBg=="
    }
  },
  "Transactions": {
    "CloseHash": "7cf7360284b3695721bf1cd1fc3e775a00c85af46a21dab8fe9e88f13b2d2277",
    "Close": "AAAAAgAAAADsxhDF9xufXYYykz8gRn0i8jBEsO4BA5ItofiGd3/3uwAAAAAAAH3lAAAACgAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMAAAABAAAAAOzGEMX3G59dhjKTPyBGfSLyMESw7gEDki2h+IZ3f/e7AAAABQAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAQAAAAEAAAABAAAAAQAAAAEAAAABAAAAAAAAAAEAAAAAetE7yNRdQBdV+KKFXg+HQ8ih1ACncgI3STb2pnfGJuYAAAAAAAAAAQAAAAAD9R1A3o4QPCVHjsHdA8QZZOir5JzkYCs5+ENKX37YTwAAAAUAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAEAAAABAAAAAQAAAAEAAAABAAAAAQAAAAAAAAABAAAAAEtI7XdDZ1t/kFRX+fbSeANEuisrnH1p9zQEKwgV3I4jAAAAAAAAAAEAAAAA7MYQxfcbn12GMpM/IEZ9IvIwRLDuAQOSLaH4hnd/97sAAAABAAAAAAP1HUDejhA8JUeOwd0DxBlk6KvknORgKzn4Q0pffthPAAAAAAAAAAAdNM6AAAAAAAAAAAA=",
    "DeclarationHash": "cadbf9c6db7ce7f60c5dccacbd3f631064c891d8bd6cd5f1569e3df099e00f58",
    "Declaration": "AAAAAgAAAADsxhDF9xufXYYykz8gRn0i8jBEsO4BA5ItofiGd3/3uwAAAAAAAH3lAAAACQAAAAIAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAB95QAAAAEAAAAAAAAAAAAAAAAAAAABAAAAA3rRO8jUXUAXVfiihV4Ph0PIodQAp3ICN0k29qZ3xibmAAAAIHz3NgKEs2lXIb8c0fw+d1oAyFr0aiHauP6eiPE7LSJ3AAAAAAAAAAEAAAAAAAAACwAAAAAAAAAAAAAAAAAAAAA="
  }
},
```

### Why
At the moment we store transactions alongside the close agreements. Everywhere we set the close agreements we must also remember to store the transactions. @acharb raised the point in https://github.com/stellar/experimental-payment-channels/pull/294#discussion_r704825595 that it could be easy to make mistakes setting one and not the other, and so it could be safer to have them inside the same object.

We get other benefits of wrapping them in a single object because we can return the transactions whenever we return the agreement from functions like `ProposePayment`/`ConfirmPayment`, or in events that transmit the agreement.

This change does not simply move `CloseTransactions` inside the existing `CloseAgreement` because that would have the side-effect of the transactions being an input to `ConfirmPayment` and `ConfirmClose`. That's a problem because we'd be saying that the transactions are part of the data blob that the two participants agree on, which is unnecessary. That would be inconvenient because any input data should be validated, so we'd have to validate the transactions passed matches also. All this is unnecessary logic we don't need in the confirm functions.

I chose the name `CloseEnvelope` for the new name for the type that holds the `CloseDetails` and the `CloseSignatures` because it is generic enough to represent an envelope of data that is passed around, and it is also the term we use in the Stellar protocol for the outer most part of a transaction that gets passed around. It's not a perfect name, and it is easy for me to change with Go refactoring tools so if anyone has a better idea now I can rename it, or if we find a better name later I can rename it later easily.

For #302

### DIFF
--- a/examples/console/main.go
+++ b/examples/console/main.go
@@ -108,9 +108,9 @@ func run() error {
 			case agentpkg.OpenedEvent:
 				fmt.Fprintf(os.Stderr, "agent channel opened\n")
 			case agentpkg.PaymentReceivedEvent:
-				fmt.Fprintf(os.Stderr, "agent channel received payment: iteration=%d balance=%s\n", e.CloseAgreement.Details.IterationNumber, amount.StringFromInt64(e.CloseAgreement.Details.Balance))
+				fmt.Fprintf(os.Stderr, "agent channel received payment: iteration=%d balance=%s\n", e.CloseAgreement.Envelope.Details.IterationNumber, amount.StringFromInt64(e.CloseAgreement.Envelope.Details.Balance))
 			case agentpkg.PaymentSentEvent:
-				fmt.Fprintf(os.Stderr, "agent channel sent payment and other participant confirmed: iteration=%d balance=%s\n", e.CloseAgreement.Details.IterationNumber, amount.StringFromInt64(e.CloseAgreement.Details.Balance))
+				fmt.Fprintf(os.Stderr, "agent channel sent payment and other participant confirmed: iteration=%d balance=%s\n", e.CloseAgreement.Envelope.Details.IterationNumber, amount.StringFromInt64(e.CloseAgreement.Envelope.Details.Balance))
 			case agentpkg.ClosingEvent:
 				fmt.Fprintf(os.Stderr, "agent channel closing\n")
 			case agentpkg.ClosedEvent:
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -256,7 +256,7 @@ func (a *Agent) Open() error {
 	enc := msg.NewEncoder(io.MultiWriter(a.conn, a.logWriter))
 	err = enc.Encode(msg.Message{
 		Type:        msg.TypeOpenRequest,
-		OpenRequest: &open,
+		OpenRequest: &open.Envelope,
 	})
 	if err != nil {
 		return fmt.Errorf("sending open: %w", err)
@@ -304,7 +304,7 @@ func (a *Agent) Payment(paymentAmount string) error {
 	enc := msg.NewEncoder(io.MultiWriter(a.conn, a.logWriter))
 	err = enc.Encode(msg.Message{
 		Type:           msg.TypePaymentRequest,
-		PaymentRequest: &ca,
+		PaymentRequest: &ca.Envelope,
 	})
 	if err != nil {
 		return fmt.Errorf("sending payment: %w", err)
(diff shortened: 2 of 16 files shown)


acharb (approved):
lgtm!

PR #307, opened 2021-09-09
examples/console: support submitting older agreements for testing bad actors

### LINKED ISSUE #302: examples/console: add ability to close old tx for demoing

leighmcculloch:
### What
Add support for submitting older agreements to the console example.

### Why
To support testing and experimenting with what happens when bad actors or confused actors submit stale agreements.

Close #302


examples/console/main.go
```diff
@@ -109,8 +113,10 @@ func run() error {
 				fmt.Fprintf(os.Stderr, "agent channel opened\n")
 			case agentpkg.PaymentReceivedEvent:
 				fmt.Fprintf(os.Stderr, "agent channel received payment: iteration=%d balance=%s\n", e.CloseAgreement.Envelope.Details.IterationNumber, amount.StringFromInt64(e.CloseAgreement.Envelope.Details.Balance))
+				closeAgreements = append(closeAgreements, e.CloseAgreement)
 			case agentpkg.PaymentSentEvent:
 				fmt.Fprintf(os.Stderr, "agent channel sent payment and other participant confirmed: iteration=%d balance=%s\n", e.CloseAgreement.Envelope.Details.IterationNumber, amount.StringFromInt64(e.CloseAgreement.Envelope.Details.Balance))
+				closeAgreements = append(closeAgreements, e.CloseAgreement)
```
acharb, line 119: ❓  will this add the same close agreement added a couple lines up? If so what if we only include this line, so that only fully signed close agreements are stored?

examples/console/main.go
```diff
@@ -283,6 +291,28 @@ func prompt(agent *agentpkg.Agent, horizonClient horizonclient.ClientInterface,
 		return agent.DeclareClose()
 	case "close":
 		return agent.Close()
+	case "declarecloseidx":
+		idx, err := strconv.Atoi(params[1])
+		if err != nil {
+			return err
+		}
+		tx := closeAgreements[idx].SignedTransactions().Declaration
```
acharb, line 299: 💡 if idx is too large this will error, thoughts on throwing an error like `"given idx is too large. Please pass value smaller than %d," len(closeAgreements)`

acharb (commented):
awesome we're adding this 🙌 , had 2 comments inline

examples/console/main.go
leighmcculloch: +1 9a458fd. I also added another command `listagreements` that will list the agreements so it is easier to figure out which one to submit.

examples/console/main.go
leighmcculloch: The sender only gets the `PaymentSentEvent` after the payment is confirmed by the other side, and the receiver only gets the `PaymentReceivedEvent`, so there won't be a duplicate.

acharb: ah right 👍

acharb (approved):
lgtm!

PR #309, opened 2021-09-09
sdk/state: move logic for building close agreement from open agreement

leighmcculloch:
### What
Move the logic for building a CloseAgreement from an OpenAgreement from ConfirmOpen to a function on the OpenAgreement.

### Why
The ConfirmOpen function has a long piece of logic for mapping the open to its close. This logic can be owned by the OpenAgreement and the other open types. It moves the logic closer to where the data is defined. It keeps the ConfirmOpen more focused on what its responsibilities are.

Related to #302

### DIFF
--- a/sdk/state/open.go
+++ b/sdk/state/open.go
@@ -11,7 +11,7 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-type OpenDetails struct {
+type OpenAgreementDetails struct {
 	ObservationPeriodTime      time.Duration
 	ObservationPeriodLedgerGap int64
 	Asset                      Asset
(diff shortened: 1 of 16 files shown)


sdk/state/open.go
```diff
@@ -116,13 +125,42 @@ func (oa OpenEnvelope) SignaturesFor(signer *keypair.FromAddress) *OpenSignature
 	return nil
 }
 
+// CloseEnvelope gets the equivalent CloseEnvelope for this OpenEnvelope.
+func (oa OpenEnvelope) CloseEnvelope() CloseEnvelope {
```
acharb, line 129:
```suggestion
func (oe OpenEnvelope) CloseEnvelope() CloseEnvelope {
```

sdk/state/open.go
```diff
@@ -84,6 +84,15 @@ type OpenTransactions struct {
 	Formation       *txnbuild.Transaction
 }
 
+func (ot OpenTransactions) CloseTransactions() CloseTransactions {
```
acharb, line 87: ❓ should these be non-exported (right word?). Probably doesn't matter because someone can access these values from the CloseAgreement() method. But it would make it obvious that these methods are intended for internal use only

acharb (commented):
awesome change 🙌 . Had two comments

sdk/state/open.go
leighmcculloch: I think it's okay if they're exported. We're only exposing data that is already exported. Actually, in #310 I removed this function 😬. Sorry.

sdk/state/open.go
leighmcculloch: 👍🏻 0005475

acharb (approved):
lgtm!

PR #310, opened 2021-09-09
sdk/state: move getting signed txs for open to the OpenAgreement

leighmcculloch:
### What
Move where the logic for getting signed transactions for an open agreement to the OpenAgreement type rather than have it live in the OpenTx function.

### Why
Mostly for consistency with CloseAgreement which we changed to this model in #306. It doesn't have all the same benefits as CloseAgreement, but I think it is helpful if the agreements largely function the same way.

One thing that is different about OpenAgreement to CloseAgreement is it includes transactions for a CloseAgreement. We currently bundle those transactions in with the opening transaction (the formation transaction) but in some ways this is confusing because those transactions do not need to be submitted for opening. As part of this change I moved the close transactions out of the OpenTransactions type. This was necessary so that the `SignedTransactions` function could return only the signed transactions that we normally need when we are thinking about executing the open, so that someone is not confused and tries to execute all three.

Note that I didn't move things like the declaration and close transaction signatures out of the OpenSignatures. That's because the signatures for those transactions are 100% part of the open transaction because the open transaction has those signatures embedded within it.

Follow up of #302


acharb (approved):
lgtm!

Issue #251, opened 2021-08-25
sdk/state: ingest balances taking into account liabilities


leighmcculloch:
In #234 and #238 we started using ingested transactions and their result meta to update the internal state tracking the escrow balances. Something we didn't take into account is that trust lines have buying and selling liabilities. We should interpret the balance of the escrow account as the balance of the trust line minus its selling liabilities.

https://developers.stellar.org/docs/glossary/accounts/#liabilities

https://github.com/stellar/stellar-core/blob/c5f6349b240818f716617ca6e0f08d295a6fad9a/src/xdr/Stellar-ledger-entries.x#L77-L81

leighmcculloch:
> Something we didn't take into account is that trust lines have buying and selling liabilities.

Something I missed when writing this issue is that the native balance of an account also has buying and selling liabilities. So this issue applies to the native balance as well, not only trust lines.

PR #311, opened 2021-09-09
sdk/state: check for liabilities when ingesting transactions and updating balances

acharb:
**WHAT**
when ingesting a transaction and updating the escrow account balances, checks for liabilities to account for them

**WHY**
the escrow account balance is used to check against when proposing a payment, and if the liabilities lessen the available balance then it should account for that

closes https://github.com/stellar/experimental-payment-channels/issues/251


examples/console/GB5.json
```diff
@@ -0,0 +1,101 @@
+{
```
leighmcculloch, line 1: ❗ Looks like this file was committed accidentally.

sdk/state/ingest.go
```diff
@@ -162,7 +162,8 @@ func (c *Channel) ingestTxMetaToUpdateBalances(resultMetaXDR string) error {
 					continue
 				}
 				ledgerEntryAddress = tl.AccountId.Address()
-				ledgerEntryBalance = int64(tl.Balance)
+				liabilities := tl.Liabilities()
+				ledgerEntryBalance = int64(tl.Balance - liabilities.Selling)
```
leighmcculloch, line 166: :bulb: I think it might make sense to rename `ledgerEntryBalance` to `ledgerEntryAvailableBalance` to clarify that this is not just the balance of the ledger entry, but the available balance, the amount that can be spent.

leighmcculloch, line 166: ❗ One thing I missed when writing the original issue is that we need to take into account liabilities for native balances too. You could do that in this PR, or also in a separate PR.

leighmcculloch (commented):
Couple things (❗) and one suggestion (💡).

sdk/state/ingest.go
acharb: good point, added here
https://github.com/stellar/experimental-payment-channels/pull/311/commits/89aff901aa1a0a91aa7aef5f043ea85c3a7ff9c2

examples/console/GB5.json
acharb: yup removed
https://github.com/stellar/experimental-payment-channels/pull/311/commits/89aff901aa1a0a91aa7aef5f043ea85c3a7ff9c2

leighmcculloch (approved):
🎉

sdk/state/ingest.go
acharb: I agree
https://github.com/stellar/experimental-payment-channels/pull/311/commits/7f731d3bdd0b1a805cf0888fe9e82f7ab6dd7f72

leighmcculloch (commented):
:100:

Issue #267, opened 2021-08-30
sdk/state: ingestion allows returning to previous states


leighmcculloch:
I discovered while making change #266 that we allow the channel to skip states. It's possible to move from StateNone to StateClosing if the ingestion sees nothing but the declaration transaction.

I'm not sure if this is an issue because the only way a closing transaction is valid is if both participants have agreed on the open and the closing anyway, so if anything this results in updating a channel that has somehow missed ingesting earlier transactions.

@acharb @nikhilsaraf @accordeiro Thoughts?

leighmcculloch:
I came across an example where this is a problem, so I'm moving this issue into our current milestone.

Example: If declaration and closing transactions are submitted close together, because both participants agreed to close, they can be ingested out-of-order. There's no trivial way for us to order transactions being streamed in because we monitor both escrow accounts and the monitoring for both will see transactions that affect both accounts twice. We can try and do fancy things up the stack to reduce duplicates, but the state machine should be resilient to this I think.

I think we can ignore any transaction that would progress state backwards.

leighmcculloch:
I've temporarily mitigated this and other problems in #265 where I'm ingesting all txs on the network, not just ones relating to the escrow account, to ensure they arrive in order and we never process a duplicate tx.

leighmcculloch:
I renamed this issue to focus on ingestion allowing the channel to return to a previous state. That's actually the biggest issue here. Being able to skip states is almost a requirement because if we lose a transaction we should accept more recent transactions that might be informing us of a close or something important.

PR #312, opened 2021-09-09
sdk/state: dont let escrow account sequence number go backwards

acharb:
**WHAT**
if the channel ingests a transaction that would set the initiator escrow account seq num backwards, then return the ingestion process early and don't update state.

**WHY**
it can't be certain that transaction will be ingested in order in case two are submitted at the same time. So always just update the state from the transaction with the latest sequence number

closes https://github.com/stellar/experimental-payment-channels/issues/267

### DIFF
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -1,14 +1,11 @@
 package state
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/stellar/go/txnbuild"
 	"github.com/stellar/go/xdr"
 )
-
-var errOldSeqNum = errors.New("old sequence number")
 
 // IngestTx accepts any transaction that has been seen as successful or
 // unsuccessful on the network. The function updates the internal state of the
@@ -64,7 +61,6 @@ func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Tr
 	}
 
 	c.setInitiatorEscrowAccountSequence(tx.SourceAccount().Sequence)
-	return
 }
 
 // ingestTxToUpdateUnauthorizedCloseAgreement uses the signatures in the transaction to
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -55,6 +55,11 @@ func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Tr
 		return
 	}
 
+	// If the transaction is from an earlier sequence number, return.
+	if tx.SourceAccount().Sequence < c.initiatorEscrowAccount().SequenceNumber {
+		return
+	}
+
 	c.setInitiatorEscrowAccountSequence(tx.SourceAccount().Sequence)
 }
 
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -32,10 +32,10 @@ func (c *Channel) IngestTx(txXDR, resultXDR, resultMetaXDR string) error {
 	}
 
 	err = c.ingestTxToUpdateInitiatorEscrowAccountSequence(tx)
+	if errors.Is(err, errOldSeqNum) {
+		return nil
+	}
 	if err != nil {
-		if errors.Is(err, errOldSeqNum) {
-			return nil
-		}
 		return err
 	}
 
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -1,14 +1,11 @@
 package state
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/stellar/go/txnbuild"
 	"github.com/stellar/go/xdr"
 )
-
-var errOldSeqNum = errors.New("old sequence number")
 
 // IngestTx accepts any transaction that has been seen as successful or
 // unsuccessful on the network. The function updates the internal state of the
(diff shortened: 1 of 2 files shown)


sdk/state/ingest.go
```diff
@@ -28,6 +28,11 @@ func (c *Channel) IngestTx(txXDR, resultXDR, resultMetaXDR string) error {
 		return fmt.Errorf("transaction unrecognized")
 	}
 
+	// If the transaction is from an earlier sequence number, return.
+	if tx.SequenceNumber() < c.initiatorEscrowAccount().SequenceNumber {
+		return nil
+	}
```
leighmcculloch, line 34: I think the change might need to be a little more involved than this. This code assumes the only transactions we're ingesting are transactions sourced by the initiator escrow account. However, we also ingest transactions sourced by the responder escrow account.

leighmcculloch: We also need to still ingest transactions from elsewhere too, such as deposits.

acharb: ahh woops 😓   I had this before and changed it last second
https://github.com/stellar/experimental-payment-channels/pull/312/commits/0479e0ae2c368617ffaa875d931098ef55104688

this ends the ingestion process early if the tx is from the initiator escrow and an earlier seq num.

the channel.State() moving backwards is currently only caused by the seq num going backwards (or an error occurring). If we ingest a transaction sourced by the responder it wouldn't affect the channel State

sdk/state/ingest.go
```diff
@@ -28,7 +31,13 @@ func (c *Channel) IngestTx(txXDR, resultXDR, resultMetaXDR string) error {
...
 	}
 
-	c.ingestTxToUpdateInitiatorEscrowAccountSequence(tx)
+	err = c.ingestTxToUpdateInitiatorEscrowAccountSequence(tx)
+	if err != nil {
+		if errors.Is(err, errOldSeqNum) {
+			return nil
+		}
+		return err
+	}
```
leighmcculloch, line 40: 💡 FYI you can check for the error ahead of other errors, which avoid nesting and keeping the logic a bit flatter.
```suggestion
	err = c.ingestTxToUpdateInitiatorEscrowAccountSequence(tx)
	if errors.Is(err, errOldSeqNum) {
		return nil
	}
	if err != nil {
		return err
	}
```

sdk/state/ingest.go
acharb: it's actually weird that blocking the responder escrow account source didn't fail any tests, I'll add a test for that in a separate PR

leighmcculloch (approved):
One suggestion (💡), otherwise LGTM!

leighmcculloch:
I have another thought. It seems somewhat fraught with difficulty to support receiving transactions out of order. How about we say the IngestTx _must_ receive txs in order. And introduce a check that errors if that happens? I'm thinking about how we could introduce that check simply.

leighmcculloch:
> support receiving transactions out of order

I know previously we had this receiving transactions out of order, but now we're receiving them in order, or at least that's the way the Horizon ingester is implemented.

acharb:
it looks like `ingestTxMetaToUpdateBalances` is the only ingest method that can go backwards atm 🤔 . Maybe a change in there?

sdk/state/ingest.go
```diff
@@ -28,7 +31,13 @@ func (c *Channel) IngestTx(txXDR, resultXDR, resultMetaXDR string) error {
...
 	}
 
-	c.ingestTxToUpdateInitiatorEscrowAccountSequence(tx)
+	err = c.ingestTxToUpdateInitiatorEscrowAccountSequence(tx)
+	if errors.Is(err, errOldSeqNum) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
```
leighmcculloch, line 40: ❗We should keep processing the tx if it's an old seq num because we might want to ingest it as an unauthorized tx or some other future reason. Inside the ingest tx that updates seq num we should have it exit early if it is a lower number. 

So I think we need this change here:
```suggestion
	if err != nil {
		return err
	}
```

And another change below.

sdk/state/ingest.go
```diff
@@ -48,14 +57,20 @@ func (c *Channel) IngestTx(txXDR, resultXDR, resultMetaXDR string) error {
 	return nil
 }
 
-func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Transaction) {
+func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Transaction) error {
```
leighmcculloch, line 60:
```suggestion
func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Transaction) {
```

sdk/state/ingest.go
```diff
@@ -48,14 +57,20 @@ func (c *Channel) IngestTx(txXDR, resultXDR, resultMetaXDR string) error {
...
+		return nil
+	}
+
+	// If the transaction is from an earlier sequence number, return.
+	if tx.SourceAccount().Sequence < c.initiatorEscrowAccount().SequenceNumber {
+		return errOldSeqNum
 	}
 
 	c.setInitiatorEscrowAccountSequence(tx.SourceAccount().Sequence)
+	return nil
```
leighmcculloch, line 73:
```suggestion
		return
	}

	// If the transaction is from an earlier sequence number, return.
	if tx.SourceAccount().Sequence < c.initiatorEscrowAccount().SequenceNumber {
		return
 	}
 
 	c.setInitiatorEscrowAccountSequence(tx.SourceAccount().Sequence)
```

leighmcculloch (changes requested):
I like your suggestion to solve the backwards behavior separately for seq num and balances. I think for that reason we should contain the changes to inside each of those functions so they protect specific things but otherwise txs can still be processed out of order, which is actually very convenient. For that reason I think my precious comment was incorrect, and suggest the below changes instead.

sdk/state/ingest.go
acharb: sgtm 
https://github.com/stellar/experimental-payment-channels/pull/312/commits/642ef2c69a783310938bc1ded660feaa72853c67

sdk/state/ingest.go
```diff
@@ -55,6 +55,11 @@ func (c *Channel) ingestTxToUpdateInitiatorEscrowAccountSequence(tx *txnbuild.Tr
 		return
 	}
 
+	// If the transaction is from an earlier sequence number, return.
+	if tx.SourceAccount().Sequence < c.initiatorEscrowAccount().SequenceNumber {
```
leighmcculloch, line 59: 💡 This probably makes little difference, however technically we should also skip updating is it is equal, because it would be a duplicate transaction.
```suggestion
	if tx.SourceAccount().Sequence <= c.initiatorEscrowAccount().SequenceNumber {
```

leighmcculloch (approved):
👏🏻

Issue #282, opened 2021-09-02
sdk/state: ingest should ignore txs if no open agreement in progress or after close


leighmcculloch:
It appears that `state`'s `IngestTx` will ingest tx's before an open is negotiated and after a close is seen. I think this could cause some problems in that the channel value will continue to be updated when the channel conceptually no longer exists. I think we should early exit in `IngestTx` in the follow cases:

1. If `openAgreement.isEmpty()` (if there was some state between `StateNone` and `StateOpened` like `StateOpenProposed` we could use that, but we don't have that yet)
2. if `state == StateClosed`

PR #316, opened 2021-09-10
sdk/state: in IngestTx end early if channel is closed or hasn't been opened

acharb:
**WHAT**
end the IngestTx process early and return error if the channel hasnt been opened or has been closed

**WHY**
if the channel doesn't exist it shouldnt be updating its state

closes https://github.com/stellar/experimental-payment-channels/issues/282

### DIFF
--- a/sdk/state/ingest.go
+++ b/sdk/state/ingest.go
@@ -11,8 +11,19 @@ import (
 // unsuccessful on the network. The function updates the internal state of the
 // channel if the transaction relates to the channel.
 func (c *Channel) IngestTx(txXDR, resultXDR, resultMetaXDR string) error {
-	// TODO: Use the transaction result to affect on success/failure.
+	// If channel has not been opened or has been closed, return.
+	if c.OpenAgreement().isEmpty() {
+		return fmt.Errorf("channel has not been opened")
+	}
+	cs, err := c.State()
+	if err != nil {
+		return fmt.Errorf("getting channel state: %w", err)
+	}
+	if cs == StateClosed {
+		return fmt.Errorf("channel has been closed")
+	}
 
+	// Get transaction object from the transaction XDR.
 	gtx, err := txnbuild.TransactionFromXDR(txXDR)
 	if err != nil {
 		return fmt.Errorf("parsing transaction xdr")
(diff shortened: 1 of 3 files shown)


sdk/state/ingest.go
```diff
@@ -11,8 +11,19 @@ import (
...
+	// If channel has not been opened or has been closed, return.
+	if c.OpenAgreement().isEmpty() {
+		return fmt.Errorf("channel has not been opened")
+	}
+	cs, err := c.State()
+	if err != nil {
+		return fmt.Errorf("getting channel state: %w", err)
+	}
+	if cs == StateClosed {
+		return fmt.Errorf("channel has been closed")
```
acharb, line 23: fyi: the ticket says to end early but if the channel isn't opened I think it should be a user error, since they're trying to ingest at a time they shouldn't be and should be told why

sdk/state/ingest_test.go
```diff
@@ -1033,35 +1033,48 @@ func TestChannel_IngestTx_seqNumCantGoBackwards(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, StateOpen, cs)
 	}
```
acharb, line 1035: had to refactor this test a bit testing the state doesn't move backwards because of this PR change

sdk/state/state.go
```diff
@@ -127,8 +127,8 @@ const (
 	StateError State = iota - 1
 	StateNone
 	StateOpen
```
acharb, line 129: when fixing the test I realized these actually were out of order, according to how the state is returned by the sequence number value, `StateClosingWithOutdatedState` has an earlier seqNum than `StateClosing`

sdk/state/ingest.go
leighmcculloch: I'm curious if this error causes the console example to error unexpectedly and therefore if we need to handle errors differently further up the stack. The reason I'm wondering this is because ingestion happens on another go routine I think we will call IngestTx when we're not supposed to. It might be all fine.

It doesn't have to occur in this PR, but I think there could be some value in us having a sentinel error (var ErrChannelClosed = errors.New("channel closed")` or an error type `type ErrChannelClosed` that we return in all places that the channel is being accessed in a way it can't be. The error could be `ErrChannelNotOpen` so it could be used pre-open and post-close.

sdk/state/state.go
leighmcculloch: Ah nice, makes sense 👍🏻.

leighmcculloch (approved):
Looks good. One question about example app, otherwise great!

acharb:
> I'm curious if this error causes the console example to error unexpectedly and therefore if we need to handle errors differently further up the stack. The reason I'm wondering this is because ingestion happens on another go routine I think we will call IngestTx when we're not supposed to. It might be all fine.

I just ran through the example with opening, couple of payments, then closing. Didn't see anything out of ordinary, I might not have tested the case correctly though. Going to merge for now and then can make that fix you mentioned if it is breaking afterward

PR #318, opened 2021-09-10
examples/benchmark: one-way benchmark

### LINKED ISSUE #40: examples/benchmarks: one-way payments

leighmcculloch:
### What
Add a client that benchmarks a bidirectional payment channel that has payments going in one direction at a time.

### Why
To learn where the bottlenecks are in both the implementation and the design.

Close #40


examples/benchmark/main.go
```diff
@@ -0,0 +1,202 @@
...
+				fmt.Fprintf(os.Stderr, "opened\n")
+				if connectAddr != "" {
+					tick := time.Tick(1 * time.Second)
+					for i := 5; i >= 0; {
+						fmt.Fprintf(os.Stderr, "%d\n", i)
+						i--
+						if i >= 0 {
+							<-tick
+						}
+					}
```
acharb, line 153: ❓ a little confused by this part. Is this just counting down from 5?

acharb (commented):
lgtm, just had 1 question, and 1 thought below

💡  eventually it may be a good idea to add a readme in this folder (or add to TLD readme) to describe how someone can test for themselves the numbers we come up with. But that can probably be added later after more of the benchmarking tests are done so we have a better idea what to write

examples/benchmark/main.go
leighmcculloch: It's just a countdown at the beginning to make sure both ends are setup. There's an issue where one side may not be ready and have ingested the open and updated their internal state. Rather than add code to handle that more gracefully I added this delay.

acharb (approved):
lgtm!

PR #319, opened 2021-09-10
sdk/state: make safe to ingest txs out of order

### LINKED ISSUE #314: sdk/state: ignore older transactions in ingestTxMetaToUpdateBalances
right now as part of the ingest process in `ingestTxMetaToUpdateBalances` we update the balance from any transaction. If the channel ingests an older transaction then it will accept that balance as the newest balance. So we should figure out a way to only update balance from newer transactions and ignore older ones

possible solution (leigh suggestion) is storing the transaction order ID from Horizon alongside the balance to compare with

leighmcculloch:
### What
Make it safe to ingest transactions from the different escrow accounts out of order, specifically in regards to the balances.

### Why
This is a follow up to #312. To make ingestion more resilient. To prevent the channel from accidentally ingesting old state and moving the balance cached within the channel back to a previous value. This can happen because updating balances uses the snapshot of the account balance in the result meta.

As it happens this should also bring the ingestion to the point where transactions for the two escrow accounts can be ingested out of order, which should allow us to stream payments for only those two accounts again. I'll open a follow up PR to do that.

Close #314


sdk/horizon/streamer.go
```diff
@@ -60,12 +61,21 @@ func (h *Streamer) streamTx(cursor string, txs chan<- agent.StreamedTransaction,
...
+				return
+			}
 			cursor = tx.PagingToken()
 			streamedTx := agent.StreamedTransaction{
-				Cursor:         cursor,
-				TransactionXDR: tx.EnvelopeXdr,
-				ResultXDR:      tx.ResultXdr,
-				ResultMetaXDR:  tx.ResultMetaXdr,
+				Cursor:             cursor,
+				TransactionOrderID: txOrderedID,
```
acharb, line 75: 💭 can we make them all `OrderID` instead of `OrderedID`? For consistency, and I think `Order` makes more sense than `Ordered`

sdk/horizon/streamer.go
```diff
@@ -60,12 +61,21 @@ func (h *Streamer) streamTx(cursor string, txs chan<- agent.StreamedTransaction,
...
 		err := h.HorizonClient.StreamTransactions(ctx, req, func(tx horizon.Transaction) {
+			txOrderedID, err := strconv.ParseInt(tx.PagingToken(), 10, 64)
+			if err != nil {
+				ctxCancel()
+				if h.ErrorHandler != nil {
+					h.ErrorHandler(err)
+				}
+				return
+			}
 			cursor = tx.PagingToken()
```
acharb, line 72: 💡  so it's clear these values are related/coming from the same source
```suggestion
			cursor = tx.PagingToken()
			txOrderedID, err := strconv.ParseInt(cursor, 10, 64)
			if err != nil {
				ctxCancel()
				if h.ErrorHandler != nil {
					h.ErrorHandler(err)
				}
				return
			}
```

acharb: ❓ for my own understanding, what's the difference between the the `cursor` and the `txOrderedID` ? ie. why can't we just save the cursor on the `escrowAccount.lastSeenTransactionOrderID`?

acharb (commented):
looks great 💯 , left some comments and a question

sdk/horizon/streamer.go
leighmcculloch: They just happen to come from the same field in Horizon's response, but they are conceptually completely different things. The cursor is the identifier used to resume paging. The transaction order ID is an identifier that is unique for each transaction and guarantee to order transactions as they are processed on network.

I made a change not quite what you had in 410beb6, but similar.

sdk/horizon/streamer.go
leighmcculloch: Good call, thanks for picking up on the inconsistency. 9ce4077

acharb (approved):
lgtm!

Issue #308, opened 2021-09-09
sdk/state: when channel closes at stale state the channel does not update to closed state


leighmcculloch:
When testing with the example console application manually and running through the scenario where a participant closes using stale data using #302, I saw that the channel of both participants update to the `ClosingWithStaleState` after the declaration transaction is ingested, but it did not change state when the close transaction was ingested.

```
> declarecloseidx 2
> ingesting cursor: 147592256163840 tx: 48e0bfaaa2fac9bb64b134bb7a3309445d9460d6cedca775e279b2d8debe0507
state before: 1
state after: 3
writing event: 3

> closeidx 2
> ingesting cursor: 147794119626752 tx: 1e5e89232907ea2d75ad5b0fa19b26b848ed9016c99307ace7a65283c3c0b134
state before: 3
state after: 3
```

State `3` is the `ClosingWithStaleState`. I expected to see state `4` which is the `Closed` state to occur after the close transaction was ingested, which was the `1e5e89` transaction.

acharb:
this is because the state depends on the sequence num, which never reaches the latest declTx or closeTx seqNums.

@leighmcculloch I think the best way to handle is adding a flag like `c.channelClosed` that gets set to true in `IngestTx` if there isn't both signers on the initiator Escrow. And then from that flag we can better interpret the channel state. thoughts?

🤔 we can also add the state: `StateClosedWithOutdatedState` (or `StateClosedPrematurely`). I think it adds more info for the user at the expense of some added complexity in the logic (can't just say StateClosed if `c.channelClosed`)

leighmcculloch:
I think what we can do is, if the channel is open, and seq num is less than latest closing, we could mod the seq to find out whether it is the closing/closed tx. Theoretically we could do that for the latest closing/closed too, but I think when you were working on that logic I suggested you use the stored value inside the transactions and check for exact equality. Maybe that was short-sighted of me, because we don't have all txns stored for earlier revisions to do the same thing. I think we could change to the approach I think you had suggested previously, checking what type of tx the seq would be.

Wdyt?

> 🤔 we can also add the state: `StateClosedWithOutdatedState` (or `StateClosedPrematurely`).

+1. I think this is worthwhile.

acharb:
sgtm 👍  seems simpler. My concern with `mod` initially was what if EI after closing on an earlier closeTx then bumped its sequence +1, then the channel would ingest and say channel is back open.

[...]

PR #324, opened 2021-09-13
sdk/state: new state StateClosedWithOutdatedState

acharb:
**WHAT**
if the seq num gets moved to a seq corresponding to an early close transaction (close transaction other than the latestAuthorized one) then the channel state will return a new state `StateClosedWithOutdatedState`, and close the channel from ingesting new transactions

**WHY**
currently the channel stays on `StateClosingWithOutdatedState` if we've ingested an earlier closeTx. The channel state should return closed since that's what occurred on the network. Also we should block new transactions from being ingested so that EI submitting new transactions does not interfere with the channel state.

closes https://github.com/stellar/experimental-payment-channels/issues/308


sdk/state/state.go
```diff
@@ -161,11 +162,14 @@ func (c *Channel) State() (State, error) {
...
 	initiatorEscrowSeqNum := c.initiatorEscrowAccount().SequenceNumber
+	s := c.openAgreement.Envelope.Details.StartingSequence
 
-	if initiatorEscrowSeqNum == c.openAgreement.Envelope.Details.StartingSequence {
+	if initiatorEscrowSeqNum == s {
 		return StateOpen, nil
-	} else if initiatorEscrowSeqNum < latestDeclSequence {
+	} else if initiatorEscrowSeqNum < latestDeclSequence && initiatorEscrowSeqNum%2 == s%2 {
 		return StateClosingWithOutdatedState, nil
+	} else if initiatorEscrowSeqNum < latestDeclSequence && initiatorEscrowSeqNum%2 != s%2 {
```
leighmcculloch, line 171: 💡 I think these operations would fit well as utility functions in the `sdk/txbuild` package.

I'm thinking:

```go
type TransactionType string

const (
	TransactionTypeDeclaration TransactionType = "declaration"
	TransactionTypeClose       TransactionType = "close"
)

func SequenceNumberToTransactionType(startingSeqNum, seqNum int64) TransactionType {
	// ...
}
```

Potentially we could do a small refactor on where we set the sequence numbers in the in the other txbuild functions, to call something like:
```go
func SequenceNumberFromTransactionType(startingSeqNum int64, tt TransactionType) int64  {
	// ...
}
```

leighmcculloch (approved):
Logic looks great 👍🏻. I think as a follow up this functionality could be refactored inside of txbuild, that would continue to keep logic about sequence numbers in that package.

sdk/state/state.go
acharb: I like it https://github.com/stellar/experimental-payment-channels/pull/324/commits/cbcd5416e68d803cbf08ab73d76c4e8b5fcd5d49

second suggestion sgtm, I can do in a separate PR

sdk/txbuild/sequence.go
```diff
@@ -5,3 +5,17 @@ const m = 2
...
+	TransactionTypeDeclaration TransactionType = "declaration"
+	TransactionTypeClose       TransactionType = "close"
+)
+
+func SequenceNumberToTransactionType(startingSeqNum, seqNum int64) TransactionType {
+	if startingSeqNum%m == seqNum%m {
+		return TransactionTypeDeclaration
+	}
+	return TransactionTypeClose
+}
```
leighmcculloch, line 21: Could you add some tests that demonstrate even and odd starting sequences, and then both tx types? 😄

acharb: that's a good idea https://github.com/stellar/experimental-payment-channels/pull/324/commits/c1426f119d29d722dc892b2e157685af61b5937d

writing the test made me think that we should add a `TransactionTypeOpen` (if they're the same)

also there's 1 unique case of the startingSequence + 1, which isn't a transaction type. Maybe we should add a type for that as well (it would be something like TransactionTypeNone)

acharb: added them here
https://github.com/stellar/experimental-payment-channels/pull/324/commits/d5e1329d027b68d5ba9745db8ebd74415bf77f2a

leighmcculloch: I wonder if we can focus on the transaction types being for close agreements then, because I don't think we get much bang out of adding other types.

I wouldn't worry about adding the none case either, we don't sign txs with those sequence numbers. I'd just panic in that case, it's code that should never be hit.

sdk/txbuild/sequence_test.go
```diff
@@ -0,0 +1,21 @@
+package txbuild
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_sequence(t *testing.T) {
```
leighmcculloch, line 9: 💡 
```suggestion
func TestSequenceNumberToTransactionType(t *testing.T) {
```

sdk/txbuild/sequence.go
acharb: added the panic https://github.com/stellar/experimental-payment-channels/pull/324/commits/0b1d1df7aaa8b51a4eef16e4d891caf4c5612fd8

imo I think the `TransactionTypeFormation` makes sense, isn't used now but may save a bug down the line if someone means to get the formation tx seq and not the declaration tx seq 🤷 

up to you though I can remove if it's added code for no reason

sdk/txbuild/sequence_test.go
acharb: https://github.com/stellar/experimental-payment-channels/pull/324/commits/84a3348c2823663d17471dba3e74176a89c09cc2

leighmcculloch (approved):

PR #328, opened 2021-09-14
sdk/state: avoid rehashing txs and remove cmp usage

### LINKED ISSUE #327: sdk/state: avoid rehashing txs we already have the hash for
Benchmarking in #318 showed that we spend sometime rehashing txs that we already have the hash for. We rehash the txs on every sign and verify which is unnecessary.

leighmcculloch:
### What
Avoid rehashing transactions.

### Why
We rehash transactions during signing and verification. It's pretty expensive because full transaction has to be serialized. We hold hashes of the transactions next to the transactions and so there's little value in rehashing.

I didn't add any benchmark tests because even without benchmarks this is a pretty clear win that we wouldn't recalculate the hash when it is literally sitting next to the transaction in the same struct.

Close #327


acharb (approved):
lgtm!

PR #329, opened 2021-09-14
sdk/state: remove use of google/go-cmp package

### LINKED ISSUE #326: sdk/state: remove use of google/go-cmp
The google/go-cmp package is slower than doing equality checks manually. Our equality checks aren't that complicated either to write by hand. Early benchmarking in #318 show we spend sometime in the cmp package. We planned to replace it anyway, just moving up the priority.

leighmcculloch:
### What
Remove use of the google/go-cmp package for checking the equality of complex types.

### Why
Benchmarking in #318 has shown that we spend time in the go-cmp package, and there's some interesting things also in the profiles showing a go routine running for detecting data races.

Close #326

### DIFF
--- a/sdk/state/payment_test.go
+++ b/sdk/state/payment_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	fuzz "github.com/google/gofuzz"
 	"github.com/stellar/experimental-payment-channels/sdk/txbuildtest"
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/network"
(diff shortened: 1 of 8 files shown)


sdk/state/payment_test.go
```diff
@@ -13,195 +14,97 @@ import (
...
-	for i, tc := range testCases {
+}
+
+func TestCloseEnvelope_Equal(t *testing.T) {
+	assert.True(t, CloseEnvelope{}.Equal(CloseEnvelope{}))
+
+	// The same value should be equal. It's common for CloseEnvelopes to start
+	// with details then have signatures added, so we check that pattern of
+	// incrementally adding fields.
+	f := fuzz.New().NilChance(0)
```
acharb, line 82: 💭  in some spots we're using `fuzz.New()` and in others `fuzz.NewWithSeed()`. Can we keep them the same for consistency?

according to [here](https://github.com/google/gofuzz/blob/f12f061189cca81b834eba9985016bc70e1dce2e/fuzz.go#L52) New() implicitly calls NewWithSeed with `time.Now().UnixNano()`, so maybe only using New()?

acharb (commented):
til about the fuzz package 💯 , just had 1 thought

sdk/state/payment_test.go
leighmcculloch: The places I'm using `NewWithSeed()` I'm using a common seed between multiple fuzzers. This means the data the fuzzer uses is identical. This means I can construct two `OpenDetails` that are entirely independent objects, even internal pointers are different, however they should be the same in terms of equality and the Equal function.

The places I'm using `New()` I don't need any consistency between the fuzzers, or rather I want inconsistency. I could use `NewWithSeed()` in those places and manually pass in different times, but that seems unnecessary?

acharb: ah that makes sense, thx for explaining. Agreed how it is lgtm

acharb (approved):
lgtm!

PR #332, opened 2021-09-22
sdk/state: concurrently verify and sign

leighmcculloch:
### What
Concurrently verify and sign transaction ed25519 signatures.

### Why
From the work benchmarking in #40 there are significant gains running verifications concurrently. The signing concurrently is less beneficial as it has a smaller gain.


sdk/state/payment.go
```diff
@@ -43,33 +44,28 @@ type CloseSignatures struct {
 	Declaration xdr.Signature
 }
 
+func (cas CloseSignatures) Set() bool {
```
acharb, line 47: 💡 `Set()` is slightly confusing to me. It seems like an action of setting something rather than getting information. What about `HasAllSignatures` like used in this [PR](https://github.com/stellar/experimental-payment-channels/pull/335)?

acharb: also I know not part of this PR but could we change the[ line here](https://github.com/stellar/experimental-payment-channels/blob/afeeaf6b8b2aab887c1a05a52d389a0b6cab1c79/sdk/state/open.go#L41) to use `!= 0` as well for consistency?

acharb (commented):
👍  just had 1 💡

sdk/state/payment.go
leighmcculloch: I agree, Set, is confusing. In this case we don't want all sigs, just any value. I'll change it Empty, like the other place we use Empty. Longer term maybe we should use pointers for these types. I regret going the other way now 😄 

Yup we can change the 0 condition check. I'll do that separately.

leighmcculloch: c824d59

leighmcculloch: I also fixed the 0 condition check too in 3016ad6.

acharb (approved):
lgtm!

sdk/state/payment.go
leighmcculloch: Addressing the 0 thing in https://github.com/stellar/experimental-payment-channels/pull/337.

PR #334, opened 2021-09-24
sdk/state: prevent opens once open begins

leighmcculloch:
### What
Don't allow a new open to be proposed if an open has already been proposed.

### Why
The logic for validating whether it is safe to propose a new open checks whether the open is complete and has all signatures, but this would make it possible for someone to propose an open, share it with a counterparty, then propose another open. That second proposal would discard the first open that was proposed. This could be unsafe because the other participant would hold transactions that could be valid that the proposer no longer has any record.


acharb (commented):
❓  my initial thought when adding that check was to allow a channel to re-propose an open in case the other party doesn't like the first open. And even though the other party has txs they can submit, they would only be able to close the channel before any payments. 

But I guess it could be the responsibility of the agent to reset the channel if the open isn't accepted?

leighmcculloch:
For the situation you describe we could have a better story for sure, but the problem with implementing that by allowing subsequent proposals is the state machine doesn't protect the user. If I propose a new open the SDK discards the current open. If I no longer have the current open, I can't close the channel if the other participant opens it.

To implement the situation you describe safely we'd need to bump the sequence on the escrow account. If Stellar had deterministic account creation, where we knew the sequence number of the new account, then we could implement this simpler for sure where the open contained a new random escrow account. i.e. I think we need to think about this situation more.

acharb:
>  If I no longer have the current open, I can't close the channel if the other participant opens it.

good point

acharb (approved):
lgtm!

PR #335, opened 2021-09-24
sdk/state: rename some functions for clarity

leighmcculloch:
### What
Rename any `isEmpty` functions to `Empty`. Rename any `isFull` functions to `HasAllSignatures`.

### Why
When I was making the change for #334 it wasn't obvious to me what the `isFull` function meant. This has happened to me before when I've forgotten what it meant and I had to go look at it to check. I think by renaming it to `HasAllSignatures` we'll make the intent of the function clearer. I also think the function can be exported. I also renamed `isEmpty` to `Empty` because I think it could equally be exported.


acharb (approved):
lgtm!

PR #336, opened 2021-09-24
all: update stellar/go version to include xdr generated with xdrgen#65

leighmcculloch:
### What
Update the version of stellar/go used in this repo to include the new commits that contain the xdr generated with xdrgen#65.

### Why
xdrgen#65 improves the performance of xdr encoding.

For #40


acharb (approved):
lgtm

PR #343, opened 2021-09-29
sdk/state: allow proposing payments with zero amounts

### LINKED ISSUE #341: sdk/state: allow proposing payments with zero amounts
The SDK prevents proposing payments that are zero amounts, but from best I can tell allows the receiver to accept a zero amount. It's not really necessary that we prevent zero payments and they can be a useful primitive for generating a new iteration for the current agreement.

This would make merging channels easier, and implementing turn taking buffered channels easier also.

leighmcculloch:
### What
Allow proposing payments with zero amounts.

### Why
The SDK prevents proposing payments that are zero amounts, but from best I can tell allows the receiver to accept a zero amount. It's not really necessary that we prevent zero payments and they can be a useful primitive for generating a new iteration for the current agreement.

This would make merging channels easier, and implementing turn taking buffered channels easier also.

Close #341

### DIFF
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -147,8 +147,8 @@ func (ca CloseAgreement) SignedTransactions() CloseTransactions {
 }
 
 func (c *Channel) ProposePayment(amount int64) (CloseAgreement, error) {
-	if amount <= 0 {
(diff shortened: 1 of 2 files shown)


acharb (approved):
lgtm!

PR #345, opened 2021-09-29
sdk/agent: keep amounts as int64

### LINKED ISSUE #344: sdk/agent: keep amounts as int64
The agent currently has amounts be strings. This is inconvenient for middlelayer code, such as the code required for buffering payments, that needs to wrap an agent and sit between application code and the agent. We use int64s in most place and it isn't a heavy lift for consumers of this package to use them either. For the sake of simplicity I think this is worth it.

leighmcculloch:
### What
Use int64 for amounts.

### Why
The agent currently has amounts be strings. This is inconvenient for middlelayer code, such as the code required for buffering payments, that needs to wrap an agent and sit between application code and the agent. We use int64s in most place and it isn't a heavy lift for consumers of this package to use them either. For the sake of simplicity I think this is worth it.

Close #344

### DIFF
--- a/.github/workflows/examples-benchmark.yml
+++ b/.github/workflows/examples-benchmark.yml
@@ -1,0 +1,41 @@
+name: 'Examples: Benchmark'
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+
+  lint:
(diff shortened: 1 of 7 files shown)


acharb (approved):
lgtm!

Issue #339, opened 2021-09-28
sdk/state: give close agreements identifiers separate to their iteration number


leighmcculloch:
I think we need to give close agreements identifiers that are separate to their iteration number. The iteration number is related to where in the history of the channel the close agreement exists. The value is an indicator of the state of the channel, from the proposer's perspective, that came prior but it doesn't provide any uniqueness. As part of the problem described in #287 it is likely that we need some way to uniquely identifier payments. This is also something @nikhilsaraf called our early that we need a unique identifier for a payment, based off what he saw in research of other payment channel designs.

Given that the close agreement is ultimately rendered as a set of transactions, I think the simplest method for constructing an identifier would be to concatenate the transaction hashes from the transaction set in a predefined order, e.g. declaration then closing, using the same hash function that Stellar uses for its transaction hashes, i.e. SHA-256. The result will be unique not only for an iteration number but for the agreement captured in the transactions that get built.

Alternatively we could construct a payload to hash off the close details, but that seems like more work with no benefits that I can see, other than we could include peripheral information like payment amount that aren't captured in the transactions.

Refs:
- https://github.com/stellar/experimental-payment-channels/discussions/287

PR #348, opened 2021-09-30
sdk/state, sdk/agent: add memos to payments

### LINKED ISSUE #347: sdk/state, sdk/agent: add memos to payments
Payments should have memos so they can be identifiable on the terms of the participants. Participants may communicate about a payment ahead of sending a payment and the memo will provide a way to annotate or identify the payment in whatever way they require.

Note that this is similar but different to #339. #339 is concerned with uniquely identifying payments with an identifier that the state machine attaches to close agreements. That would not be useful in the situation described above where an identifier must be assigned prior to the payment being made. 

This is useful for things like buffered channels, something I'm experimenting with right now for #340.

Refs:
- #339
- #340

leighmcculloch:
### What

Add a memo field to close agreements that is optional. The memo is a string with unspecified bounds. The memo is not included in the final transaction so its contents does not leak to the public ledger when the channel is closed. Subsequently the signatures for close agreements to do not cover the memo.

### Why

Payments should have memos so they can be identifiable on the terms of the participants. Participants may communicate about a payment ahead of sending a payment and the memo will provide a way to annotate or identify the payment in whatever way they require.

Note that this is similar but different to #339. #339 is concerned with uniquely identifying payments with an identifier that the state machine attaches to close agreements. That would not be useful in the situation described above where an identifier must be assigned prior to the payment being made.

This is useful for things like buffered channels, something I'm experimenting with right now for #340.

Close #347

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -276,6 +276,15 @@ func (a *Agent) Open() error {
 // The payment is not authorized until the remote participant signs the payment
 // and returns the payment.
 func (a *Agent) Payment(paymentAmount int64) error {
+	return a.PaymentWithMemo(paymentAmount, "")
+}
+
+// Payment makes a payment of the payment amount to the remote participant using
+// the open channel. The process is asynchronous and the function returns
+// immediately after the payment is signed and sent to the remote participant.
+// The payment is not authorized until the remote participant signs the payment
+// and returns the payment. The memo is attached to the payment.
+func (a *Agent) PaymentWithMemo(paymentAmount int64, memo string) error {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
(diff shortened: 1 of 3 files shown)


sdk/agent/agent.go
```diff
@@ -276,6 +276,15 @@ func (a *Agent) Open() error {
 // The payment is not authorized until the remote participant signs the payment
 // and returns the payment.
 func (a *Agent) Payment(paymentAmount int64) error {
+	return a.PaymentWithMemo(paymentAmount, "")
+}
+
+// Payment makes a payment of the payment amount to the remote participant using
```
acharb, line 282: 💡 can we remove the dup method comment from `Payment()` above? Since we're re-explaining what the method does here. The Payment comment above could be something like:
```
// Payment creates a payment using the PaymentWithmemo method with an empty memo.
func (a *Agent) Payment(paymentAmount int64) error {
```

acharb (approved):
lgtm! had 1 💡  but non-blocking

sdk/agent/agent.go
leighmcculloch: #351

PR #352, opened 2021-10-06
all: pin golangci-lint action version

leighmcculloch:
### What
Pin the version of the golangci-lint action to a specific commit.

### Why
So that we don't introduce new action code into our builds without us knowing about it and choosing to explicitly do that.

It is a good practice noted here: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions


acharb (approved):
good to know, lgtm!

PR #353, opened 2021-10-07
examples: add a lightweight dashboard with http endpoints to agents

leighmcculloch:
### What
Add simple HTTP endpoint to agents for getting the agents current state. Use the HTTP endpoint in a very lightweight dashboard.

https://user-images.githubusercontent.com/351529/136469717-7a6e1231-b4f3-4f08-a776-8066ec36f242.mp4

### Why
To make it easier to see what is going on during demos.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -197,7 +197,7 @@ func (a *Agent) Snapshot() Snapshot {
 	return a.buildSnapshot()
 }
 
-func (a *Agent) snapshot() {
+func (a *Agent) takeSnapshot() {
 	if a.snapshotter == nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


sdk/agent/agent.go
```diff
@@ -167,10 +167,45 @@ type Agent struct {
...
+}
+
+// Snapshot returns a snapshot of the agent and its channel.
+func (a *Agent) Snapshot() Snapshot {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.buildSnapshot()
+}
+
 func (a *Agent) snapshot() {
```
acharb, line 200: 💡 can we change the unexported snapshot name to something like `takeSnapshot` or `saveSnapsot`? Since the exported (which seems like a getter) has the same name but is doing something different

PR #354, opened 2021-10-08
examples/bufferedbenchmark: variety of improvements

leighmcculloch:
### What
A variety of changes to the bufferedbenchmark:
- Added a max queue size that limits how big the queue can get.
- Added a wait function that allows the calling code to wait until the queue is empty and idle.
- Added cli parameters for controlling the max queue size and the test size.
- Changed the internal to use the same patterns employed in stellar/go#3980

### Why
Adding a max queue size lets us run benchmarks with a variety of queue sizes to see how that impacts performance. Large queues use more memory, which affects application performance, so testing this is valuable.

Having a state where the queue is full requires that we have a way to find out when it is available again. Adding the wait functionality required restructuring the internals of how the signaling worked, and I used the same pattern that was used in stellar/go#3980 because I think it was easier to extend.

### DIFF
--- a/sdk/agent/bufferedagent/agent.go
+++ b/sdk/agent/bufferedagent/agent.go
@@ -8,6 +8,7 @@
 package bufferedagent
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"sync"
@@ -16,9 +17,13 @@ import (
 	"github.com/stellar/experimental-payment-channels/sdk/agent"
 )
 
+var ErrQueueFull = errors.New("queue full")
+
 type Config struct {
 	Agent       *agent.Agent
 	AgentEvents <-chan agent.Event
+
+	MaxQueueSize int
 
 	LogWriter io.Writer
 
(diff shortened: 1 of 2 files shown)


sdk/agent/bufferedagent/agent.go
```diff
@@ -78,66 +87,55 @@ func (a *Agent) Open() error {
...
 // identifier for the settlement is returned.
-func (a *Agent) Payment(paymentAmount int64) (settlementID string) {
+func (a *Agent) Payment(paymentAmount int64) (queueID string, err error) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
+	if a.maxQueueSize != 0 && len(a.queue) == a.maxQueueSize {
+		return "", ErrQueueFull
+	}
+	if paymentAmount > math.MaxInt64-a.queueTotalAmount {
+		return "", ErrQueueFull
```
acharb, line 97: 💡  this is technically a different error right? Might be deceiving if a test hits this limit and this error gets returned. What about a custom error like `max payment amount reached`

PR #358, opened 2021-10-08
sdk/agent: remove non-idiomatic event type

leighmcculloch:
### What
Remove the agent.Event interface type.

### Why
It is not very Go-like to define an interface for identification. Interfaces should be defined by consumers.

### DIFF
--- a/examples/benchmark/main.go
+++ b/examples/benchmark/main.go
@@ -109,7 +109,7 @@ func run() error {
 	// Wait for state of escrow accounts to be ingested by Horizon.
 	time.Sleep(2 * time.Second)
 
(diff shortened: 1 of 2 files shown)


acharb (approved):
lgtm 👍 

❓ `Interfaces should be defined by consumers.` 
do you mean the structs that satisfy an interface should be defined by consumers?

acharb:
huh I hadn't thought about that before but makes sense, found this to wrap my head around: ["that way, new methods can be added to implementations without requiring extensive refactoring."](https://github.com/golang/go/wiki/CodeReviewComments#interfaces)

PR #359, opened 2021-10-08
agent/bufferedagent: add stats and rename queue to buffer

leighmcculloch:
### What
Add stats. Rename the use of the terms queue and settlements to buffer.

### Why
To reduce the terms being used in multiple places to refer to the same thing. It was beginning to get confusing. When I was adding the stats it became clear that the existing use of multiple terms was a problem.


examples/console/stats.go
```diff
@@ -0,0 +1,50 @@
...
+
+type stats struct {
+	paymentsSent             int64
+	paymentsReceived         int64
+	bufferedPaymentsSent     int64
+	bufferedPaymentsReceived int64
+}
+
+func (s *stats) Reset() {
+	atomic.StoreInt64(&s.paymentsSent, 0)
```
acharb, line 18: ❓ jw why you chose atmoic here (since I haven't used before). A quick google search says it's because `atomic` is quicker but `mutex` is better for complex data structures, is that why you chose atomic here?

PR #362, opened 2021-10-09
sdk/agent/bufferedagent: add memos for buffered payments

leighmcculloch:
### What
Add memos to buffered payments.

### Why
So that buffered payments are identifiable.

### DIFF
--- a/examples/console/main.go
+++ b/examples/console/main.go
@@ -133,9 +133,9 @@ func run() error {
 				stats.AddPaymentsSent(1)
 
 			case bufferedagent.BufferedPaymentsReceivedEvent:
-				stats.AddBufferedPaymentsReceived(len(e.Amounts))
+				stats.AddBufferedPaymentsReceived(len(e.Payments))
 			case bufferedagent.BufferedPaymentsSentEvent:
-				stats.AddBufferedPaymentsSent(len(e.Amounts))
+				stats.AddBufferedPaymentsSent(len(e.Payments))
 
 			case agentpkg.ClosingEvent:
 				fmt.Fprintf(os.Stderr, "channel closing\n")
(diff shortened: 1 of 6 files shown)


examples/console/main.go
```diff
@@ -370,8 +370,9 @@ func prompt(agent *bufferedagent.Agent, stats *stats, submitter agentpkg.Submitt
 		agent.SetMaxBufferSize(bufferSize)
 		stats.MarkStart()
 		for i := 0; i < x; i++ {
+			memo := "tx-" + strconv.Itoa(i)
```
acharb, line 373: 💭 I think I remember reading somewhere using `+` for string concatenation is slower than using a built in function like `fmt.Sprintf`. Since this line is going to happen a lot it may make some tps difference if using that instead?

leighmcculloch: I doubt there is much difference because fmt.Sprintf still has to accept a string and create a new string. In fact Sprintf has to parse the %d out too and do param checking. Out of curiosity I benchmarked it, and I think it's fine as is.

```
goos: darwin
goarch: amd64
pkg: github.com/stellar/go
cpu: Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
BenchmarkStringConcat-8         19532268                60.56 ns/op           23 B/op          1 allocs/op
BenchmarkStringJoin-8           22743591                58.35 ns/op           23 B/op          1 allocs/op
BenchmarkStringSprintf-8        12789973                98.36 ns/op           23 B/op          2 allocs/op
BenchmarkStringBuilder-8        16457984                77.42 ns/op           31 B/op          2 allocs/op
PASS
ok      github.com/stellar/go   5.420s
```

```go
package main

import (
        "fmt"
        "strconv"
        "strings"
        "testing"
)

var s = ""

func BenchmarkStringConcat(b *testing.B) {
        for i := 0; i < b.N; i++ {
                s = "tx-" + strconv.Itoa(i)
        }
}

func BenchmarkStringJoin(b *testing.B) {
        for i := 0; i < b.N; i++ {
                s = strings.Join([]string{"tx-", strconv.Itoa(i)}, "")
        }
}

func BenchmarkStringSprintf(b *testing.B) {
        for i := 0; i < b.N; i++ {
                s = fmt.Sprintf("tx-%d", i)
        }
}

func BenchmarkStringBuilder(b *testing.B) {
        for i := 0; i < b.N; i++ {
                sb := strings.Builder{}
                sb.WriteString("tx-")
                sb.WriteString(strconv.Itoa(i))
                s = sb.String()
        }
}
```

acharb: good to know! thx for looking into it

acharb (approved):

PR #381, opened 2021-10-21
examples/console: make user payments random amounts

leighmcculloch:
### What
Make user payments random amounts between 1 and the amount entered.

### Why
To make the buffered payments tests more realistic.


examples/console/main.go
```diff
@@ -381,12 +382,14 @@ func prompt(agent *bufferedagent.Agent, stats *stats, submitter agentpkg.Submitt
...
 			return err
 		}
+		rand := rand.New(rand.NewSource(time.Now().UnixNano()))
 		stats.Reset()
 		agent.SetMaxBufferSize(bufferSize)
 		stats.MarkStart()
 		for i := 0; i < x; i++ {
 			memo := "tx-" + strconv.Itoa(i)
 			for {
+				amt := rand.Int63n(amt) + 1
```
acharb, line 392: ❓ will this hit the limit of the amount we deposit? i thought that's why you would use a very small value less than 1 during the demos

leighmcculloch: If you run a test with an amount of 0.01, it will find random values between 0.0000001 and 0.01.

PR #382, opened 2021-10-21
all: update stellar/go dependency to use updated with protocol 18 branch

acharb:
**WHAT**
updates the go-sdk dependency to the[ updated branch](https://github.com/stellar/go/tree/cap21and40) we have on the stellar/go repo now

**WHY**
updating the go-sdk with the protocol 18 changes, and updating our sdk since it depends on it

### DIFF
--- a/examples/benchmark/go.mod
+++ b/examples/benchmark/go.mod
@@ -2,13 +2,11 @@ module github.com/stellar/experimental-paymentment-channels/examples/benchmark
 
 go 1.17
 
-replace github.com/stellar/go => github.com/stellar/go v0.0.0-20210924221542-a72304844558
-
 replace github.com/stellar/experimental-payment-channels/sdk => ../../sdk
 
 require (
 	github.com/stellar/experimental-payment-channels/sdk v0.0.0-00010101000000-000000000000
-	github.com/stellar/go v0.0.0-00010101000000-000000000000
+	github.com/stellar/go v0.0.0-20211019190220-c47964999c2a
 	gopkg.in/yaml.v2 v2.3.0 // indirect
 )
 
--- a/sdk/state/asset.go
+++ b/sdk/state/asset.go
@@ -42,7 +42,7 @@ func (a Asset) StringCanonical() string {
 	return fmt.Sprintf("%s:%s", a.Code(), a.Issuer())
 }
 
-func (a Asset) EqualsTrustLineAsset(ta xdr.TrustLineAsset) bool {
+func (a Asset) EqualTrustLineAsset(ta xdr.TrustLineAsset) bool {
 	if ta.Type == xdr.AssetTypeAssetTypePoolShare {
 		return false
 	}
--- a/sdk/state/asset.go
+++ b/sdk/state/asset.go
@@ -41,3 +41,10 @@ func (a Asset) StringCanonical() string {
 	}
 	return fmt.Sprintf("%s:%s", a.Code(), a.Issuer())
 }
+
+func (a Asset) EqualsTrustLineAsset(ta xdr.TrustLineAsset) bool {
+	if ta.Type == xdr.AssetTypeAssetTypePoolShare {
+		return false
+	}
+	return ta.ToAsset().StringCanonical() == a.StringCanonical()
+}
--- a/sdk/state/asset.go
+++ b/sdk/state/asset.go
@@ -43,8 +43,9 @@ func (a Asset) StringCanonical() string {
 }
 
 func (a Asset) EqualTrustLineAsset(ta xdr.TrustLineAsset) bool {
-	if ta.Type == xdr.AssetTypeAssetTypePoolShare {
-		return false
+	switch ta.Type {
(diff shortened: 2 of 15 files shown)


sdk/state/ingest.go
```diff
@@ -277,7 +277,7 @@ func (c *Channel) ingestFormationTx(tx *txnbuild.Transaction, resultXDR string,
 
 			switch updated.Data.Type {
 			case xdr.LedgerEntryTypeTrustline:
-				if updated.Data.TrustLine.Asset.StringCanonical() != c.openAgreement.Envelope.Details.Asset.StringCanonical() {
+				if updated.Data.TrustLine.Asset.ToAsset().StringCanonical() != c.openAgreement.Envelope.Details.Asset.StringCanonical() {
```
leighmcculloch, line 280: 💥 `ToAsset()` is unfortunately not safe to call on all asset types at the moment. If the trustline is a liquidity pool, it will panic.

See https://github.com/stellar/go/blob/5742327250c1468a68700a41c48ea16cb74dd8e4/xdr/asset.go#L257-L269.

I think we should add an `EqualsTrustLineAsset(ta xdr.TrustLineAsset)` method to the `state.Asset` type that will do a safer comparison. Wdyt?

sdk/txbuild/create_escrow.go
```diff
@@ -35,8 +36,12 @@ func CreateEscrow(p CreateEscrowParams) (*txnbuild.Transaction, error) {
 		},
 	}
 	if !p.Asset.IsNative() {
+		cta, err := p.Asset.ToChangeTrustAsset()
+		if err != nil {
+			return nil, fmt.Errorf("getting change trust asset from nonnative asset: %w", err)
```
leighmcculloch, line 41: 🔡 
```suggestion
			return nil, fmt.Errorf("getting change trust asset from non-native asset: %w", err)
```

sdk/txbuild/create_escrow.go
```diff
@@ -35,8 +36,12 @@ func CreateEscrow(p CreateEscrowParams) (*txnbuild.Transaction, error) {
 		},
 	}
 	if !p.Asset.IsNative() {
+		cta, err := p.Asset.ToChangeTrustAsset()
```
leighmcculloch, line 39: 💡 I'd call `MustToChangeTrustAsset` instead. We only support native assets and credit assets, and both of these assets will never return an error from that other function.

```suggestion
		cta, err := p.Asset.MustChangeTrustAsset()
```

examples/benchmark/go.mod
```diff
@@ -2,13 +2,11 @@ module github.com/stellar/experimental-paymentment-channels/examples/benchmark
...
 go 1.17
 
-replace github.com/stellar/go => github.com/stellar/go v0.0.0-20210924221542-a72304844558
-
 replace github.com/stellar/experimental-payment-channels/sdk => ../../sdk
 
 require (
 	github.com/stellar/experimental-payment-channels/sdk v0.0.0-00010101000000-000000000000
-	github.com/stellar/go v0.0.0-00010101000000-000000000000
+	github.com/stellar/go v0.0.0-20211019190220-c47964999c2a
```
leighmcculloch, line 9: 💯 Good call on removing the replaces 🎉.

leighmcculloch (commented):
I think we should make one change (💥), otherwise looks great.

sdk/state/asset.go
```diff
@@ -41,3 +41,10 @@ func (a Asset) StringCanonical() string {
 	}
 	return fmt.Sprintf("%s:%s", a.Code(), a.Issuer())
 }
+
+func (a Asset) EqualsTrustLineAsset(ta xdr.TrustLineAsset) bool {
```
leighmcculloch, line 45: 💡 To follow convention of not making equal plural. This is my fault sorry, I suggested equals before.
```suggestion
func (a Asset) EqualTrustLineAsset(ta xdr.TrustLineAsset) bool {
```

sdk/state/asset.go
```diff
@@ -41,3 +41,10 @@ func (a Asset) StringCanonical() string {
...
+	if ta.Type == xdr.AssetTypeAssetTypePoolShare {
+		return false
+	}
+	return ta.ToAsset().StringCanonical() == a.StringCanonical()
```
leighmcculloch, line 49: 💡 I'd flip this so as to compare the asset types we know about, rather than attempt to exclude all possible asset types we don't know about, since that set of asset types could get larger over time and it keeps this function scoped to the things it knows about.
```suggestion
	switch ta.Type {
	case xdr.AssetTypeNative, xdr.AssetTypeAssetTypeCreditAlphanum4, xdr. AssetTypeAssetTypeCreditAlphanum12:
		return ta.ToAsset().StringCanonical() == a.StringCanonical()
	default:
		return false
	}
```

acharb: i was leaning towards simple, but agreed this does seem safer long term
https://github.com/stellar/experimental-payment-channels/pull/382/commits/323404c7a23c31714c22823246c502972d351df9

PR #400, opened 2021-11-01
sdk/msg, sdk/agent/bufferedagent: use gob for encoding data and make memos bytes

leighmcculloch:
### What
Use gob for encoding instead of json, and make payment memos are byte slice instead of a string.

### Why
Gob encoding is faster and more lightweight, and since memos can store lots of different types of data they are easier to work with if they are byte slices, which can contain strings anyway.

### DIFF
--- a/sdk/msg/msg.go
+++ b/sdk/msg/msg.go
@@ -1,7 +1,7 @@
 package msg
 
 import (
-	"encoding/json"
+	"encoding/gob"
 	"io"
 
 	"github.com/stellar/experimental-payment-channels/sdk/state"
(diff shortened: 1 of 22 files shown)


sdk/msg/msg.go
```diff
@@ -11,43 +11,43 @@ import (
...
 
 const (
-	TypeHello           Type = 100
-	TypeOpenRequest     Type = 200
-	TypeOpenResponse    Type = 201
-	TypePaymentRequest  Type = 300
-	TypePaymentResponse Type = 301
-	TypeCloseRequest    Type = 400
-	TypeCloseResponse   Type = 401
+	TypeHello           Type = 10
```
acharb, line 14: ❓ jw was this part of the gob addition? or something separate (im assuming bc the codes were too similar to http ones?)

acharb (approved):
lgtm!

sdk/msg/msg.go
leighmcculloch: Gob stores integers in a variable int format. I think it is more complex than this, but it is something like 7 bits for the data bits and 1 bit to signal if there's more bytes. These integers are just identifiers for the different message types and they do not need to be that big. Changing them so they all fit inside one byte.

PR #403, opened 2021-11-03
examples/console: use ishell for console example

acharb:
**WHAT**
add to the example a more feature complete interactive shell. Using: https://github.com/abiosoft/ishell

**WHY**
the added features ishell gives makes it easier for a developer to interact with the console app. Will also improve the demo since the recording shows us interacting with the example console app.

note: there shouldn't be any code changes for how the agent is used, the only code changing is starting the shell and adding the commands. (Removed the help and exit commands. ishell has those built in)

### DIFF
--- a/examples/console/main.go
+++ b/examples/console/main.go
@@ -358,7 +358,7 @@ func runShell(agent *bufferedagent.Agent, stats *stats, submitter agentpkg.Submi
 	})
 	shell.AddCmd(&ishell.Cmd{
 		Name: "open",
-		Help: "open - open a channel with optional asset",
+		Help: "open [asset-code] - open a channel with optional asset",
 		Func: func(c *ishell.Context) {
 
 			assetCode := ""
--- a/examples/console/main.go
+++ b/examples/console/main.go
@@ -355,7 +355,6 @@ func runShell(agent *bufferedagent.Agent, stats *stats, submitter agentpkg.Submi
 		Help: "connect <addr>:<port> - connect to a peer",
 		Func: func(c *ishell.Context) {
 			c.Err(agent.ConnectTCP(c.Args[0]))
-			return
 		},
 	})
 	shell.AddCmd(&ishell.Cmd{
@@ -371,7 +370,6 @@ func runShell(agent *bufferedagent.Agent, stats *stats, submitter agentpkg.Submi
 				asset = state.Asset(assetCode + ":" + signer.Address())
 			}
 			c.Err(agent.Open(asset))
-			return
 		},
 	})
 	shell.AddCmd(&ishell.Cmd{
@@ -425,7 +423,6 @@ func runShell(agent *bufferedagent.Agent, stats *stats, submitter agentpkg.Submi
 			stats.Reset()
 			_, err = agent.Payment(amt)
 			c.Err(err)
-			return
 		},
 	})
 	shell.AddCmd(&ishell.Cmd{
@@ -474,7 +471,6 @@ func runShell(agent *bufferedagent.Agent, stats *stats, submitter agentpkg.Submi
 			fmt.Println(stats.Summary())
 			agent.SetMaxBufferSize(1)
 			c.Err(err)
-			return
 		},
 	})
 	shell.AddCmd(&ishell.Cmd{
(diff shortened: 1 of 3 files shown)


examples/console/main.go
```diff
@@ -325,197 +323,219 @@ func run() error {
...
+	shell.AddCmd(&ishell.Cmd{
+		Name: "connect",
+		Help: "connect <addr>:<port> - connect to a peer",
+		Func: func(c *ishell.Context) {
+			c.Err(agent.ConnectTCP(c.Args[0]))
+		},
+	})
+	shell.AddCmd(&ishell.Cmd{
+		Name: "open",
+		Help: "open - open a channel with optional asset",
```
leighmcculloch, line 361: :bulb:
```suggestion
		Help: "open [asset-code] - open a channel with optional asset",
```

examples/console/main.go
```diff
@@ -325,197 +323,219 @@ func run() error {
...
+		Func: func(c *ishell.Context) {
+
```
leighmcculloch, line 363: 💡 
```suggestion
```

examples/console/main.go
```diff
@@ -325,197 +323,219 @@ func run() error {
...
+				c.Err(err)
+			}
+			stats.Reset()
+			_, err = agent.Payment(amt)
+			c.Err(err)
+		},
+	})
+	shell.AddCmd(&ishell.Cmd{
+		Name: "payx",
+		Help: "payx <amount> <number of times> - pay an amount a certain number of times",
```
leighmcculloch, line 425: 💡 
```suggestion
		Help: "payx <max-amount> <number of times> <max buffer size> - pay an amount a certain number of times",
```

examples/console/main.go
```diff
@@ -325,197 +323,219 @@ func run() error {
...
+		Func: func(c *ishell.Context) {
+			fmt.Fprintf(os.Stdout, "sending %s payment %s times\n", c.Args[0], c.Args[1])
+			amt, err := amount.ParseInt64(c.Args[0])
+			if err != nil {
+				c.Err(err)
```
leighmcculloch, line 430: Do we need to return at all these points too?

leighmcculloch (approved):

examples/console/main.go
acharb: yup good catch. c.Err informs shell there's an error but doesn't end the func. Added here: https://github.com/stellar/starlight/pull/403/commits/00b822be0358d759ee31f8450aa7a4504661c71c

examples/console/main.go
```diff
@@ -325,197 +323,238 @@ func run() error {
...
-			var receiver string
-			if signer.FromAddress().Equal(a.Envelope.Details.ConfirmingSigner) {
-				receiver = "me"
-			} else {
-				receiver = "them"
+			stats.MarkFinish()
+			fmt.Println(stats.Summary())
+			agent.SetMaxBufferSize(1)
+			c.Err(err)
+			return
```
leighmcculloch, line 477: This return is redundant?

examples/console/main.go
```diff
@@ -325,197 +323,238 @@ func run() error {
...
+			err = submitter.SubmitTx(tx)
+			if err != nil {
+				c.Err(err)
+				return
+			}
+		},
+	})
+
+	shell.Run()
+	return
```
leighmcculloch, line 559:
```suggestion
```

examples/console/main.go
acharb: yup lint caught it. Was trying to make it consistent, but removed
https://github.com/stellar/starlight/pull/403/commits/5b8e9b1c42912817a5e6a2f91d1efb8432e4e22d

examples/console/main.go
acharb: https://github.com/stellar/starlight/pull/403/commits/91ef9b5b287d67ecae82e8a540ac97f9bc7cd908

acharb: forgot this one's actually needed
https://github.com/stellar/starlight/pull/403/commits/9c6ef4db4bc8bc357831b9f34e7182e8566ea48a

PR #405, opened 2021-11-03
sdk/state: remove signers from close agreements

leighmcculloch:
### What
Remove signers from close agreements.

### Why
They don't actually need to be in the close agreement envelope that we transmit. Both sides already know the signers and all we really need to communicate is who is proposing this transaction, the initiator or the responder.

### DIFF
--- a/Getting Started.md
+++ b/Getting Started.md
@@ -1,7 +1,13 @@
 # Getting Started
 
-## Running Core and Horizon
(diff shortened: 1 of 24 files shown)


leighmcculloch:
I'm going to close this PR. I believe this is going in a good direction and it would be ideal to not transmit the signers on every message exchange. However, it has little impact on performance since signers are quite small anyway.

PR #406, opened 2021-11-03
sdk/state, sdk/agent: finalize payment

leighmcculloch:
### What
Add a finalize payment function that a payment proposer can use to finalize a payment when the confirmer sends back their signatures.

### Why
Today this functionality is provided by the `ConfirmPayment` function that is used by both the proposer and the confirmer. Because it is used by both it accepts a full close agreement envelope and does a lot of validation of the contents. The proposer gets back the signatures from the confirmer and doesn't actually need to verify the agreement. They just need to attach the signatures and see if they're valid. That's what the finalize payment function does.

This change has no material performance improvement. I attempted the change to see if it improved performance and it did not in a meaningful way. Reflecting on it though, this does feel like it naturally makes sense and it simplifies some logic we have in the agent. When the agent is confirming the payment for a proposer it no longer needs to get the close agreement and attach the signatures.

### DIFF
--- a/sdk/state/payment.go
+++ b/sdk/state/payment.go
@@ -352,3 +352,34 @@ func (c *Channel) ConfirmPayment(ce CloseEnvelope) (closeAgreement CloseAgreemen
 
 	return c.latestAuthorizedCloseAgreement, nil
 }
+
+// FinalizePayment finalizes a payment, making it authorized, by attaching the
+// close signatures to the agreement as the confirmers signatures. The proposer
+// of a payment calls this once with the confirmers signatures when the
+// confirmer provides them. This can only be used to finalize the most recent
+// unauthorized payment.
+func (c *Channel) FinalizePayment(cs CloseSignatures) (closeAgreement CloseAgreement, err error) {
+	if c.latestUnauthorizedCloseAgreement.Envelope.Empty() {
+		return CloseAgreement{}, fmt.Errorf("no unauthorized close agreement to finalize")
+	}
+
+	txs := c.latestUnauthorizedCloseAgreement.Transactions
+
+	// If remote has not signed the txs or signatures is invalid, error as is invalid.
+	verifyInputs := []signatureVerificationInput{
+		{TransactionHash: txs.DeclarationHash, Signature: cs.Declaration, Signer: c.remoteSigner},
+		{TransactionHash: txs.CloseHash, Signature: cs.Close, Signer: c.remoteSigner},
+	}
+	err = verifySignatures(verifyInputs)
+	if err != nil {
+		return CloseAgreement{}, fmt.Errorf("invalid signature: %w", err)
+	}
+
+	// All signatures are present that would be required to submit all
+	// transactions in the payment.
+	c.latestUnauthorizedCloseAgreement.Envelope.ConfirmerSignatures = cs
+	c.latestAuthorizedCloseAgreement = c.latestUnauthorizedCloseAgreement
+	c.latestUnauthorizedCloseAgreement = CloseAgreement{}
+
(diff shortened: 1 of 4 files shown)


sdk/state/payment.go
```diff
@@ -352,3 +352,34 @@ func (c *Channel) ConfirmPayment(ce CloseEnvelope) (closeAgreement CloseAgreemen
...
+	txs := c.latestUnauthorizedCloseAgreement.Transactions
+
+	// If remote has not signed the txs or signatures is invalid, error as is invalid.
+	verifyInputs := []signatureVerificationInput{
+		{TransactionHash: txs.DeclarationHash, Signature: cs.Declaration, Signer: c.remoteSigner},
+		{TransactionHash: txs.CloseHash, Signature: cs.Close, Signer: c.remoteSigner},
+	}
+	err = verifySignatures(verifyInputs)
+	if err != nil {
+		return CloseAgreement{}, fmt.Errorf("invalid signature: %w", err)
```
acharb, line 375: 💡 
```suggestion
		return CloseAgreement{}, fmt.Errorf("invalid remote signature: %w", err)
```

sdk/state/payment.go
```diff
@@ -352,3 +352,34 @@ func (c *Channel) ConfirmPayment(ce CloseEnvelope) (closeAgreement CloseAgreemen
 
 	return c.latestAuthorizedCloseAgreement, nil
 }
+
+// FinalizePayment finalizes a payment, making it authorized, by attaching the
+// close signatures to the agreement as the confirmers signatures. The proposer
+// of a payment calls this once with the confirmers signatures when the
+// confirmer provides them. This can only be used to finalize the most recent
+// unauthorized payment.
+func (c *Channel) FinalizePayment(cs CloseSignatures) (closeAgreement CloseAgreement, err error) {
```
acharb, line 361: 💭 this can be only be called by the proposer right? if so, can we rename to `FinalizeProposedPayment`? imo I think makes it clearer who calls + when the method should be called

acharb (commented):
👍 had 2 comment/thoughts

sdk/state/payment.go
leighmcculloch: I'd rather keep this error uniform with the invalid signature errors returned from the `ConfirmPayment` function. For either remote or local errors this is the form we return. Lmk if you disgree.

acharb: no that makes sense 👍  nvrm

sdk/state/payment.go
leighmcculloch: Good point. It's not clear today that both participants call Confirm. The good thing about the API is that a mistake calling the wrong function has no unsafe affect. Any incorrect call will cause an error rather than break the channel, but we should do better to lead the developer for sure.

I think introducing the word Proposed could still be misleading because a receiver could think of the close agreement they've received as a proposed agreement, and could think they need to finalize the proposed agreement. So I don't think the name change addresses the overarching problem with understanding what functions to call when. 

I've pushed some new docs in 68c671dee1f6c916532f8f30945a658b002346ff. I think this can be the start of more comprehensive docs for the code. What do you think?

acharb: > I think introducing the word Proposed could still be misleading because a receiver could think of the close agreement they've received as a proposed agreement

good point. The case I was thinking of is the receiver receives a new agreement and thinks they can "finalize" it by calling `FinalizePayment` instead of `ConfirmAgreement`. The diagram you added makes it pretty clear though that they should call Confirm, so lgtm 👍  if they do call Finalize by mistake it fails anyways

acharb (approved):

PR #408, opened 2021-11-04
all: replace escrow term with channel account

acharb:
**WHAT**
replaces the term escrow with ~multi sig~ channel account

**WHY**
escrow implies the use of a third party, which is not what the accounts in the payment channels are. ~Using `MultiSig` instead makes it clearer what the account's purpose in the payment channel process is.~
`channel account` is free of this implicit meaning, and makes it clear what the use of these accounts are. Ie the accounts that make up the channel.

as part of https://github.com/stellar/starlight/issues/371

this PR is making the code changes. The SEP changes are done in: https://github.com/stellar/starlight/pull/410


specifications/starlight-payment-channel-mechanics.md
```diff
@@ -171,7 +172,7 @@ To setup the payment channel:
    - A closing transaction C_i, that closes the channel without any payment.
 6. R signs and submits F.
 
-Participants should check the state of the other participant's escrow account
+Participants should check the state of the other participant's multi-sig account
```
leighmcculloch, line 175:
```suggestion
Participants should check the state of the other participant's account
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -180,8 +181,8 @@ Signatures for F, D_i, and C_i may be shared in a single message.
 
 The transactions are constructed as follows:
 
-- F, the _open transaction_, changes escrow accounts EI and ER to be 2-of-2
-multisig accounts. F has source account E, and sequence number set to s.
+- F, the _open transaction_, changes multi-sig accounts MI and MR to be 2-of-2
+multi-sig accounts. F has source account E, and sequence number set to s.
```
leighmcculloch, line 185: The accounts aren't multisig initially, but become multisig.
```suggestion
- F, the _open transaction_, changes accounts MI and MR to be 2-of-2
multi-sig accounts. F has source account E, and sequence number set to s.
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -197,47 +198,47 @@ multisig accounts. F has source account E, and sequence number set to s.
...
+  - Operations sponsored by I for MI:
     - One `BEGIN_SPONSORING_FUTURE_RESERVES` operation that specifies
     participant I as a sponsor of future reserves.
-    - One `CHANGE_TRUST` operation configuring trustlines on EI if the asset is 
+    - One `CHANGE_TRUST` operation configuring trustlines on MI if the asset is 
     not the native asset.
-    - One `SET_OPTIONS` operation adjusting escrow account EI's thresholds such
-    that I and R's signers must both sign.
-    - One `SET_OPTIONS` operation adding I signers to EI.
+    - One `SET_OPTIONS` operation adjusting multi-sig account MI's thresholds
```
leighmcculloch, line 206:
```suggestion
    - One `SET_OPTIONS` operation adjusting account MI's thresholds 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -197,47 +198,47 @@ multisig accounts. F has source account E, and sequence number set to s.
...
+  - Operations sponsored by R for MR:
     - One `BEGIN_SPONSORING_FUTURE_RESERVES` operation that specifies reserve
     account R as a sponsor of future reserves.
-    - One `CHANGE_TRUST` operation configuring trustlines on ER if the asset is 
+    - One `CHANGE_TRUST` operation configuring trustlines on MR if the asset is 
     not the native asset.
-    - One `SET_OPTIONS` operation adjusting escrow account ER's thresholds such
-    that R and I's signers must both sign.
-    - One `SET_OPTIONS` operation adding R's signers to ER.
+    - One `SET_OPTIONS` operation adjusting multi-sig account MR's thresholds
```
leighmcculloch, line 222:
```suggestion
    - One `SET_OPTIONS` operation adjusting account MR's thresholds 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -197,47 +198,47 @@ multisig accounts. F has source account E, and sequence number set to s.
...
+  - Operations sponsored by R for MI:
     - One `BEGIN_SPONSORING_FUTURE_RESERVES` operation that specifies reserve
     account R as a sponsor of future reserves.
-    - One `SET_OPTIONS` operations adding R's signers to EI.
+    - One `SET_OPTIONS` operations adding R's signers to MI.
     - One `END_SPONSORING_FUTURE_RESERVES` operation that stops R sponsoring
     future reserves of subsequent operations.
 
-  The escrow accounts EI and ER will likely have the necessary trustline
+  The multi-sig accounts MI and MR will likely have the necessary trustline
```
leighmcculloch, line 234:
```suggestion
  Accounts MI and MR will likely have the necessary trustline
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -197,47 +198,47 @@ multisig accounts. F has source account E, and sequence number set to s.
...
   before the open transaction is built. This means the `CHANGE_TRUST`
   operation will likely be a no-op. The `CHANGE_TRUST` operation must be
   included in the open transaction so that participants are guaranteed the
   trustline is still in the same state after the channel is open. If the 
   operation is not included a participant could intentionally or accidentally 
-  remove a trustline between escrow account setup and open causing the presigned
-  closing transaction to become invalid.
+  remove a trustline between multi-sig account setup and open causing the
```
leighmcculloch, line 240:
```suggestion
  remove a trustline between the account setup and open causing the 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -285,19 +286,19 @@ ledger count).
...
-  - Zero or one `PAYMENT` operations that disburses funds from EI to ER, or from
-  ER to EI, that may be omitted if the final state at this update does not
+  - Zero or one `PAYMENT` operations that disburses funds from MI to MR, or from
+  MR to MI, that may be omitted if the final state at this update does not
   require the movement of funds.
-  - One or more `SET_OPTIONS` operation adjusting escrow account EI's thresholds
-  to give I full control of EI, and removing R's signers.
-  - One or more `SET_OPTIONS` operation adjusting reserve account ER's
-  thresholds to give R full control of ER, and removing I's signers.
+  - One or more `SET_OPTIONS` operation adjusting multi-sig account MI's
```
leighmcculloch, line 292:
```suggestion
  - One or more `SET_OPTIONS` operation adjusting account MI's 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -353,8 +354,8 @@ Close](#Uncoordinated-Close) process with a declaration transaction that is not
 the most recently signed declaration transaction.
 
 The other participant can identify that the close process has started at an
-earlier state by monitoring changes in escrow account EI's sequence. If the
-other participant sees the sequence number of escrow account EI change to a
+earlier state by monitoring changes in multi-sig account MI's sequence. If the
+other participant sees the sequence number of multi-sig account MI change to a
```
leighmcculloch, line 358:
```suggestion
earlier state by monitoring changes in account MI's sequence. If the
other participant sees the sequence number of account MI change to a
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -381,16 +382,16 @@ Some operations are implemented in a two-step process. Participants agree on a
 new closing state at a future iteration by signing C_i and D_i transactions
 where i has skipped an iteration that is not yet executable because the D_i's
 `minSeqNum` is also set in the future. Participants then sign a transaction to
-make the change that only moves the sequence of escrow account EI to satisfy the
-`minSeqNum` of the future D_i.
+make the change that only moves the sequence of multi-sig account MI to satisfy
```
leighmcculloch, line 385:
```suggestion
make the change that only moves the sequence of account MI to satisfy 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -381,16 +382,16 @@ Some operations are implemented in a two-step process. Participants agree on a
...
-`minSeqNum` of the future D_i.
+make the change that only moves the sequence of multi-sig account MI to satisfy 
+the `minSeqNum` of the future D_i.
 
 Operations that can fail and change the balances of the channel have the
 following requirements as well:
 
 - The transaction that can fail must have its source account set to an account
-that is not escrow account EI.
+that is not multi-sig account MI.
```
leighmcculloch, line 392:
```suggestion
that is not account MI.
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -381,16 +382,16 @@ Some operations are implemented in a two-step process. Participants agree on a
...
 - The transaction that can fail must contain a `BUMP_SEQUENCE` operation that
-bumps escrow account EI's sequence number to a sequence number that makes the
+bumps multi-sig account MI's sequence number to a sequence number that makes the
```
leighmcculloch, line 394:
```suggestion
bumps account MI's sequence number to a sequence number that makes the
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -404,16 +405,16 @@ Operations that can fail and where the additional requirements apply:
 ##### Deposit / Top-up
 
 Participants may deposit into the channel without coordination, as long as both
-escrow accounts EI and ER already have a trustline for the asset being
+multi-sig accounts MI and MR already have a trustline for the asset being
```
leighmcculloch, line 408:
```suggestion
accounts MI and MR already have a trustline for the asset being
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -404,16 +405,16 @@ Operations that can fail and where the additional requirements apply:
...
 Participant I deposits or tops-up their balance by using a standard payment
-operation to EI.
+operation to MI.
 
 Participant R deposits or tops-up their balance by using a standard payment
-operation to ER.
+operation to MR.
 
-If participants wish to deposit an asset that escrow accounts EI or ER do not
+If participants wish to deposit an asset that multi-sig accounts MI or MR do not
```
leighmcculloch, line 417:
```suggestion
If participants wish to deposit an asset that accounts MI or MR do not
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -437,17 +438,17 @@ withdrawing and X is the participant witnessing the withdrawal:
 7. I or R submit W_i.
 
 If the withdrawal transaction W_i fails or is never submitted, the C_i and D_i
-are not executable because escrow account EI's sequence number was not bumped to
-s_i.  The participants should take the following steps since the withdrawal did
-not succeed:
+are not executable because multi-sig account MI's sequence number was not bumped
```
leighmcculloch, line 441:
```suggestion
are not executable because account MI's sequence number was not bumped 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -437,17 +438,17 @@ withdrawing and X is the participant witnessing the withdrawal:
...
 
 8. Set e to the value of e'.
 
 The transactions are constructed as follows:
 
-- W_i, the _withdrawal transaction_, makes one or more payments from the escrow
-account EI and/or ER to any Stellar account. W_i has any source account that is
-not EI, typically the participant proposing the change.
+- W_i, the _withdrawal transaction_, makes one or more payments from the 
+multi-sig account MI and/or MR to any Stellar account. W_i has any source
```
leighmcculloch, line 450:
```suggestion
account MI and/or MR to any Stellar account. W_i has any source 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -463,10 +464,10 @@ not EI, typically the participant proposing the change.
 
   W_i contains operations:
 
-  - One `PAYMENT` operations withdrawing assets from escrow accounts EI and/or
-  ER.
-  - One `BUMP_SEQUENCE` operation bumping the sequence number of escrow account
-  EI to s_i.
+  - One `PAYMENT` operations withdrawing assets from multi-sig accounts MI
```
leighmcculloch, line 467:
```suggestion
  - One `PAYMENT` operations withdrawing assets from accounts MI 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -463,10 +464,10 @@ not EI, typically the participant proposing the change.
 
   W_i contains operations:
 
-  - One `PAYMENT` operations withdrawing assets from escrow accounts EI and/or
-  ER.
-  - One `BUMP_SEQUENCE` operation bumping the sequence number of escrow account
-  EI to s_i.
+  - One `PAYMENT` operations withdrawing assets from multi-sig accounts MI 
+  and/or MR.
+  - One `BUMP_SEQUENCE` operation bumping the sequence number of multi-sig
```
leighmcculloch, line 469:
```suggestion
  - One `BUMP_SEQUENCE` operation bumping the sequence number of
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -482,7 +483,7 @@ to the network.
 The participants may agree at anytime to increase period O by using a larger
 value for O in the next and future transaction sets, or regenerating the most
 recent transaction set, then signing and submitting a transaction that bumps the
-sequence number of the escrow account to the sequence before the most recent
+sequence number of the multi-sig account to the sequence before the most recent
```
leighmcculloch, line 486:
```suggestion
sequence number of the account to the sequence before the most recent
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -502,9 +503,9 @@ participants:
 
 The transactions are constructed as follows:
 
-- B_i, the _bump transaction_, bumps the sequence number of escrow account E
+- B_i, the _bump transaction_, bumps the sequence number of multi-sig account E
```
leighmcculloch, line 506:
```suggestion
- B_i, the _bump transaction_, bumps the sequence number of account E
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -524,7 +525,7 @@ EI, sequence number s_i.
 
 #### Reusing a Channel
 
-After close, escrow accounts EI and ER can be reused for another channel with
+After close, multi-sig accounts MI and MR can be reused for another channel with
```
leighmcculloch, line 528: At this point, after close, they are no longer multisig.
```suggestion
After close, accounts MI and MR can be reused for another channel with
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -535,15 +536,15 @@ being set to zero.
 All transaction fees are paid by the participant submitting the transaction to
 the Stellar network.
 
-All transactions defined in the protocol with escrow account EI as the source
+All transactions defined in the protocol with multi-sig account MI as the source
```
leighmcculloch, line 539:
```suggestion
All transactions defined in the protocol with account MI as the source
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -535,15 +536,15 @@ being set to zero.
...
 account have their fees set to zero.  The submitter of a transaction wraps the
 transaction in a fee bump transaction envelope and provides an appropriate fee,
 paying the fee themselves.
 
-Credits and debits to escrow accounts EI and ER only ever represent deposits or
-withdrawals by I or R, and the sum of all disbursements at close equal the sum
-of all deposits minus the sum of all withdrawals.  Network transaction fees do
-not change the balance of the channel.
+Credits and debits to multi-sig accounts MI and MR only ever represent deposits
```
leighmcculloch, line 544:
```suggestion
Credits and debits to accounts MI and MR only ever represent deposits 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -552,36 +553,36 @@ supplied by the participant who will be in control of the ledger entry at
 channel close.  Participants should have no impact or dependence on each other
 after channel close, and so they must not sponsor ledger entries that only the
 other party controls after channel close, either directly or indirectly through
-the escrow or reserve accounts.
+the multi-sig or reserve accounts.
```
leighmcculloch, line 556:
```suggestion
the multisig accounts they temporarily share control.
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -552,36 +553,36 @@ supplied by the participant who will be in control of the ledger entry at
...
 Ledger entries that do not survive channel close, such as signers, are sponsored
 by their beneficiary.  Participants pay for their own key and signing
 requirements.
 
 Participant I provides reserves for:
-- Escrow account EI
-- Trustlines added to EI
-- Signers added to EI for I
-- Signers added to ER for I
+- Multi-sig account MI
```
leighmcculloch, line 563: Participant I provides reserves for MI when it is and isn't multisig.
```suggestion
- Account MI
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -552,36 +553,36 @@ supplied by the participant who will be in control of the ledger entry at
...
+- Trustlines added to MI
+- Signers added to MI for I
+- Signers added to MR for I
 
 Participant R provides reserves for:
-- Escrow account ER
-- Trustlines added to ER
-- Signers added to ER for R
-- Signers added to EI for R
+- Multi-sig account MR
```
leighmcculloch, line 569:
```suggestion
- Account MR
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -552,36 +553,36 @@ supplied by the participant who will be in control of the ledger entry at
...
+- Signers added to MR for R
+- Signers added to MI for R
 
 The total reserves required for each participant are:
 
 - Participant I
 
-  - 1 (Escrow Account EI)
-  - \+ Number of Assets (for Trustlines on EI, always 0 or 1)
+  - 1 (Multi-sig Account MI)
```
leighmcculloch, line 578:
```suggestion
  - 1 (Account MI)
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -552,36 +553,36 @@ supplied by the participant who will be in control of the ledger entry at
...
+  - \+ Number of Assets (for Trustlines on MI, always 0 or 1)
   - \+ 2 x Number of I's Signers
 
 - Participant R
 
-  - 1 (Escrow Account ER)
-  - \+ Number of Assets (for Trustlines on ER, always 0 or 1)
+  - 1 (Multi-sig Account MR)
```
leighmcculloch, line 584:
```suggestion
  - 1 (Account MR)
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -648,37 +649,37 @@ may also produce closing transactions that could fail if trustline limits would
...
+compromise the payment channels ability to close successfully.
 
 If the issuer of a clawback enabled trustline submits a clawback operation for
-amounts in either escrow account, the close transaction may fail to process if
-its payment operation is dependent on amounts clawed back.
+amounts in either multi-sig account, the close transaction may fail to process 
+if its payment operation is dependent on amounts clawed back.
 
-### Escrow Account and Trustline State
+### Multi-sig Account and Trustline State
```
leighmcculloch, line 665:
```suggestion
### Account and Trustline State
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -648,37 +649,37 @@ may also produce closing transactions that could fail if trustline limits would
...
 
-Participants can inspect the state of escrow accounts and trustlines before
+Participants can inspect the state of multi-sig accounts and trustlines before
```
leighmcculloch, line 667:
```suggestion
Participants can inspect the state of accounts and trustlines before
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -648,37 +649,37 @@ may also produce closing transactions that could fail if trustline limits would
...
 execution of the open to check the state of the accounts and trustlines are
 acceptable, but there is no guarantee that state will remain constant until
 after the open transaction is executed.
 
-It is critical to check the state of the other participants escrow account and
-their trustlines after opening the channel because there is no way for 
+It is critical to check the state of the other participants multi-sig account
```
leighmcculloch, line 672:
```suggestion
It is critical to check the state of the other participants account 
```

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -648,37 +649,37 @@ may also produce closing transactions that could fail if trustline limits would
...
+It is critical to check the state of the other participants multi-sig account 
+and their trustlines after opening the channel because there is no way for 
 participants to guarantee that the other participant has not altered its state. 
 For example, the other participant could add an additional signer to their 
 account. Or, for example, the other participant could intentionally or 
 accidentally cause flags on their trustline to be changed, such as the clawback 
 enabled flag.
 
-### Escrow Account and Trustline Balance
+### Multi-sig Account and Trustline Balance
```
leighmcculloch, line 680:
```suggestion
### Account and Trustline Balance
```

leighmcculloch (commented):
Thanks for tackling this large concept change that needs to happen.

Could you separate the code in this PR into a separate PR so we can review and get the SEP merged faster first?

After reading the SEP through I think we should avoid using 'multisig' as a noun, which it feels like we're doing here. There are places it doesn't make sense to call the accounts multisig accounts because they aren't multisig accounts at that moment. At the beginning they are not multisig, but are made multisig as part of the open process, and they lose that property during the close process.

Ultimately the only accounts actually named in the SEP are MI and MR (or EI and ER) and so the multisig quality is only necessary, or useful, in the moments where it adds depth to more accurately reflect the nature of the account in that moment or in the context of the safety we're trying to create in the SEP.

I've gone through and commented on the places I think we should remove multisig.

FYI also, I think we should go with the simpler single-word multisig. It is common in literature, more common I believe than multi-sig. It has the advantage that when we take this to code we can write `Multisig` and `multisig` instead of `MultiSig` and `multi_sig`.

specifications/starlight-payment-channel-mechanics.md
```diff
@@ -40,21 +40,21 @@ make it easier to do this.
 ## Abstract
 
 This protocol defines the Stellar transactions that two participants use to open
-and close a payment channel by using escrow accounts to hold a single asset.
+and close a payment channel by using multi-sig accounts to hold a single asset.
```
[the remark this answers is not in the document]
acharb: i originally had it w/o the hyphen then second guessed myself at the end and thought the hyphen was better 😆 . no hyphen sgtm

specifications/starlight-payment-channel-mechanics.md
acharb: https://github.com/stellar/starlight/pull/408/commits/c87b50f84202e4b60630718f83034df1fab2d313

acharb:
@leighmcculloch ported the SEP changes over to: https://github.com/stellar/starlight/pull/410

will leave this PR for the code changes

examples/benchmark/main.go
```diff
@@ -72,7 +72,7 @@ func run() error {
 	}
 
 	accountKey := keypair.MustRandom()
-	escrowAccountKey := keypair.MustRandom()
+	multiSigAccountKey := keypair.MustRandom()
```
leighmcculloch, line 75: We should use `multisig` all lower case throughout, for the same reasons we used lower case in the SEP.

examples/console/main.go
```diff
@@ -232,33 +232,33 @@ func run() error {
...
+		tx, err := txbuild.CreateMultiSig(txbuild.CreateMultiSigParams{
 			Creator:        accountKey.FromAddress(),
-			Escrow:         escrowAccountKey.FromAddress(),
+			MultiSig:       multiSigKey.FromAddress(),
 			SequenceNumber: accountSeqNum + 1,
 			Asset:          txnbuild.NativeAsset{},
 		})
 		if err != nil {
-			return fmt.Errorf("creating escrow account tx: %w", err)
+			return fmt.Errorf("creating multisig account tx: %w", err)
```
leighmcculloch, line 247: For similar reasons as in the SEP change, I think it is confusing to use the term multisig here, because the account is not a multisig account in this moment, it's a singlesig account. I think we should probably try to omit the term multisig, simply replacing escrow with nothing, in the vast majority of places. Within the SDK we should find that the SDK only cares about a single account for each participant.

Where that might get a bit fuzzy is in the examples because in there each participant has two accounts. I think we can refer to that other account as the `fundingAccount`. Or `personalAccount`.

examples/benchmark/main.go
acharb: https://github.com/stellar/starlight/pull/408/commits/9afa4fc8779a5b22c6d22c2580330366c548857d 👍

examples/console/main.go
acharb: are you thinking we should change `type MultiSigAccount` to `type Account`?

A problem with that I can foresee is the confusion between the non-multisig account and the multisig account, since the channel has to be aware of the non-multisig accounts. ie. `Channel.localSigner` and `Channel.remoteSigner` 

if we removed `multisig` from the `multisigAccount` fields and variable names, I think it would be pretty easy to conflate the non-multisig account field/variables with the multisig ones.
eg. `channel.localSigner` seems confusing to have next to `channel.localAccount`

though I agree with removing `multiSig` from the error strings, as you pointed out.

wdyt?

examples/benchmark/main.go
acharb: changed `MultiSig` to `Multisig` here:
https://github.com/stellar/starlight/pull/408/commits/533547ead48c6dd883f31d32ee520b6fd8f2cc8e

examples/console/main.go
leighmcculloch: I think some of the confusion is that in the examples, and in the txbuild.CreateEscrow, we treat the localSigner as the sponsor and the signer, as this account that does both of those things. I think we should refer to that account as the sponsoring account and treat the signer as a different key. I think if we do that the channel will have an account that it is using, then anything outside the channel, like the sponsoring account, is out of scope for most of the code.

leighmcculloch: I think this will be less confusing if we get rid of CreateEscrow. It really doesn't need to be part of the SDK.

acharb: we could change `channel.localSigner` to `channel.localFundingSigner` though 🤔

```
type Channel struct {
   ...
	initiator             bool
	localAccount  *Account
	remoteAccount *Account

	localFundingSigner  *keypair.Full
	remoteFundingSigner *keypair.FromAddress
}
```
> Within the SDK we should find that the SDK only cares about a single account for each participant.

what's the main reason for this again?

and could we do in a separate PR?  Seems to me to be a separate simplification, since this PR was for removing `escrow` only

leighmcculloch: > `channel.localSigner` to `channel.localFundingSigner`

I don't think the distinction is meaningful. The local signer is the signer of the escrow account. That's actually all the channel needs to know, so `localSigner` is still a good name.

The fact that it is also the signer of the sponsoring account is happenstance and shouldn't be required.

I understand the desire to merge this as is, but there's still places where the term multisig is used where it is confusing as the account isn't a multisig account. Unfortunately, just replacing escrow with multisig doesn't retain the same clarity and I think we need to do more here.

acharb: ok I'm in for doing here 😄 . Are you thinking we should change `type MultisigAccount` to `type Account`?

leighmcculloch: We might need to iterate on this or work with it to try and figure it out. But I think target `type Account`, and I think in most of the SDK this should mean referring to the escrow account as simply the account in most places. i.e. localAccount and remoteAccount (different to now), and localSigner and remoteSigner (as they are now). It's probably too much to drop CreateEscrow in this PR, but if there's confusion with other accounts, we could use the term "sponsorAccount" in those places?

acharb: one last thought before I make this change 😄  

what if we used a different word rather than `Account`? Account by itself to me is pretty generic. Eg. there are methods and variables we have atm that use the word `account`, which is not a payment channel account. That may be confusing with this new type that is a special kind of Account. Instead of changing those to be `sponsorAccount` or another type of `account`, we could change this new type to have a different name.
eg. `initAccounts`, `createAccount` in the integration tests, some other spots we use the word `account` generically as a variable

what about `type Participant`?

leighmcculloch: That could work? We use the `<Noun>Details` pattern elsewhere in the SDK, maybe similar, to indicate this is just a data bag and isn't an actor or similar.

```go
type ParticipantDetails struct {
	Account *keypair.FromAddress
	Signer *keypair.FromAddress
}
```

I wonder if this could get confusing because the channel will have an account, and we would say the participant has an account, but the participants "account" is probably one way to talk about the sponsor's "account".

It's difficult to imagine this stuff without trying it out.

acharb:
@leighmcculloch changed multisig to channel account here
https://github.com/stellar/starlight/pull/408/commits/638bcbba50bf6863ad1f472ff3af06c6e01ca9d6 

I think this solves the problem you pointed out that the way `multisig` was used before was incorrect since it's not a `multisig` account at the start. But the `channel account` is created to be the `channel account` from the start, whether w/ one signer or multisig.

lmk what you think

leighmcculloch (approved):

PR #409, opened 2021-11-04
examples/console: fix deposits to other

leighmcculloch:
### What
When 'other' is included in the deposit command as the second parameter, have the deposit go to the other participant.

### Why
This is how the deposit used to work, but a bug was introduced recently which prevented us from following the code path where the deposit was directed to the other participant.


acharb (approved):

acharb:
oops, forgot to change that, thx

PR #411, opened 2021-11-05
all: update to use latest stellar/go branch

leighmcculloch:
### What
Update the stellar/go dependency to use the latest commit of the cap21and40 branch.

### Why
The latest commit contains updates to the version of go-xdr that is used that provides more efficient transaction encoding and decoding. It also contains updated XDR.


acharb (approved):
(•̪●)

PR #413, opened 2021-11-05
benchmarks: add updated benchmark report

leighmcculloch:


acharb (commented):
any reason the order is AB1->AB3->AB2->AB4?

acharb (approved):
lgtm!

PR #421, opened 2021-11-09
sdk/state: fix flakey test

leighmcculloch:
### What
Change CloseSigantures Equal function test to guarantee that two randomly generated CloseSignatures are different.

### Why
Both fuzzers are randomly generating `nil` values for the slices at the same time. There's a decent chance this will happen because by default they'll each independently do this 20% of the time, and that is a pretty high chance that does coincide with each other from time to time. This is causing intermittently failing tests.


acharb (approved):
Lgtm!

PR #423, opened 2021-11-09
txnbuild: add doc.go file

acharb:


acharb:
@leighmcculloch how much documentation do you think we should add? is commenting on all exported methods sufficient or overkill?

leighmcculloch (approved):
I think we can do our best to cover all exported functions, and we can split it up.

PR #425, opened 2021-11-09
sdk/agent/submit: add docs

leighmcculloch:
### What
Add some docs to the sdk/agent/submit package.

### Why

### DIFF
--- a/sdk/agent/submit/submitter.go
+++ b/sdk/agent/submit/submitter.go
@@ -1,3 +1,11 @@
+// Package submit contains a type and logic for submitting transactions that need to be fee bumped.
+//
+// The Submitter type can be used to wrap any submission logic, and it will
+// check if the transactions fee is below a threshold that indicates it needs to
+// be fee bumped, and if so, it will wrap it in a fee bump transaction before
+// submission.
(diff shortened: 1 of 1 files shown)


sdk/agent/submit/submitter.go
```diff
@@ -1,3 +1,11 @@
+// Package submit contains a type and logic for submitting transactions that need to be fee bumped.
+//
+// The Submitter type can be used to wrap any submission logic, and it will
+// check if the transactions fee is below a threshold that indicates it needs to
+// be fee bumped, and if so, it will wrap it in a fee bump transaction before
+// submission.
+//
+// This package is intended for use in example payment channel implementations.
```
acharb, line 8: it's not only for examples right? ie someone could use this for an app they're building
```suggestion
// This package is intended for use in payment channel implementations.
```

leighmcculloch: Right now the level of quality and flexibility is really suited towards writing examples and experimenting.  The state package not so, but the agent package is pretty incomplete.

acharb (approved):
lgtm!

PR #426, opened 2021-11-09
sdk/agent/horizon: add docs

leighmcculloch:
### What
Add some docs to the sdk/agent/horizon package.

### Why


acharb (approved):
lgtm!

PR #427, opened 2021-11-09
sdk/agent/agenthttp: add docs

leighmcculloch:
### What
Add some docs to the sdk/agent/agenthttp package.

### Why


acharb (approved):
lgtm!

PR #428, opened 2021-11-09
sdk/agent/bufferedagent: add docs

leighmcculloch:
### What
Add some docs to the sdk/agent/bufferedagent package.

### Why
To communicate the intent of the package, and the compatibility guarantees, or lack of, within.

### DIFF
--- a/sdk/agent/bufferedagent/agent.go
+++ b/sdk/agent/bufferedagent/agent.go
@@ -58,7 +58,7 @@ func NewAgent(c Config) *Agent {
 	return agent
 }
 
-// BufferedAgent coordinates a payment channel over a TCP connection, and
+// Agent coordinates a payment channel over a TCP connection, and
 // buffers payments by collapsing them down into single payments while it waits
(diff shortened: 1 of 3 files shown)


sdk/agent/bufferedagent/agent.go
```diff
@@ -55,7 +60,10 @@ func NewAgent(c Config) *Agent {
 
 // BufferedAgent coordinates a payment channel over a TCP connection, and
```
acharb, line 61: ❓ ik not technically part of this PR but for cleaning up comments should this be `Agent` instead of `BufferedAgent` here? I think it would be confusing for the reader to see `BufferedAgent` when the type is `Agent`, even though it is a buffered agent.

leighmcculloch:
```suggestion
// Agent coordinates a payment channel over a TCP connection, and
```

acharb (approved):
lgtm!

PR #429, opened 2021-11-09
sdk/agent: add docs

leighmcculloch:
### What
Add some docs to the sdk/agentpackage.

### Why
Most types and functions were documented, but some were missing, and this fills the gaps.

### DIFF
--- a/sdk/agent/agent.go
+++ b/sdk/agent/agent.go
@@ -83,7 +83,7 @@ type Config struct {
 	Events chan<- interface{}
 }
 
-// NewAgent constructs a new buffered agent with the given config.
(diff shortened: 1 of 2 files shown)


sdk/agent/agent.go
```diff
@@ -81,6 +83,7 @@ type Config struct {
 	Events chan<- interface{}
 }
 
+// NewAgent constructs a new buffered agent with the given config.
```
acharb, line 86:
```suggestion
// NewAgent constructs a new agent with the given config.
```

acharb (approved):
lgtm!

PR #434, opened 2021-11-10
examples/console: support friendbot for deployed environments

leighmcculloch:
### What
Add support for deployed environments with a friendbot for creating the accounts in the console application.

### Why
To make the console application compatible with deployed environments where there is a friendbot for creating new test accounts.


acharb (approved):
lgtm!

PR #436, opened 2021-11-11
sdk/state: add godoc comments for state.go

acharb:
**WHAT**
add godoc comments to sdk/state/state.go

**WHY**
better documentation, clearer intent

### DIFF
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -81,11 +81,12 @@ func NewChannelFromSnapshot(c Config, s Snapshot) *Channel {
 	return channel
 }
 
-// ChannelAccount holds the details that a Channel tracks for a Stellar account. A Channel tracks the details of two Stellar accounts, one for each participant in the channel.
+// ChannelAccount holds the details that a Channel tracks for a Stellar account.
+// A Channel tracks the details of two Stellar accounts, one for each
+// participant in the channel.
 //
 // The channel accounts hold the assets that are used for payments and will be
-// re-distributed at close.  The initiator channel account's sequence number
-// serves as the sequence number for the channel.
+// re-distributed at close.
 type ChannelAccount struct {
 	Address                    *keypair.FromAddress
 	SequenceNumber             int64
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -95,7 +95,7 @@ type ChannelAccount struct {
 	LastSeenTransactionOrderID int64
 }
 
-// Channel holds the data for a Starlight payment channel.
+// Channel holds the state of a single Starlight payment channel.
 type Channel struct {
 	networkPassphrase string
 	maxOpenExpiry     time.Duration
--- a/sdk/state/state.go
+++ b/sdk/state/state.go
@@ -81,9 +81,7 @@ func NewChannelFromSnapshot(c Config, s Snapshot) *Channel {
 	return channel
 }
 
-// ChannelAccount is an account that, with another channel account, form a
-// channel. Two participants are needed to form a channel, referred to as remote
-// and local.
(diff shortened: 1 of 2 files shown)


sdk/state/state.go
```diff
@@ -79,13 +81,21 @@ func NewChannelFromSnapshot(c Config, s Snapshot) *Channel {
...
+// re-distributed at close.  The initiator channel account's sequence number
+// serves as the sequence number for the channel.
 type ChannelAccount struct {
 	Address                    *keypair.FromAddress
 	SequenceNumber             int64
 	Balance                    int64
 	LastSeenTransactionOrderID int64
 }
 
+// Channel holds the data for a Starlight payment channel.
```
leighmcculloch, line 98:
```suggestion
// Channel holds the state of a single Starlight payment channel.
```

sdk/state/state.go
```diff
@@ -79,13 +81,21 @@ func NewChannelFromSnapshot(c Config, s Snapshot) *Channel {
 	return channel
 }
 
+// ChannelAccount is an account that, with another channel account, form a
+// channel. Two participants are needed to form a channel, referred to as remote
+// and local.
```
leighmcculloch, line 86: Suggested rewording here. I removed the local/remote, because this type has no knowledge of those concepts and so one could look at the type and struggle to identify where is local and remote captured in the fields that are there.
```suggestion
// ChannelAccount holds the details that a Channel tracks for a Stellar account. A Channel tracks the details of two Stellar accounts, one for each participant in the channel.
```

sdk/state/state.go
```diff
@@ -79,13 +81,21 @@ func NewChannelFromSnapshot(c Config, s Snapshot) *Channel {
...
+//
+// The channel accounts hold the assets that are used for payments and will be
+// re-distributed at close.  The initiator channel account's sequence number
+// serves as the sequence number for the channel.
```
leighmcculloch, line 90: Similar here, I wouldn't identify initiator because again it's not clear from looking at the fields within which field identifies the account as an initiator.

I think these details are awesome, but they belong in the package doc comment.

sdk/state/state.go
```diff
@@ -183,10 +193,15 @@ func (c *Channel) setInitiatorChannelAccountSequence(seqNum int64) {
...
+// IsInitiator returns true if this channel initiated the process of opening a
+// channel, else false.
 func (c *Channel) IsInitiator() bool {
 	return c.initiator
 }
 
+// NextIterationNumber returns the next iteration number for the channel. If
+// there is a pending unauthoried close agreement, then that agreement iteration
+// is used, else the latest authorized agreeement is used.
 func (c *Channel) NextIterationNumber() int64 {
```
leighmcculloch, line 205: 💭 Maybe we should make this function not-exported?

leighmcculloch (approved):
Some suggestions.

sdk/state/state.go
leighmcculloch: I'm not suggesting we address this in this PR, I'm just noticing this.

sdk/state/state.go
acharb: 👍 
https://github.com/stellar/starlight/pull/436/commits/16344c3b9e9566b1307dae144e7b7c1924f6869c

sdk/state/state.go
acharb: 🤔  trying to think if there's a use case for an external app needing to know. I guess only the channel needs to know it. Makes sense to me to not-export

though we'd have to change modify the integration tests that is using it for logging

leighmcculloch: I don't think there's a need. It'd keep the API slimmer which would be nice. I'll have a look if the integration tests really need it.

leighmcculloch: This doesn't need to block this Pr though, this PR is great.

leighmcculloch: #437

PR #437, opened 2021-11-11
sdk/state: unexport Channel.NextIterationNumber

leighmcculloch:
### What
Unexport the NextIterationNumber method of the Channel type.

### Why
It's dubious why a consumer of the package would need this function. If anything it exposes a little too much about the internals of the package. It may actually be difficult to use the function correctly or to fully understand why it returned the number it returned. In some cases it returns the current iteration, other times the next, depending on state and what is happening. This is because it always returns the iteration number of the next authorized transaction but that is counter intuitive when a participant has proposed a new payment because that payment is unauthorized. Unexporting the function removes this complexity from the packages surface API.

The only place it is used outside the package is in the integration tests, and there only for logging. The logging may be useful, but I don't think it is critical.


acharb (approved):
sgtm 👍

PR #445, opened 2022-03-26
all: migration to protocol 19

### LINKED ISSUE #443: all: update to use formal implementations of CAP-21/40
### What
Update to use formal implementations of CAP-21/40.

### Why
CAP-21 and CAP-40 have both been accepted. Once functional versions of stellar-core and horizon are available we should start migrating Starlight to use and test against those rather than the prototypes it does now.

leighmcculloch:
### What
Migration to Protocol 19.

### Why
Protocol 19 is under development and will contain implementations of CAP-21 and CAP-40.

Close #443 

Dependent on #448 
Dependent on #449


acharb (approved):
🙌  lgtm!

PR #446, opened 2022-04-07
all: move compat to go118 and use go.work

leighmcculloch:
### What
Move compatibility to Go 1.18 and use go.work.

### Why
Go.work was added in Go 1.18 and makes it easier to work in repos that have multiple go.mod files. This repo has always maintained compatibility only with the most recent version of Go, so moving compat to Go 1.18 is fine as it is the latest version.

### DIFF
--- a/go.work
+++ b/go.work
@@ -1,0 +1,8 @@
+go 1.18
+
+use (
+	examples/benchmark
+	examples/bufferedbenchmark
+	examples/console
+	sdk
+)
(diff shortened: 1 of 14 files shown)


go.work
```diff
@@ -0,0 +1,8 @@
+go 1.18
```
acharb, line 1: ooh go.work looks awesome 🎉

acharb (commented):
lgtm!

acharb (approved):

PR #447, opened 2022-04-08
examples: remove old examples

leighmcculloch:
### What
Remove the benchmark, bufferedbenchmark, and stctxs examples.

### Why
These examples are out-of-date. The first two are replaced by the console example and all benchmarks we recorded in the benchmarks directory are using the console app.


acharb (approved):
sad to see these go but agreed the console example is sufficient 🥲 . lgtm!

PR #449, opened 2022-04-09
sdk: rewrite agent tests to be resilient to xdr changes

leighmcculloch:
### What
Replace hardcoded xdr with using the txbuildtest package to generate test xdr.

### Why
To make the tests resilient to xdr changes. The xdr is changing for the actual protocol 19 implementation, and so these need rewriting. May as well use the helper.

Dependent on https://github.com/stellar/starlight/pull/448


acharb (approved):
lgtm!

PR #454, opened 2022-04-29
sdk/agent/bufferedagent: allow close after declare close

leighmcculloch:
### What
Allow close to be called on the buffered agent after declare close has been called.

### Why
If close is called after declare close a panic occurs because the bufferReady Go channel has already been closed, and Go channels can only be closed once. We've used this code in examples and demos frequently so it is surprising we didn't run into this problem before. I think it is likely when we've tested it we've more often than not tested scenarios where the close happens automatically, or the other participant has executed the close so close isn't being called on the same buffered agent as the declare close.

### Limitations
The buffered agent is very experimental and we don't have any tests for it yet, so I haven't gone out of my way to write some foundational test to update for this fix.

### DIFF
--- a/sdk/agent/bufferedagent/agent.go
+++ b/sdk/agent/bufferedagent/agent.go
@@ -52,6 +52,12 @@ func NewAgent(c Config) *Agent {
 
 		events: c.Events,
 	}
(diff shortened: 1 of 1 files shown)


acharb (approved):
lgtm!

PR #455, opened 2022-04-30
examples/console: add log for when close occurs that is outdated

leighmcculloch:
### What
Add log for when close occurs that is outdated.

### Why
Helpful for debugging.


acharb (approved):
lgtm!

PR #458, opened 2022-06-13
Rough in a Rationale section outlining why CAP 21 + 40 usage makes sense

tyvdh:


specifications/starlight-payment-channel-mechanics.md
```diff
@@ -589,6 +589,20 @@ Changes in the networks base reserve do not impact the channel.
...
+The protocol does not, and must not, create new sponsorship of ledger entries
+while the channel is open. Any Stellar operation that creates a ledger entry
+depends on sponsorship may fail when being applied if the sponsorship cannot be
+satisfied. When the operation fails, the transaction containing it fails,
+consuming the sequence number of the transaction even though it was not
+successful. If the sequence number of the declaration or close transactions are
+consumed without being successful the channel may be in a state where
+participants would need to collaborate honestly to close the channel.
+
+This constraint is why the protocol does not use claimable balances or
```
tyvdh, line 603: I don't quite understand why this has much to do with claimable balances. Or at least not from this comment. Would love more clarity over statements or when there are statements like this that clarity is added on how the logic here holds for claimable balances.

PR #462, opened 2023-04-14
Fix broken link in readme

leighmcculloch:
### What
Change the Discord link to point to the public Stellar Dev Discord.

### Why
The link in the readme looks like it is an expired link. There's no risk created by this, however we should fix it so the link is not broken and still leads to the Stellar Dev Discord. Since the Discord server is public now there's no need for an invite link link before.

### DIFF
--- a/examples/console/stats.go
+++ b/examples/console/stats.go
@@ -108,7 +108,7 @@ func (s *stats) AgreementsPerSecond() float64 {
 	if timeFinish.IsZero() {
 		timeFinish = time.Now()
 	}
-	duration := s.timeFinish.Sub(s.timeStart)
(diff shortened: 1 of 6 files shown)


ire-and-curses (approved):

examples/console/stats.go
```diff
@@ -108,7 +108,7 @@ func (s *stats) AgreementsPerSecond() float64 {
 	if timeFinish.IsZero() {
 		timeFinish = time.Now()
 	}
-	duration := s.timeFinish.Sub(s.timeStart)
+	duration := timeFinish.Sub(s.timeStart)
```
leighmcculloch, line 111: Golangci-lint picked up on a bug here in an example, not in the sdk.

